PR #25, opened 2016-01-08
Fix tests for GitHub failing

orta:
I was getting fails due to not having `DANGER_GITHUB_API_TOKEN` in my personal env. CI wouldn't register this, as it has it set up.


KrauseFx:
LGTM :rocket:

PR #26, opened 2016-01-08
Use bundler caching in CI

orta:
Not sure if I added the:

```
- gem update --system
- gem install bundler
```

If I did, they can go, if you did, you can choose.


KrauseFx:
Looks good. Why is cache not default?

PR #29, opened 2016-01-09
Don't show the warnings / error tables if you have no errors / warnings

orta:
( Note: I'm sending this to #27 to keep the PR readable. )

PR #30, opened 2016-01-09
Support issue editing instead of new issue creation

orta:
Note this goes to #29 in order to keep the PR sane.

Rather than creating a new issue each time, look up whether we can edit an existing issue.

### DIFF
--- a/lib/danger/request_sources/github.rb
+++ b/lib/danger/request_sources/github.rb
@@ -50,9 +50,16 @@ module Danger
         # Just remove the comment, if there's nothing to say.
         delete_old_comments!
       else
+        issues = client.issue_comments(ci_source.repo_slug, ci_source.pull_request_id)
+        editable_issues = issues.reject { |issue|  issue[:body].include?("generated_by_danger") == false }
         body = generate_comment(warnings: warnings, errors: errors, messages: messages)
-        comment_result = client.add_comment(ci_source.repo_slug, ci_source.pull_request_id, body)
-        delete_old_comments!(except: comment_result[:id])
+
+        if editable_issues.empty?
+          comment_result = client.add_comment(ci_source.repo_slug, ci_source.pull_request_id, body)
+        else
+          original_id = editable_issues.first[:id]
+          comment_result = client.update_comment(ci_source.repo_slug, original_id ,body)
+        end
       end
 
       # Now, set the pull request status.
@@ -105,7 +112,7 @@ module Danger
       end
     end
 
-    def generate_comment(warnings: nil, errors: nil, messages: nil)
+    def generate_comment(warnings: [], errors: [], messages: [])
       require 'erb'
 
       md_template = File.join(Danger.gem_path, "lib/danger/comment_generators/github.md.erb")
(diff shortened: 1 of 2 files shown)


orta:
That works, adding specs now.

orta:
Specs are in, this is :+1:

KrauseFx:
Yah, so this keeps the comment in place. Why don't we like to get an email for every commit? How about we only post a comment when it has actually changed since last time?

orta:
PRs would look like this in my inbox:

screen shot 2016-01-09 at 14 42 39

I'm not sure if it's worth emailing to everyone involved in the project on every update that Danger has that's different either. It's a good idea. These these results are important, sure, but it's not _that_ important to everyone else when they change.

lib/danger/request_sources/github.rb
```diff
@@ -50,9 +50,16 @@ def update_pull_request!(warnings: nil, errors: nil, messages: nil)
         # Just remove the comment, if there's nothing to say.
         delete_old_comments!
       else
+        issues = client.issue_comments(ci_source.repo_slug, ci_source.pull_request_id)
+        editable_issues = issues.reject { |issue|  issue[:body].include?("generated_by_danger") == false }
```
KrauseFx: Should we move the `generated_by_danger` string to some central space @orta?

spec/sources/github_spec.rb
```diff
@@ -74,18 +73,51 @@ def fixture(file)
...
+
+        body = @g.generate_comment(warnings: ["hi"], errors: [], messages: [])
+        expect(@g.client).to receive(:add_comment).with("artsy/eigen", "800", body).and_return({})
+
+        @g.update_pull_request!(warnings: ["hi"], errors: [], messages: [])
+      end
+
+      it "updates the issue if no danger comments exist" do
+        issues = [{:body => "generated_by_danger", :id => "12"}]
+        allow(@g.client).to receive(:issue_comments).with("artsy/eigen", "800").and_return(issues)
```
KrauseFx: Shouldn't this be `expect` as this method must be called?

lib/danger/request_sources/github.rb
orta: Yeah, eventually, I figure that'd happen when someone builds support for non-github stuff :+1:

spec/sources/github_spec.rb
orta: Yes, and No. :dancer: Heh.

Yes, because it is part of the methods chain that are called. No, because that is not what we are testing here. 

To use@jonreid [here](https://speakerdeck.com/jonreid/unit-testing-control-your-dependencies?slide=6)

screen shot 2016-01-10 at 08 10 40

The stubbing of `issue_comments` is a part of the arrangement for the test, not the assertion that `:update_comment`gets called. That's why once these tests became longer, I started splitting it into three sections.

orta:
poke

KrauseFx:
You are so wise @orta. Ready to go from my side. I can do another release once this is merged :+1:

orta:
haha, cool let's do it :+1:

PR #35, opened 2016-01-20
Improve support on Circle CI

orta:
I realised that I made the Circle source work differently to the others, and it was the source of my confusion. 

In doing this PR I figured I'd improve our changelog and get Danger involved in making sure we keep it up to date ;)

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -17,3 +17,7 @@ end
 if pr_body.length < 5
   fail "Please provide a changelog summary in the Pull Request description @#{pr_author}"
 end
+
+unless files_modified.any? { |a| a.include?("CHANGELOG.md") }
+  warn("Please include a CHANGELOG entry")
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+## Master
+
+* Makes the CircleCI provider validate, but not run on non-PR builds - orta
+
+## 0.2.1
+
+* Edits an existing ticket rather than making a new one - orta
+
+## 0.2
+
+* Support making comments on a GitHub PR - Felix
+* Use GitHub status API to provide extra info on a PR - Felix
+* DRY the HTML comment - orta
+* Don't show a message if there are not warnings/errors - orta
+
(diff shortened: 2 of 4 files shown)


ashfurrow:
Nice, using Danger to build Danger :+1:

orta:
:dancers:

Dangerfile
```diff
@@ -17,3 +17,7 @@ end
 if pr_body.length < 5
   fail "Please provide a changelog summary in the Pull Request description @#{pr_author}"
 end
+
+unless files_modified.any? { |a| a.include?("CHANGELOG.md") }
```
orta: @KrauseFx would it feel reasonable to allow something like this using custom objects that act array-ish but provide dictionary-like DSL attribtues?

``` ruby
warn("Please include a CHANGELOG entry") unless file_modified["CHANGELOG.md"]
```
```

Comment (KrauseFx) [inline Dangerfile]: YES! And if it's in a subfolder, do we want `file_modified["docs/something.md"]` or `file_modified["docs"]["something.md"]`

**Comment by KrauseFx** [comment]
Looks good to me, however I've been using the GitHub releases up until now. I know you prefer `Changelog.md`

File: Dangerfile (hunk 1 above)
Comment (orta) [inline Dangerfile]: I'd opt for just dealing with one level of allowing "docs/thingy"  - I'd imagine some special casing  around `file_modified["docs/"]` working if any file inside that dir is changes.

Other options is to support whatever glob thing CocoaPods supports ( I think it uses `Dir.glob` under the hood )

Comment (KrauseFx) [inline Dangerfile]: Let's use `glob`:

```
file_modifieds["docs/**/*.md"].count == 4

file_modified?["docs/**/Changelog.md"]
```

What do you think?

orta: are it `files_modified` for the first one, and I'm sold. :+1:

orta:
I don't believe in github releases, changelogs should be in the project - not a github specific special web page

KrauseFx:
Okay :rocket:

PR #38, opened 2016-01-28
Add support for GitHub compare URLs that don't conform to URI

orta:
This PR https://github.com/artsy/eigen/pull/1076 totally threw me for a loop, I have no idea why we're being passed such a funky URL, but now we [don't crash](https://circleci.com/gh/artsy/eigen/1389).


KrauseFx:
Looks good :rocket:

Issue #32, opened 2016-01-13
Ensure that all comments are downloaded 


orta:
I wonder if in [this PR](https://github.com/artsy/eigen/pull/1023) Danger has paged off at 5 comments, and now continues to re-post

KrauseFx:
Could you clarify what you mean?

orta:
Sure, I don't know if [this line](https://github.com/KrauseFx/danger/blob/2ccaeb064a728d7874b78e6f21b6ba6502569924/lib/danger/request_sources/github.rb#L53) which grabs the comments does it via paging or not. If it does use paging, then it might only download 5 at a time for example. Then the logic about deletion / editing etc is based off bad data.

marcelofabri:
This was fixed by #39, right?

orta:
Yes!

PR #39, opened 2016-01-30
Make sure we grab all comments for github

orta:
in #32 I wondered if there were x amounts of comments being downloaded, looks like 30 are downloaded. It's definitely possible that a PR can get that many comments, so I think it's worth letting danger deal with all of them.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,9 +1,20 @@
 require "bundler/gem_tasks"
 require "rspec/core/rake_task"
+require 'rubocop/rake_task'
 
-RSpec::Core::RakeTask.new(:spec)
+RSpec::Core::RakeTask.new(:specs)
 
 task default: :spec
+
+task :spec do
+  Rake::Task['specs'].invoke
+  Rake::Task['rubocop'].invoke
+end
+
+desc 'Run RuboCop on the lib/specs directory'
+RuboCop::RakeTask.new(:rubocop) do |task|
+  task.patterns = ['lib/**/*.rb', 'spec/**/*.rb']
+end
 
 task :test do
   sh "fastlane test"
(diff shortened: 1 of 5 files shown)


orta:
I've added rubocop to `rake spec` too :+1:

orta:
screen shot 2016-01-30 at 16 34 04
err, hrm, that did not work.

Rakefile
```diff
@@ -1,10 +1,15 @@
 require "bundler/gem_tasks"
 require "rspec/core/rake_task"
 
-RSpec::Core::RakeTask.new(:spec)
+RSpec::Core::RakeTask.new(:specs)
 
 task default: :spec
 
+task :spec do
+    Rake::Task['specs'].invoke
```
KrauseFx: Should be 2 spaces, right?

lib/danger/request_sources/github.rb
```diff
@@ -50,7 +50,7 @@ def update_pull_request!(warnings: nil, errors: nil, messages: nil)
         # Just remove the comment, if there's nothing to say.
         delete_old_comments!
       else
-        issues = client.issue_comments(ci_source.repo_slug, ci_source.pull_request_id)
+        issues = client.issue_comments(ci_source.repo_slug, ci_source.pull_request_id, auto_pagination: true)
```
KrauseFx: AFAIK it's possible to set this globally, so we don't have to do this for each request

orta:
Alright, using the global setting for auto paginate, and fixed the `rake spec` doing all of ruby

KrauseFx:
Looks good, and tests are passing :rocket:

KrauseFx:
Shipping so fast :heart:

PR #43, opened 2016-02-04
Remove the specifity of claide as there's now cp 1.0

orta:
Danger can't be used on a project with CLAide 1.0, as there are no breaking API changes it's fine to allow both.


KrauseFx:
Looks good :ship:

Issue #51, opened 2016-02-10
Customize the merge branch 


marcelofabri:
Today, danger only runs on PRs against `master` (at least the README says so).

It'd be awesome to be able to customize the merge branch. We, for example, keep a shipped version on master and all the development goes on a `development` branch.

I can try to take a look at this, but it seems it should be easy. Unless I'm missing something, only two places uses `master` hardcoded (https://github.com/danger/danger/blob/master/lib/danger/ci_source/buildkite.rb#L20 and https://github.com/danger/danger/blob/df58dcbad790cc14010fe1343e9511fce206597a/lib/danger/scm_source/git_repo.rb#L9). 

I think one important question is where this branch would be specified: on the `Dangerfile` (if so, this probably would be harder because it seems that the file actually is run) or as a command line parameter.

orta:
This is in with #46 

It actually is pretty smart about this, depending on your CI provider. It depends on what it exposes as ENV vars, as you spotted. 

TBH though, I think we'd need to do more. Danger needs to have a access to the origin's merge branch for the base and a remote branch for a fork maybe. I think the fork would be what the CI runs on, so it needs to be able to pull the origin if needed. Making the process more like:

* Find out what CI it's on
* Find out the PR number
* Grab PR metadata
* Look inside PR JSON for the commit range
* Check if the remotes are the same, otherwise ensure that we have the full commit range
* Now Danger can grab the full PR's commits to do git checking on

marcelofabri:
I'm kind of lost on what's done and what's not. What would be the next steps (and how I could help)?

orta:
Done:

* ~~Find out what CI it's on~~
* ~~Find out the PR number~~
* ~~Grab PR metadata~~
* Look inside PR JSON for the commit range
* Check if the remotes are the same, otherwise ensure that we have the full commit range
* Now Danger can grab the full PR's commits to do git checking on

The logic around deciding what the commit range to choose should move from the CI Source to the request source (GitHub from the CI.) The CI source should only exist to help the request  _find_ the PR to look at. Then all commit range data can come from the PR's canonical JSON.

marcelofabri:
@orta moving to request source wouldn't make #45 happen again? If there's no `CI_PULL_REQUEST` set, there's (apparently) no way to get the PR.

marcelofabri:
I've started this on #53 but I'm not sure why "Check if the remotes are the same, otherwise ensure that we have the full commit range" is needed and how we would do that.

[...]

Issue #45, opened 2016-02-08
Circle CI's commit range is  a subset of the PR


orta:
This was the diff for [this PR](https://github.com/artsy/eigen/pull/1125) `git diff  f12ea82b719a 9d9fece70aac` - [see build](https://circleci.com/gh/artsy/eigen/1476)

We need to find the _first_ commit which in that PR could be either `608f2d71a0ac3c88d7681c464ec0cbe9fa5e319d` or it's parent `c1af413`

the env vars from Circle are:

```
  CIRCLE_ARTIFACTS=/tmp/circle-artifacts.oihhrPZ
  CIRCLE_BRANCH=orta-deploy
  CIRCLE_BUILD_NUM=1476
  CIRCLE_COMPARE_URL=https://github.com/artsy/eigen/compare/f12ea82b719a...9d9fece70aac
  CIRCLE_NODE_INDEX=0
  CIRCLE_NODE_TOTAL=1
  CIRCLE_PREVIOUS_BUILD_NUM=1471
  CIRCLE_PROJECT_REPONAME=eigen
  CIRCLE_PROJECT_USERNAME=artsy
  CIRCLE_SHA1=9d9fece70aacd8e01f9b316664e4180629910f5f
  CIRCLE_TEST_REPORTS=/tmp/circle-junit.R5MI8iJ
  CIRCLE_USERNAME=orta
```

orta:
We can ignore the `compare/f12ea82b719a` bit of the `CIRCLE_COMPARE_URL` - from there though it can either grab the full PR metadata ( which should provide the base commit to build from ) or we just do `master` for the moment.

orta:
https://circleci.com/gh/artsy/eigen/1497 tried working against a commit that didn't exist

PR #53, opened 2016-02-11
Look inside PR JSON for the commit range

marcelofabri:
This was discussed on #51. However, this makes necessary to have an environment variable to get the PR number, so it might reintroduce #45 problem.


orta:
CHANGELOG please, then you're fine to merge this - going to turn that into a fail.

PR #55, opened 2016-02-11
Updating README and CHANGELOG after pr_labels

marcelofabri:
Forgot to change these :sweat:


orta:
Ah, the circle token was revoked for being in a GitHub commit, shame.

Issue #48, opened 2016-02-10
Allow posting comments regardless of ability to change PR status


orta:
Currently it won't write a message. You can see it in #47

marcelofabri:
Shouldn't be this working, since `submit_pull_request_status!` is only called as the last thing in `update_pull_request!`?

orta:
Yeah, it is actually, turns out all of the github accounts I've been trying to use get flagged as bots. So they're actually posting messages to our threads, but GH has banned us from seeing them

orta:
![screen shot 2016-02-13 at 08 14 12](https://cloud.githubusercontent.com/assets/49038/13026712/d24fd764-d229-11e5-8a33-8b7ced16bdb2.png)

PR #56, opened 2016-02-11
Enforce a changelog entry

orta:
:+1: 

/cc @marcelofabri


KrauseFx:
:+1:

KrauseFx:
So sad we don't use danger for danger though

orta:
We are:

```
The command "bundle exec rake spec" exited with 0.
1.60s$ bundle exec danger --verbose
Printing message 'Tests were updated / added'
All green. Congrats.

```

I've just not done https://github.com/danger/danger/issues/48 - which will support sending the message, but not doing the status

marcelofabri:
:clap:

orta:
Doing it, man, this CI stuff is complicated ;)

PR #57, opened 2016-02-11
Calling CircleCI API when CI_PULL_REQUEST is not set

### LINKED ISSUE #49: Circle CI can sometimes create a commit build for pull requests
This CI build https://circleci.com/gh/artsy/eigen/1501 - is attached to this PR: https://github.com/artsy/eigen/pull/1130 

screen shot 2016-02-10 at 09 50 20

it shows that at the top, but the ENV vars don't reflect that that:

``` sh
Setting build environment variables
  CIRCLE_ARTIFACTS=/tmp/circle-artifacts.gyNRoD5
  CIRCLE_BRANCH=orta-dm_pilot
  CIRCLE_BUILD_NUM=1501
  CIRCLE_COMPARE_URL=https://github.com/artsy/eigen/commit/1b064fe71ac7
  CIRCLE_NODE_INDEX=0
  CIRCLE_NODE_TOTAL=1
  CIRCLE_PREVIOUS_BUILD_NUM=1500
  CIRCLE_PROJECT_REPONAME=eigen
  CIRCLE_PROJECT_USERNAME=artsy
  CIRCLE_SHA1=1b064fe71ac7ec073223221889d96dcecb292000
  CIRCLE_TEST_REPORTS=/tmp/circle-junit.X2xv4ph
  CIRCLE_USERNAME=orta
```

our validation: 
``` ruby
def self.validates?(env)
  return !env["CIRCLE_BUILD_NUM"].nil? && !env["CI_PULL_REQUEST"].nil?
end
```

marcelofabri:
- [x] CHANGELOG entry
- [x] Create CLI option and environment variable to set `circle-token`

It should fix #49


danger.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'colored'
   spec.add_runtime_dependency 'nap'
   spec.add_runtime_dependency 'octokit'
+  spec.add_runtime_dependency 'rest-client'
```
orta: rest client is an extra dependency we don't need, faraday is already in the dependency graph, any chance you can use that? https://github.com/lostisland/faraday

marcelofabri: sure, I'll take a look at that later. I hadn't used rest-client before also, just seen some gem using it and thought it was good enough.

orta:
Everything feels good in the implementation though!

danger.gemspec
orta: Ace :+1:

marcelofabri:
Travis failed because of CHANGELOG, so #56 really worked!

I'm updating it after, so I can update the README at the same time with information about the new parameter.

marcelofabri:
I don't know if this deserves a mention on README :grin:

orta:
I think it's best that we start building out docs, will try take a stab at this in the morning :+1:

lib/danger/runner.rb
```diff
@@ -20,13 +21,15 @@ def validate!
     def self.options
       [
         ['--base=[master|dev|stable]', 'A branch/tag/commit to use as the base of the diff'],
-        ['--head=[master|dev|stable]', 'A branch/tag/commit to use as the head']
+        ['--head=[master|dev|stable]', 'A branch/tag/commit to use as the head'],
+        ['--circle-ci-token=[token]', 'A Circle CI API token to be used if needed']
```
orta: IMO, kill this option, only allow setting this form an env var, otherwise you'd be leaking it in the build logs

spec/sources/circle_api_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    expect(api.fetch_build("artsy/eigen", "1500")).to eql(@expected_json_response)
+  end
+
+  it 'fetches the build info with token' do
+    api = Danger::CircleAPI.new('123456')
+    allow(api.client).to receive(:get).with(api_path, { :'circle-token' => '123456' }, accept_json).and_return(@mocked_response)
+
+    expect(api.fetch_build("artsy/eigen", "1500")).to eql(@expected_json_response)
+  end
+end
```
orta: cool, this all looks good!

lib/danger/runner.rb
marcelofabri: makes sense :+1:

orta:
Yep, yep yep.

orta:
Thanks sir

PR #66, opened 2016-02-13
Fix commands help

marcelofabri:
```
✗ danger --help     
Usage:

    $ danger COMMAND

      Run the Dangerfile.

Commands:

    + init                       Creates a Dangerfile.
    + local                      Run the Dangerfile locally.

Options:

    --base=[master|dev|stable]   A branch/tag/commit to use as the base of the diff
    --head=[master|dev|stable]   A branch/tag/commit to use as the head
    --version                    Show the version of the tool
    --verbose                    Show more debugging information
    --no-ansi                    Show output without ANSI codes
    --help                       Show help banner of specified command
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Master
 
+* Show command summary in help - marcelofabri
 * Use 100% width tables for messages - marcelofabri
 
 ## 0.3.0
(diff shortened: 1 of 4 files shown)


orta:
ohhhh, good reading up!
OK, we should also add the CHANGELOG to the gitconfig as automerge-both

orta:
https://github.com/artsy/eigen/blob/master/.gitattributes

marcelofabri:
should that be another PR?

marcelofabri:
Done in #67

PR #84, opened 2016-02-27
Syntax improvement. #trivial

amonshiz:
Using an active voice instead of passive. Also, `being` was misspelled.


orta:
Ace :+1: shipppppit

PR #87, opened 2016-02-27
Lock dependencies #trivial

orta:
By using an older ( like by 2 weeks or something :P ) version of Fastlane, my Gemfile resolved our grit version to be a version many versions ago, and so https://github.com/artsy/eigen/pull/1188 failed. I've locked grit to the latest build and made everything else pretty flex

### DIFF
--- a/lib/danger/version.rb
+++ b/lib/danger/version.rb
@@ -1,4 +1,4 @@
 module Danger
-  VERSION = "0.3.0"
+  VERSION = "0.5.0"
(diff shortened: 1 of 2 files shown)


lib/danger/version.rb
```diff
@@ -1,4 +1,4 @@
 module Danger
-  VERSION = "0.3.0"
+  VERSION = "0.5.0"
```
KrauseFx: Wow, we're going fast :rocket:

PR #91, opened 2016-02-28
Removing <p> tag from table comment #trivial

### LINKED ISSUE #81: Remove <p> tags from table
The `<p>` tags make the text not centered on each cell, as seen in https://github.com/artsy/eigen/pull/1185#issuecomment-186255794

This was a side effect of https://github.com/danger/danger/pull/65

marcelofabri:
This is kind of hacky, but fixes #81. 

I also replaced the repo URL to the new organization.


orta:
Yep - I think this is a fine way to do it.

PR #99, opened 2016-02-29
Replaced … with ... #trivial

KrauseFx:
These close 3 dots look weird


orta:
Any specific reason? I went out of my way to use the real unicode characters here.

KrauseFx:
These close 3 dots look weird. Any reason why we should use them? Just having 3 dots looks better :+1:

KrauseFx:
Ready to merge @orta :rocket:

KrauseFx:
:dancers:

PR #100, opened 2016-02-29
Fixed crash on `danger init`

KrauseFx:
Fixes
```
In order to expose an environment variable, go to:
/Users/felixkrause/Developer/danger/lib/danger/commands/init.rb:234:in `circle_token': undefined local variable or method `repo_slug' for #<Danger::Init:0x007f9cfcdfc208> (NameError)
	from /Users/felixkrause/Developer/danger/lib/danger/commands/init.rb:159:in `setup_danger_ci'
```

### DIFF
--- a/lib/danger/commands/init.rb
+++ b/lib/danger/commands/init.rb
@@ -231,7 +231,7 @@ module Danger
       end
 
       ui.say "In order to expose an environment variable, go to:"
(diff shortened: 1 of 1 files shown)


orta:
This is non-trivial, should have a changelog entry

PR #111, opened 2016-02-29
Update CHANGELOG + Dangerfile

orta:
I've moved your CHANGELOG notes to master, and started improving the Dangerfile. Now PRs to non-app parts don't have to be declared trivial, and we only warn when there are no tests and app code has been changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Master
+
+* Added internal plugin system - KrauseFx
+* Refactored unit tests - KrauseFx
+
 ## 0.5.2
 
 * Typo fixes for `danger init` - lumaxis 
@@ -17,8 +22,6 @@
 * Remove `nap` dependency - marcelofabri
 * Show command summary in help - marcelofabri
 * Use 100% width tables for messages - marcelofabri
-* Added plugin system - KrauseFx
-* Refactored unit tests - KrauseFx
 
 ## 0.3.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Master
+
+* Added internal plugin system - KrauseFx
```
KrauseFx: Wups, sorry

Dangerfile
```diff
@@ -1,25 +1,26 @@
...
+has_test_changes = !files_modified.grep(/spec/).empty?
 
 if ["KrauseFx", "orta"].include?(pr_author) == false
-  warn("Author @#{pr_author} is not a contributor")
+  warn "Author @#{pr_author} is not a contributor"
 end
 
 if (pr_body + pr_title).include?("WIP")
-  warn("Pull Request is Work in Progress")
+  warn "Pull Request is Work in Progress"
```
KrauseFx: NOOOOO

orta: haha, you not a fan of the braceless code?

KrauseFx: Not really, it gets confusing easily once you add a second parameter and / or an inline if

```ruby
warn "hi", 3 if something 5
```

KrauseFx: I think it the long run it's better to just always use `(` for method calls

orta: I think those are cases for using them full stop, but that not using an awesome language feature because one day you might add an extra parenthese is a shame. I could imagine this code  using them, `declared_trivial = pr_title.include? "#trivial" || !has_app_changes` but I don't think that becomes particularly confusing to read because of the lack of `(` & `)`

PR #113, opened 2016-03-04
Renamed public_info to public_repo #trivial

KrauseFx:
[There isn't a GitHub access right called `public_info`]


orta:
I've fixed your build, can you please start making them pass Danger?

PR #114, opened 2016-03-04
Improved circle output #trivial

KrauseFx:
- Added `-` when showing inline
- Made use of variable when printing out in code block

### DIFF
--- a/lib/danger/commands/init.rb
+++ b/lib/danger/commands/init.rb
@@ -179,7 +179,7 @@ module Danger
     end
 
     def uses_circle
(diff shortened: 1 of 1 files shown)


lib/danger/commands/init.rb
```diff
@@ -192,7 +192,7 @@ def uses_circle
         ui.say "Add this to the bottom of your circle.yml file:"
         ui.say "test:".green
         ui.say "  post:".green
-        ui.say "    - bundle exec danger".green
+        ui.say "    #{danger}".green
```
orta: ace

Issue #108, opened 2016-02-29
Add support for enterprise github


dbgrandi:
Should be a few minor points to change.

- add support for setting a custom GH host
- make sure that host is passed into the Octokit config

jeroenvisser101:
- it'd also need to work with `danger local`, which includes some references to github.com too

orta:
@dbgrandi I want to pair on this sometime

dbgrandi:
Definitely. I've got `Octokit` working with our GH instance now.

dbgrandi:
Started here: https://github.com/danger/danger/pull/139/files

jeroenvisser101:
@dbgrandi can this be closed now that #139 is merged? :)

orta:
Yes

jeroenvisser101:
We're getting closer and closer to 1.0.0 :dancer:

PR #124, opened 2016-03-09
Add support for git:// remote urls

jeroenvisser101:
As highlighted in the hub documentation (https://hub.github.com/), hub
adds remotes with `git://` prefix, instead of what github.com uses
(`git@` as prefix).

Danger should now support both.

Along with that, the logic has been simplified with a regular expression for all types of urls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Added internal plugin system - KrauseFx
 * Refactored unit tests - KrauseFx
 * Fixed issue when PR Title or PR body is nil - KrauseFx
+* Added support for `git://`-prefixed url as remote - jeroenvisser101
 
 ## 0.5.2
 
--- a/lib/danger/ci_source/local_git_repo.rb
+++ b/lib/danger/ci_source/local_git_repo.rb
@@ -25,14 +25,9 @@ module Danger
         # get the remote URL
         remote = run_git "remote show origin -n | grep \"Fetch URL\" | cut -d ':' -f 2-"
         if remote
-          url = remote[0].strip
-          # deal with https://
-          if url.start_with? "https://github.com/"
-            self.repo_slug = url.gsub("https://github.com/", "").gsub(/.git$/, '')
-
-          # deal with SSH origin
-          elsif url =~ %r{^(git://)?(git@)?github\.com(:|/)}
-            self.repo_slug = url.gsub(%r{(git://)?(git@)?github\.com(:|/)}, "").gsub(/.git$/, '')
+          remote_url_matches = remote.first.chomp.match(%r{github\.com(:|/)(?<repo_slug>.+/.+?)(?:\.git)?$})
+          if !remote_url_matches.nil? and remote_url_matches["repo_slug"]
+            self.repo_slug = remote_url_matches["repo_slug"]
           else
             puts "Danger local requires a repository hosted on github."
           end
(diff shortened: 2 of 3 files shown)


jeroenvisser101:
My local test fail:

```
  1) Danger::Dangerfile keeps track of the original Dangerfile
     Failure/Error: expect(dm.defined_in_file).to be file.path

       expected #<String:70180931302200> => "/var/folders/8f/wnyyflm90jqdw6nygmv5z4yr0000gn/T/dangefile_tests20160309-28419-1gv8xnp"
            got #<String:70180931302880> => "/var/folders/8f/wnyyflm90jqdw6nygmv5z4yr0000gn/T/dangefile_tests20160309-28419-1gv8xnp"

       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     # ./spec/dangerfile_spec.rb:17:in `block (2 levels) in <top (required)>'
```

This change seems to fix it, but it doesn't occur on circle/travis. Any ideas?

``` diff
--- a/spec/dangerfile_spec.rb
+++ b/spec/dangerfile_spec.rb
@@ -14,7 +14,7 @@ describe Danger::Dangerfile do
     file = make_temp_file ""
     dm = Danger::Dangerfile.new
     dm.parse file.path
-    expect(dm.defined_in_file).to be file.path
+    expect(dm.defined_in_file).to eq file.path
   end

   it 'runs the ruby code inside the Dangerfile' do
```

orta:
Great - this looks perfect - @marcelofabri would you do the honours?

jeroenvisser101:
@orta, you know by any chance why that test is failing with me locally?

orta:
Hrm, that one is hard to say, it might really be about ruby internals,  it's the difference of comparing memory addresses vs comparing the actual data. Which Ruby version are you using?

jeroenvisser101:
@orta – I'm using Using ruby 2.3.0

jeroenvisser101:
It's indeed an issue with 2.3.0 working differently with this. I'll send a PR for that, since the change showed in the diff works on both versions.

jeroenvisser101:
@orta I just learned that `ssh://` urls are meant to be read-only. I still think this would be useful since it would still allow to use `danger local` on repositories that you just forked.

orta:
This adds `git`, not `ssh` though TBH, it doesn't matter as long as we can reliably work with the repo slug

jeroenvisser101:
Oops, that's what I meant, I've been mixing them up all the time. `git://` is read-only, but still reliably get's the repository slug

jeroenvisser101:
@orta – I was thinking however, that you could get the slug instead by matching it with a regex like this:
`%r{github\.com/(?<repo_slug>.+/.+?)(?:\.git)?$}`

In which case it would make things a lot easier to match the repo_slug. Also, this would be a good preparation for #108, as we could quite easily do other domains, too.

https://regex101.com/r/bQ0cA6/1

orta:
Yep, I think this is a good idea :+1:

jeroenvisser101:
I've changed the logic. I might've to mention this change in the changelog as well, since it's quite a big change.

orta:
Cool, makes sense, you'll need to rebase too :100:

jeroenvisser101:
@orta done!

orta:
Thanks!

marcelofabri:
:clap:

PR #126, opened 2016-03-10
Initial implementation of violation suppression

### LINKED ISSUE #118: Comment based fail suppression?
Looking a real example: https://github.com/artsy/eigen/pull/1236

In this case I _definitely_ know that this will fail the build, but I want this to be correct. What about if inside the message for the PR I wrote:

> Danger: Ignore "Developer Specific file shouldn't be changed"

Wherein it won't fail the build if a fail is found with the same string. 

Ideas / Opinions?

marcelofabri:
Fixes #118 

Next steps:
- [x] tests to validate if warnings/errors are really suppressed
- [x] CHANGELOG
- [x] Update README


jeroenvisser101:
Looks good :+1:

orta:
Yeah, awesome

spec/dangerfile_spec.rb
```diff
@@ -33,6 +33,31 @@ def make_temp_file(contents)
     end.to raise_error(Danger::DSLError)
   end
 
+  it 'respects ignored violations' do
+    code = "message 'A message'\n" +
+           "warn 'An ignored warning'\n" +
+           "warn 'A warning'\n" +
+           "fail 'An ignored error'\n" +
+           "fail 'An error'\n"
```
jeroenvisser101: I believe this is the only code style thing here, and it's also the only think keeping the red lights on, apart from this, should be good, right?

orta:
Yep @marcelofabri you can run `bundle exec rubocop -a` to have these fixes automatically - and it needs a rebase

marcelofabri:
I've added an entry to CHANGELOG and a simple explanation on README, but I feel that can be improved.

jeroenvisser101:
@marcelofabri what happens when the message to be ignored contains a double quote?

lib/danger/request_sources/github.rb
```diff
@@ -30,6 +30,13 @@ def markdown_parser
     def fetch_details
       self.pr_json = client.pull_request(ci_source.repo_slug, ci_source.pull_request_id)
       fetch_issue_details(self.pr_json)
+      self.ignored_violations = ignored_violations_from_pr(self.pr_json)
+    end
+
+    def ignored_violations_from_pr(pr_json)
+      pr_body = pr_json[:body]
+      return [] if pr_body.nil?
+      pr_body.chomp.scan(/>\s*danger\s*:\s*ignore\s*"(.*)"/i).flatten
```
jeroenvisser101: Should we rather use `"(.+}"` to make sure there's at leas one letter?

marcelofabri: I can change that, but it shouldn't matter unless someone triggers an empty warning/error

marcelofabri:
> @marcelofabri what happens when the message to be ignored contains a double quote?
We're screwed right now :grimacing: 
We could support scaping, but I don't know if it's worth

lib/danger/request_sources/github.rb
jeroenvisser101: Yeah, you're right, it should be fine like this :)

jeroenvisser101:
> We're screwed right now  
> We could support scaping, but I don't know if it's worth

Probably not, lets see how this works out!

jeroenvisser101:
LGTM :+1:

orta:
Me too, @jeroenvisser101 - would you do the honors of merging please

jeroenvisser101:
My pleasure!

marcelofabri:
:tada:

jeroenvisser101:
@orta totally, Dangerfile is -I guess- just supposed to point in directions and warn about specific things to make it easier for the reviewer, not to replace it.

PR #135, opened 2016-03-12
Use branches instead of commits

orta:
Using two SHAs for diff-ing means that any merges that occur on the diff are taken into account. We want to work specifically with branches, which can then remove the merges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ bundle exec danger init
 bundle exec danger
 ```
 
-This will look at your `Dangerfile` and update the pull request accordingly. While you are setting up Danger, you may want to use: `--verbose` for more debug information. 
+This will look at your `Dangerfile` and update the pull request accordingly. While you are setting up Danger, you may want to use: `--verbose` for more debug information.
 
 ## What happens?
 
@@ -51,7 +51,7 @@ Danger runs at the end of a CI build, she will execute a `Dangerfile`. This file
 :sparkles: | `lines_of_code` | The total amount of lines of code in the diff
 :pencil2:  | `files_modified` |  The list of files modified
 :ship: | `files_added` | The list of files added
-:recycle: | `files_removed` | The list of files removed
+:recycle: | `files_deleted` | The list of files removed
 :abc:  | `pr_title` | The title of the PR
 :book:  | `pr_body` | The body of the PR
 :busts_in_silhouette:  | `pr_author` | The author who submitted the PR
--- a/lib/danger/dangerfile.rb
+++ b/lib/danger/dangerfile.rb
@@ -36,8 +36,8 @@ module Danger
         rows << [key.to_s, value]
       end
 
-      pr_copy = pr_body[0..80]
-      pr_copy += "..." if pr_body.length == 81
+      pr_copy = pr_body[0..79]
+      pr_copy += "..." if pr_copy.length == 80
       rows << ["PR Body", pr_copy]
 
       rows << ["---", "---"]
--- a/lib/danger/dangerfile.rb
+++ b/lib/danger/dangerfile.rb
@@ -40,8 +40,6 @@ module Danger
       rows << ["SCM", env.scm.class]
       rows << ["Source", env.ci_source.class]
       rows << ["Requests", env.request_source.class]
-      rows << ["Base Ref", env.ci_source.base_ref]
-      rows << ["Head Ref", env.ci_source.head_ref]
 
       params = {}
       params[:rows] = rows.each { |current| current[0] = current[0].yellow }
--- a/lib/danger/dangerfile.rb
+++ b/lib/danger/dangerfile.rb
@@ -44,8 +44,8 @@ module Danger
       rows << ["SCM", env.scm.class]
       rows << ["Source", env.ci_source.class]
       rows << ["Requests", env.request_source.class]
-      rows << ["Base Commit", env.scm.exec("--no-pager log #{env.danger_base_branch} -n1")]
-      rows << ["Head Commit", env.scm.exec("--no-pager log #{env.danger_head_branch} -n1")]
+      rows << ["Base Commit", env.meta_info_for_base]
+      rows << ["Head Commit", env.meta_info_for_head]
 
       params = {}
       params[:rows] = rows.each { |current| current[0] = current[0].yellow }
--- a/lib/danger/dangerfile.rb
+++ b/lib/danger/dangerfile.rb
@@ -36,10 +36,16 @@ module Danger
         rows << [key.to_s, value]
       end
 
+      pr_copy = pr_body[0..80]
+      pr_copy += "..." if pr_body.length == 81
+      rows << ["PR Body", pr_copy]
+
       rows << ["---", "---"]
       rows << ["SCM", env.scm.class]
       rows << ["Source", env.ci_source.class]
       rows << ["Requests", env.request_source.class]
+      rows << ["Base Commit", env.scm.exec("--no-pager log #{env.danger_base_branch} -n1")]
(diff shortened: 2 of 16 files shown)


orta:
screen shot 2016-03-12 at 15 58 54

vs

screen shot 2016-03-12 at 15 38 48

jeroenvisser101:
@orta – in the last 2 images, the CHANGELOG.md is not a entry, could it be that that hasn't been pushed yet? Or is it really the diff that's not working?

orta:
Yeah, there's a lot going on here.

I need to do some more digging into Grit now, I think the difference lies in how we create the diff

orta:
OK, using ruby-git, I can get the exact match: 

screen shot 2016-03-12 at 5 24 23 pm

orta:
screen shot 2016-03-12 at 5 56 05 pm

vs

screen shot 2016-03-12 at 5 55 42 pm

jeroenvisser101:
Oh that's cool! Any reasons not to use ruby-git?

orta:
so far one, it doesn't allow you to execute arbitrary git commands
but that trades off against the pros of: it has less dependencies and it works.

jeroenvisser101:
> ...and it works.

That seems to be quite important, plus, we can (if needed) execute git commands by just executing shell commands, right?

orta:
> we can (if needed) execute git commands by just executing shell commands, right?

:tada: yep - that's the plan

jeroenvisser101:
@orta it's not noticing that you added the changelog, is it?

orta:
Specifically for Circle CI - the commit that introduces those changes is the first commit in the PR, so I wonder if there's something to do with that.

jeroenvisser101:
@orta shouldn't be the case, since it runs on the last commit and it takes all of them, but it could be, weird. Also really weird that travis isn't failing.

jeroenvisser101:
@orta can you retrigger the build to see if that fixes it?

orta:
Shouldn't, but this has been happening on a bunch of builds, specifically on circle, gonna add some info to the `--verbose` output which should help. Also, triggered.

jeroenvisser101:
Thanks, some extra info would make debugging it a lot easier indeed. Right now, I've my fingers crossed for the (extremely slow) CircleCI build :)

jeroenvisser101:
The diff on Circle is definitely wrong

lib/danger/dangerfile.rb
```diff
@@ -35,13 +35,13 @@ def print_known_info
 
         rows << [key.to_s, value]
       end
-
+      rows << ["PR Body", pr_body.scan(/.{80}/)]
```
jeroenvisser101: We'd have to do `scan(...).first` here

jeroenvisser101: Or `pr_body[0..80]`

jeroenvisser101:
But confirmed: CircleCI takes a different base commit

jeroenvisser101:
The 'base' commit in Circle is rather the first commit of the branch, not the base, whereas travis actually uses the commit the branch was based upon.

lib/danger/dangerfile.rb
orta: I was hoping I could chuck in an array as lines and it would wrap, but it looks like it will just try and go on forever. Agree, switching to 80 char preview.

jeroenvisser101:
Which is odd, considering it both comes from the GH API, right? ([here?](https://github.com/danger/danger/blob/master/lib/danger/request_sources/github.rb#L47-L53)

orta:
> Which is odd, considering it both comes from the GH API, right? (here?)

Yes, this is odd.

lib/danger/dangerfile.rb
jeroenvisser101: You could do `scan(/.{80}/).join("\n")` or something like that?

orta: No, we tried that first

jeroenvisser101: Oh, do bad :(

orta:
The API gives us the right SHA from which to work from, so I'm just gonna see what happens if we go back.

screen shot 2016-03-12 at 7 08 27 pm

orta:
OK, this is consistent - I'm calling it done. 

@jeroenvisser101 you're welcome to review / merge now

lib/danger/dangerfile.rb
```diff
@@ -36,12 +36,16 @@ def print_known_info
...
+      rows << ["PR Body", pr_copy]
+
       rows << ["---", "---"]
       rows << ["SCM", env.scm.class]
       rows << ["Source", env.ci_source.class]
       rows << ["Requests", env.request_source.class]
-      rows << ["Base Commit", env.request_source.base_commit]
-      rows << ["Head Commit", env.request_source.head_commit]
+      rows << ["Base Commit", env.scm.exec("--no-pager log #{env.danger_base_branch} -n1")]
+      rows << ["Head Commit", env.scm.exec("--no-pager log #{env.danger_head_branch} -n1")]
```
marcelofabri: shouldn't these be methods on `scm` to make it easier to add other sources later?

orta: yes :+1:

lib/danger/environment_manager.rb
```diff
@@ -32,5 +32,26 @@ def fill_environment_vars
...
+      # As this currently just works with GitHub, we can use a github specific feature here:
+      pull_id = ci_source.pull_request_id
+      test_branch = request_source.base_commit
+
+      # Next, we want to ensure that we have a version of the current branch that at a know location
+      scm.exec "branch #{danger_base_branch} #{test_branch}"
+
+      # OK, so we want to ensure that we have a known head branch, this will always represent
+      # the head ( e.g. the most recent commit that will be merged. )
+      scm.exec "fetch origin +refs/pull/#{pull_id}/merge:#{danger_head_branch}"
```
marcelofabri: So we create local branches to make sure we have the right commits? Should we delete this branch after we're done?

marcelofabri:
This is awesome, @orta :tada:

lib/danger/environment_manager.rb
orta: agreed, added a clean up function

orta:
OK!

README.md
```diff
@@ -51,7 +51,7 @@ Danger runs at the end of a CI build, she will execute a `Dangerfile`. This file
 :sparkles: | `lines_of_code` | The total amount of lines of code in the diff
 :pencil2:  | `files_modified` |  The list of files modified
 :ship: | `files_added` | The list of files added
-:recycle: | `files_removed` | The list of files removed
+:recycle: | `files_deleted` | The list of files removed
```
jeroenvisser101: I think this needs a changelog entry

lib/danger/dangerfile.rb
```diff
@@ -36,12 +36,16 @@ def print_known_info
         rows << [key.to_s, value]
       end
 
+      pr_copy = pr_body[0..80]
+      pr_copy += "..." if pr_body.length == 81
```
jeroenvisser101: Shouldn't this be `< 80`?

orta: 0..80 is 81 chars

jeroenvisser101: Alright, but then you should check the `pr_copy.length`?

README.md
orta: This isn't me changing the name of it - this is me updating the READMe to the current one.

jeroenvisser101: Oh, you're right! My bad :)

orta:
Cool, address-ed

jeroenvisser101:
Looks good to me! :shipit:?

orta:
MRW I get a new job and my boss asks me how I'm doing

PR #144, opened 2016-03-13
Show full PR body, wrapping at 80 characters #trivial

jeroenvisser101:
This will allow us to see it, without messing up the layout in CI.

Long text for demo:
Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.

### DIFF
--- a/lib/danger/dangerfile.rb
+++ b/lib/danger/dangerfile.rb
@@ -26,8 +26,8 @@ module Danger
       rows = []
 
       AvailableValues.all.each do |key|
-        next if key == :pr_body
         value = self.send(key)
+        value = value.scan(/.{,80}/).to_a.each(&:strip!).join("\n") if key == :pr_body
 
         # So that we either have one value per row
         # or we have [] for an empty array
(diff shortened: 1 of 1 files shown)


orta:
screen shot 2016-03-13 at 17 08 59


:tada:

jeroenvisser101:
It does break in the middle of the word, but better than nothing, right? :tada:

jeroenvisser101:
@orta I noticed we use 'key.to_s' everywhere, except for the `pr_body`, should I rename it now?

dbgrandi:
:rocket:

Issue #131, opened 2016-03-10
Danger Plugins as Gems


orta:
They should have two major abilities:

- Provide DSL attributes
- Support providing it's own linting, if requested by a Dangerfile, with the current context variables provided by the system.

What does this look like?

```  Ruby
# Gemfile
gem 'danger'
gem 'danger-jekyll'
```

```  Ruby
# Dangerfile
[... usual stuff specific to a project]

# `danger-jekyll` exposes a DSL attribute for `validate_prose`
validate_prose "_site/pages/*.md"

# runs any plugin linting
plugin 'jekyll', { :project => "Artsy" }
```

This isn't too new, CocoaPods does all this stuff, and we spend a long time figuring out how this should work.

TODOs for this issue to be done:

- [ ] Adds plugin DSL attributes
- [x] Adds plugin linter hooks, with optional hash support
- [ ] Documentation
- [ ] `danger plugin create` - generates a stubbed plugin, see [cocoapods-plugins](https://github.com/CocoaPods/cocoapods-plugins/blob/master/lib/pod/command/plugins/create.rb) - this can be expanded upon in a different issue if needed.

jeroenvisser101:
Oh that'd be cool! I was thinking about allowing custom templates for the comment, but that could totally be possible with a plugin, too!

orta:
One of the advantages of this, is that you can have org-wide Dangerfiles without duplication, so `danger-cocoapods` could validate rubocop how we like it, and the same for specific types of CHANGELOG styles. We put it all in a gem, it is done on every build. Ace.

jeroenvisser101:
Plus: you could have plugins for framework specific stuff, such as Rails checking if tests are written when certain files change, or if you ran `bundle install` after changing your `Gemfile`.

jeroenvisser101:
There seems to already be some logic for plugins, but they can now only exist within the danger repository.

- [`protected_files.rb`](https://github.com/danger/danger/blob/99425352b62bbff7f559448234c72896ecab351f/lib/danger/plugins/protect_files.rb)
- [`def load_plugins`](https://github.com/danger/danger/blob/60a72be4907bb8011e00df4e3bb29d15818b30a7/lib/danger/dangerfile_dsl.rb#L19-L23)

jeroenvisser101:
So then `plugin '...'` would register a plugin? And it's DSL would only be available after registering? Or would it auto-register itself if it's present in the `Gemfile`?

orta:
[...]

PR #156, opened 2016-03-24
[Ready] Remote plugins

KrauseFx:
Working on local and remote actions using simple Ruby classes (no gems)

**Open points:**

- [x] Update plugins to be classes and not methods
- [x] `danger new_plugin` command
- [x] Tests


KrauseFx:
This PR is ready for review. Circle is failing for a random bundler reason. 
I don't know why Travis fails with 

```
lib/assets/ActionTemplate.rb:4:28: E: unexpected token tLT
(Using Ruby 2.0 parser; configure using TargetRubyVersion parameter, under AllCops)
      class [[CLASS_NAME]] < Plugin
                           ^
lib/assets/ActionTemplate.rb:20:1: E: unexpected token kEND
(Using Ruby 2.0 parser; configure using TargetRubyVersion parameter, under AllCops)
end
^^^
44 files inspected, 2 offenses detected
RuboCop failed!
```

This file is excluded in `.rubocop.yml` and works locally. Any ideas?

orta:
@dbgrandi - I'll let you review etc 

I don't think `danger new_plugin` should exist though, it should be replaced by your `danger plugins create` is that close yet?

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 ## Master
 
+* Added support for local plugins
+* Added support for remote plugins
+* Added new `danger new_plugin` command to create plugins in the fastlane
 * Added new `modified_files.include?("rakelib/*_stats.rake") file globbing support - KrauseFx
+* Added printing of table summaries after running danger
```
dbgrandi: You left the backtick around `modified_files.include?` on the previous line open, so the following lines may show up as monospaced.

lib/assets/ActionTemplate.rb
```diff
@@ -0,0 +1,20 @@
...
+        def self.description
+          [
+            "Describe what this plugin does",
+            "and how the user can use it"
+          ].join(" ")
+        end
+      end
+    end
+  end
+end
```
dbgrandi: Should this be `ActionTemplate.rb.template` or something to emphasize that it's not just an `.rb` file?

lib/danger/dangerfile_dsl.rb
```diff
@@ -58,18 +91,38 @@ def message(message)
...
         if AvailableValues.request_source.include?(method_sym)
-          # Request Source
           return env.request_source.send(method_sym)
         end
 
-        raise "Unknown method '#{method_sym}', please check out the documentation for available variables".red
+        # Plugins
+        class_name = method_sym.to_s.danger_class
+        if Danger::Dangerfile::DSL.const_defined?(class_name)
+          plugin_ref = Danger::Dangerfile::DSL.const_get(class_name)
```
dbgrandi: Is it worth adding a check here to make sure the `plugin_ref` extends from `Plugin` and not just a freestanding method or something?

lib/danger.rb
```diff
@@ -3,6 +3,7 @@
 require "danger/environment_manager"
 require "danger/commands/runner"
 require "danger/available_values"
+require "danger/core_ext/string"
```
dbgrandi: Are these in a particular order? The ones below are all :abc:

lib/assets/ActionTemplate.rb
jeroenvisser101: or even rendering with erb?

dbgrandi:
Looks good! :telescope: 

I've still got a bunch to do on refactoring for the plugin manager.

lib/assets/ActionTemplate.rb
jeroenvisser101: (that might be a bit overkill though...)

KrauseFx: Yah, it's too much since we just replace one piece of the string

lib/danger/dangerfile_dsl.rb
KrauseFx: Yeah, great point, I'll add this! Also, I should add a `.run` method to the super class that raises an exception if you didn't implement it

lib/assets/ActionTemplate.rb
jeroenvisser101: yep, I thought the same after I actually checked that code :+1:

jeroenvisser101: however, the .tpl or .template would make it more obvious that it's a template. Also, since 'assets' is only templates, we could consider renaming that folder all together.

README.md
```diff
@@ -102,6 +102,32 @@ snapshots_url = build_log.match(%r{https://eigen-ci.s3.amazonaws.com/\d+/index.h
...
+
+To create a new plugin run
+
+```
+danger new_plugin
+```
+
+This will generate a new Ruby file which you can modify to fit your needs. 
+
+If you think your new plugin is useful for other developers too, feel free to submit a PR to danger by copying the action to `lib/danger/plugins`.
```
orta: Blocker; I don't want to see Danger getting filled with Actions the way fastlane does - people should be using gems for this stuff, not relying on us to keep everything working in one repo. Please remove this line.

jeroenvisser101: We could have one 'gem', `danger-useful` or similar, containing a bunch of useful (WIP, Not a contributor, No spec update) plugins that you could start with?

KrauseFx: @jeroenvisser101 Great idea, this would be a separate repo in the `danger` org, maybe `danger-essentials`? 
We should make it easy for new users to get started with `danger` fast

jeroenvisser101: :+1: essentials makes total sense!

jeroenvisser101: @KrauseFx – How would we add multiple plugin classes to one plugin?

orta: > We could have one 'gem', `danger-useful`

Terminology wise, I think `danger-core` would work here. If it's essentials, it should ship with danger. There's no point in over-abstracting too far yet.

The reason I call this a blocker is that I can see how much time & effort it takes felix to maintain fastlane, and that's got a big audience. Danger could have an even bigger audience, given it's usefulness in most dev communities. So, we want to make it _really, really_ easy for others to build on top of danger, and not _into_ danger. Specifically, so that we're not spending time on minutae of actions instead of building a better platform.

If it makes sense to use something in every danger file ( like `protect_files` does ) then it should be in with Danger, if we feel like it would be better for all of our provided actions to go into `danger-core` cool, but it needs to be all or nothing, and do we need that extra level of abstraction yet?

jeroenvisser101: We could limit `danger-{core,useful,essentials}` in scope by including guidelines in both the PR/Issue templates and `CONTRIBUTING.md` so people know what will be merged and what will have to be in a separate plugin/repo. Also, I think the help of having a very limited, but often-used plugins might help so much people get started with Danger.

> If it's essentials, it should ship with danger. There's no point in over-abstracting too far yet.

I do agree that this looks like over-abstracting for now, and we should hit that roadblock first before going around it.

orta: I'm cool with making that jump at some point, but I want everything from `message` to `protect_files` to be in it, so that the actual "core/essentials" api is just another plugin ( that is included by default )

Agreed re: templates, once I get back from siberia I'll have a think

KrauseFx: Alright, so basically we do leave the most important methods in `danger`, however we remove the comment about submitting PRs when this plugin is useful? 

If that's ok, I'll update the PR according to all the comments

lib/danger.rb
KrauseFx: That shouldn't really matter

CHANGELOG.md
KrauseFx: Nice catch :+1:

KrauseFx:
Okay, addressed all the feedback and updated the PR, thanks everyone.

@orta Do you want `danger new_plugin` to be changed to `danger plugins create`? If so, could you point me to the right direction on how to implement this?

KrauseFx:
@dbgrandi @jeroenvisser101 I'd appreciated another review of the changes :+1:

jeroenvisser101:
@KrauseFx LGTM! :rocket:

orta:
> Do you want danger new_plugin to be changed to danger plugins create

No, we'll have a real one coming in one of these days with #131, I'd recommend dropping it from this PR

KrauseFx:
@orta I should completely remove the `danger new_plugin` command? How are people supposed to create new local plugins? This is different from #131 as it's mostly used for local additions to danger

KrauseFx:
@orta just wanted to ask you about the last remaining open point https://github.com/danger/danger/pull/156#issuecomment-204472675

orta:
Keep it in, but don't be surprised if it gets removed once we can push gems properly

KrauseFx:
Alright, resolved merge conflicts - working on finishing things up now :+1:

KrauseFx:
I pushed a few more commits with more tests and documentation, they should be live soon, just waiting for https://twitter.com/githubstatus/status/717464295454470145.

Ready to merge from my side. Tests are passing :+1:

orta:
Let's do it! :+1:

KrauseFx:
Oh yeah, thanks for merging @orta :+1:

PR #154, opened 2016-03-18
Added new modified_files.include?("pattern*") method

KrauseFx:
Fixes https://github.com/danger/danger/issues/82

PR #159, opened 2016-03-29
Fixing gemspec warning #trivial

marcelofabri:
Let's use #154


KrauseFx:
Yeah, let's do it! Could you rebase so that the build is green, then we can merge :+1:

PR #161, opened 2016-04-04
Fixed danger local crash when there are no recent PRs #trivial

KrauseFx:
Show an actual error message if there haven't been any PRs yet

### DIFF
--- a/lib/danger/ci_source/local_git_repo.rb
+++ b/lib/danger/ci_source/local_git_repo.rb
@@ -36,13 +36,15 @@ module Danger
 
         # get the most recent PR merge
         pr_merge = run_git "log --since='2 weeks ago' --merges --oneline | grep \"Merge pull request\" | head -n 1".strip
-        if pr_merge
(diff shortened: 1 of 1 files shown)


jeroenvisser101:
@KrauseFx the local command will never work with repositories that use a rebase workflow, should we have a look at that some time? (Should I create a issue?)

jeroenvisser101:
Rest looks good! 👍

KrauseFx:
@jeroenvisser101 in this case it was just a new project I started working on, where I just haven't created a PR yet, but plan on doing so, since it's open source

orta:
👍🏾  good call

jeroenvisser101:
@KrauseFx Alright, but the issue that rebase workflows don't work persist, although it seems that GitHub can see which commit belonged to which branches, we should be able to retrieve that somewhere. I might have a look tonight and see if that's doable. Nevertheless, it's outside the scope of this PR :)

orta:
Cool, yeah, make an issue, that's an interesting problem

PR #162, opened 2016-04-04
Support for raw markdown as well

KrauseFx:
I'm working on a better fastlane / danger integration. I need a way to add markdown comments to a PR, to show inline links and in general have more control over the design and layout.

The UI is simple, you just use `markdown("### h3 here")` to print markdown below the table or as the only content.

By supporting raw markdown comments, we don't lock our users in our tables. Assume someone wants a little more customization, they can just use the `markdown` method, without having to fork and maintain `danger`.

What do you all think? I'll add some more tests and docs :+1:


marcelofabri:
I like this idea :sparkles:

KrauseFx:
Waiting for an ok from @orta :+1:

orta:
Yep, I like this :+1:

jeroenvisser101:
This is cool! 👍

KrauseFx:
Alright, thanks everybody, I'll add some more tests and docs :+1:

KrauseFx:
All green, here we go :rocket:

jeroenvisser101:
🎉

Issue #171, opened 2016-04-11
Is it possible to use danger within the Xcode bots CI?


Antondomashnev:
I'm not sure If I'm doing something wrong.
But my script trigger fails with the message
`Could not find a CI source`

Thanks.

orta:
No-one has worked on adding Xcode Bots support, I imagine you'll need to set it up like [jenkins](https://github.com/danger/danger/blob/master/lib/danger/ci_source/jenkins.rb)?

Antondomashnev:
Thanks for the quick reply.

Probably that's should be like jenkins, I'll try to take a look on it later.

KrauseFx:
I guess that's where @czechboy0 is an expert :rocket:

czechboy0:
It should be easy or super easy to get danger working with Xcode Server.

*Super easy* version:
If you're using [Buildasaur](https://github.com/czechboy0/Buildasaur), the automatically created bots for each PR already follow a scheme which contains both the repo name and the PR number. So all you need to do in the danger adapter is to read the bot name from the `XCS_BOT_NAME` and parse out the repo slug and the PR number. AFAIR danger takes care of the rest.

*Easy* version:
If you're **not** using Buildasaur and for some reason you manually create Bots for each PR, you would need to manually inject the environment variables in the prebuild/postbuild script with something like `export REPO_SLUG=danger/danger` and `export PR_NUMBER=17`, which you'd detect in the danger adapter. It's a stretch, tbh I think the first version using Buildasaur is just much simpler.

In both cases, you'd trigger danger from the prebuild script.

Antondomashnev:
@czechboy0 amazing, I'm using [Buildasaur](https://github.com/czechboy0/Buildasaur). 
I think I could do a pr to make danger support Xcode Server CI.

jeroenvisser101:
@Antondomashnev that'd be great!

orta:
fixed by #172

PR #172, opened 2016-04-11
Add xcode_server ci_source;

Antondomashnev:
Feature request from issue #171 
This will only work if use [Buildasaur](https://github.com/czechboy0/Buildasaur) to build PR on Xcode Server.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,13 +1,40 @@
-# Sometimes it's a README fix, or something like that - which isn't relevant for
-# including in a project's CHANGELOG for example
-declared_trivial = pr_title.include? "#trivial"
+# Sometimes its a README fix, or something like that - which isn't relevant for
+# including in a CHANGELOG for example
 
-# Make it more obvious that a PR is a work in progress and shouldn't be merged yet
-warn("PR is classed as Work in Progress") if pr_title.include? "[WIP]"
+has_app_changes = !modified_files.grep(/lib/).empty?
+has_test_changes = !modified_files.grep(/spec/).empty?
 
-# Warn when there is a big PR
-warn("Big PR") if lines_of_code > 500
+if ["KrauseFx", "orta"].include?(pr_author) == false
+  warn "Author @#{pr_author} is not a contributor"
 
-# Don't let testing shortcuts get into master by accident
-fail("fdescribe left in tests") if `grep -r fdescribe specs/`.length > 1
-fail("fit left in tests") if `grep -r "fit specs/ `.length > 1
+  if modified_files.include?("*.gemspec")
+    warn "External contributors should not modify gemspec files"
+  end
+end
+
+work_in_progress_warning
+
+if has_app_changes && !has_test_changes
+  warn "Tests were not updated"
+end
+
+if pr_body.length < 5
+  fail "Please provide a summary in the Pull Request description"
+end
+
+declared_trivial = pr_title.include?("#trivial") || !has_app_changes
+if !modified_files.include?("CHANGELOG.md") && !declared_trivial
+  fail "Please include a CHANGELOG entry. \nYou can find it at [CHANGELOG.md](https://github.com/danger/danger/blob/master/CHANGELOG.md)."
+end
+
+### Oddly enough, it's quite possible to do some testing of Danger, inside Danger
+### So, you can ignore these, if you're looking at the Dangerfile to get ideas.
+
+# If these are all empty something has gone wrong, better to raise it in a comment
+if modified_files.empty? && added_files.empty? && deleted_files.empty?
+  fail "This PR has no changes at all, this is likely a developer issue."
+end
+
+protect_files(path: "danger.gemspec",
+           message: ".gemspec modified",
+        fail_build: false)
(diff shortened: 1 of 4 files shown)


KrauseFx:
This is missing some docs

lib/danger/ci_source/xcode_server.rb
```diff
@@ -0,0 +1,22 @@
...
+module Danger
+  module CISource
+    class XcodeServer < CI
+      def self.validates?(env)
+        return !env["XCS_BOT_NAME"].nil?
+      end
+
+      def initialize(env)
+        bot_name = env["XCS_BOT_NAME"]
+        unless bot_name.nil?
```
jeroenvisser101: you might want to do `return if bot_name.nil?` instead here

Antondomashnev: @jeroenvisser101 thanks, that is make sense.

Dangerfile
```diff
@@ -1,40 +1,13 @@
...
-if modified_files.empty? && added_files.empty? && deleted_files.empty?
-  fail "This PR has no changes at all, this is likely a developer issue."
-end
-
-protect_files(path: "danger.gemspec",
-           message: ".gemspec modified",
-        fail_build: false)
+# Don't let testing shortcuts get into master by accident
+fail("fdescribe left in tests") if `grep -r fdescribe specs/`.length > 1
+fail("fit left in tests") if `grep -r "fit specs/ `.length > 1
```
orta: any chance you could undo this change?

lib/danger/ci_source/xcode_server.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      def initialize(env)
+        bot_name = env["XCS_BOT_NAME"]
+        return if bot_name.nil?
+
+        repo_matches = bot_name.match(/\[(.*?)\]/)
+        self.repo_slug = repo_matches[1] unless repo_matches.nil?
+        pull_request_id_matches = bot_name.match(/((?<=#)\d+)/)
+        self.pull_request_id = pull_request_id_matches[1] unless pull_request_id_matches.nil?
+      end
```
orta: :+1:

orta:
Looks good

Dangerfile
jeroenvisser101: Possibly we could add this version to a `example/xcode/Dangerfile` so it would serve as quick-start, and we'd do the same with other providers/integrations? But this change should be undone indeed.

Antondomashnev: Sure, my mistake sorry.

orta: This was @Antondomashnev very reasonably running `bin/danger init` to test - we could look in the command to see if you're actually in the Danger git repo and not make changes to the existing `Dangerfile`?

lib/danger/ci_source/xcode_server.rb
```diff
@@ -0,0 +1,21 @@
...
+    class XcodeServer < CI
+      def self.validates?(env)
+        return !env["XCS_BOT_NAME"].nil?
+      end
+
+      def initialize(env)
+        bot_name = env["XCS_BOT_NAME"]
+        return if bot_name.nil?
+
+        repo_matches = bot_name.match(/\[(.*?)\]/)
```
jeroenvisser101: this regex should probably be `\[(.+)\]`, since it may never be empty.

lib/danger/ci_source/xcode_server.rb
```diff
@@ -0,0 +1,21 @@
...
+        self.repo_slug = repo_matches[1] unless repo_matches.nil?
+        pull_request_id_matches = bot_name.match(/((?<=#)\d+)/)
```
jeroenvisser101: Do you need a positive lookbehind here? Wouldn't `/#(\d+)/` not work on it's own?

Dangerfile
jeroenvisser101: Or change our Dangerfile to warn when the `Dangerfile` is changed, and explain when this is allowed, and when it's not. Either way, we could have some safeguards there for sure!

orta: Our Dangerfile would be overwritten, and that's what'd get parsed - so it wouldn't be able to tell itself to fail :dancers: 

But I agree on the concept :+1:

jeroenvisser101: Shit, you're right haha, that'd be bad. Yeah, we could probably change `danger init` and check if `danger/danger` is in the output of `git remote -v` (or something similar)?

lib/danger/ci_source/xcode_server.rb
Antondomashnev: yes, you're right

lib/danger/ci_source/xcode_server.rb
Antondomashnev: not really, if the [Buildasaur](https://github.com/czechboy0/Buildasaur) pattern has only one `#`. I think we can omit positive lookbehind here

jeroenvisser101: 👍

jeroenvisser101:
Looks good to me!

orta:
Yep - same here, great job @Antondomashnev

czechboy0:
Great work! Just one piece of feedback:
Currently when someone uses Xcode Server but doesn't use Buildasaur, the validates function seems to return true in that case. I suggest you try to parse the bot name, and if it's there, also check that it parses into the two expected parts (a string and a number). Only then it should validate. 

You could even print a warning if someone does try to run danger on Xcode Server without Buildasaur, maybe even just a link to this issue would be helpful.

PR #174, opened 2016-04-13
Support file renaming

MatejBalantic:
Hey all,

First of all, thanks for a great open source contribution. We at SoundCloud just implemented this project into our iOS pipeline and we love it.

Here's what we've found however:

When a file is moved or renamed in the pull request, `modified_files` will not return the new file's name, but rather something in this format
`/path/to/OldFile.h => NewFile.h`

This comes from upstream call to `@diff.stats[:files].keys`.

Obviously it is impossible to read the file in this format, so if developer is relying on `modified_fiels` to return valid file paths, this may be considered as a bug.

Unit test should be pretty much self-explanatory.

### DIFF
--- a/lib/danger/scm_source/git_repo.rb
+++ b/lib/danger/scm_source/git_repo.rb
@@ -43,7 +43,7 @@ module Danger
     private
 
     def extract_new_file(git_file)
-      git_file.gsub(/(.*) => (.*)/, "\\2")
+      git_file.gsub(/{?(.*) => (.*)}?/, "\\2")
     end
   end
 end
--- a/lib/danger/scm_source/git_repo.rb
+++ b/lib/danger/scm_source/git_repo.rb
@@ -24,7 +24,8 @@ module Danger
     end
 
     def modified_files
-      Danger::FileList.new(@diff.stats[:files].keys)
+      files = @diff.stats[:files].keys.map(&method(:extract_new_file))
+      Danger::FileList.new(files)
     end
 
     def lines_of_code
@@ -37,6 +38,12 @@ module Danger
 
     def insertions
       @diff.insertions
+    end
+
+    private
+
+    def extract_new_file(git_file)
+      git_file.gsub(/(.*) => (.*)/, "\\2")
     end
   end
 end
--- a/spec/sources/git_spec.rb
+++ b/spec/sources/git_spec.rb
@@ -76,6 +76,27 @@ describe Danger::GitRepo do
       end
     end
 
+    it 'handles moved as expected' do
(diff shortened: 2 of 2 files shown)


orta:
Looks good, Danger will complain about a changelog but everything else seems 👍🏾  to me

MatejBalantic:
Thanks. Looks like the diff output for moved is different between git versions, I was kind of afraid of this. Let me find a more bulletproof regext

lib/danger/scm_source/git_repo.rb
```diff
@@ -38,5 +39,11 @@ def deletions
     def insertions
       @diff.insertions
     end
+
+    private
+
+    def extract_new_file(git_file)
+      git_file.gsub(/[{]*(.*) => (.*)[}]*/, "\\2")
```
jeroenvisser101: `[}]*` in this case is equal to using `}?`, right?

MatejBalantic: Good point, fixed that

MatejBalantic:
Looks like this might be a more complex issue than it seems. It appears that a version of GIT running on CI actually lists both old and a new file when moving files. Running this on OS X will list files as described above. I'll try to run this on a linux machine to get more insights

jeroenvisser101:
@MatejBalantic it's only CircleCI that's failing because of the tests, travis works fine, but is complaining because of the missing changelog

MatejBalantic:
@jeroenvisser101 I see, didn't realize travis also runs tests. I guess this different behavior between environments is still a problem that we need to address before this PR could be merged?

jeroenvisser101:
@MatejBalantic Yeah, that'd be needed to be have this mergeable. I may have some time tonight to look into cross OS issues if you need help with that?

MatejBalantic:
Hey @jeroenvisser101, it would be great if you can help, I can't reproduce this on any environment available to me. If you can get me raw diff dump I could work from there

jeroenvisser101:
@MatejBalantic can you send me the code you use to generate those diffs, so I try and see if I can get it to execute in Circle?

jeroenvisser101:
@MatejBalantic if I log `git_file` passed to the `extract_new_file` method, the output is this:

```
"newFile"
"oldFile"
```

Both are passed on their own, which might make it hard to see if they are moved or not.

orta:
Any ideas on what we want/can do here?

orta:
Alright, gonna call this idle for now and close the PR, I'm open to re-opening and continuing but it looks like this has tapered off 👍

PR #178, opened 2016-04-18
Message pluralisation #trivial

### LINKED ISSUE #177: Pluralise error messages at the end of a build when cannot set PR status
From this build: https://travis-ci.org/CocoaPods/CocoaPods/jobs/123843848

`Found 1 error(s) and I don't have write access to the PR set a PR status.`

jeroenvisser101:
Fixes #177

### DIFF
--- a/lib/danger/core_ext/string.rb
+++ b/lib/danger/core_ext/string.rb
@@ -3,6 +3,10 @@ class String
     split('_').collect!(&:capitalize).join
   end
 
+  def danger_pluralize(count)
+    "#{count} #{self}#{'s' unless count == 1}"
+  end
+
   def danger_underscore
     self.gsub(/::/, '/').
       gsub(/([A-Z]+)([A-Z][a-z])/, '\1_\2').
--- a/lib/danger/request_sources/github.rb
+++ b/lib/danger/request_sources/github.rb
@@ -129,7 +129,7 @@ module Danger
       # use a read-only GitHub account
       if errors.count > 0
         # We need to fail the actual build here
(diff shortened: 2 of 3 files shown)


KrauseFx:
I think we have this code on multiple parts already. Could we move this to the String class similar to what Rails does, and have a `danger_pluralize` method or similar?

jeroenvisser101:
@KrauseFx sure we can, I'll update it tonight, heading home now!

jeroenvisser101:
Actually, inflection in ActiveSupport is somewhat more complex than we need:

https://github.com/rails/rails/blob/861b70e92f4a1fc0e465ffcf2ee62680519c8f6f/activesupport/lib/active_support/inflector/methods.rb#L30

We could maybe create a helper method `pluralize(count, '1 warning', '%d warnings')` and use that, but I think that would maybe be a bit too much.

Another option is having a method that just adds a 's' when count doesn't equal 1, which would be correct in most cases.

KrauseFx:
Sorry, that's not what I meant. 

I mean something like
```ruby
class String
  def danger_plura_le_name(count)
     self + "s" # or whatever here
  end
end
```

jeroenvisser101:
Ah, I misunderstood, I'll do that :)

jeroenvisser101:
@KrauseFx I edited it a little, can you check if this is what you meant?

orta:
This is what I think he was referring to now 👍

jeroenvisser101:
Cool! :)

KrauseFx:
Perfect, thanks for updating! 👍

jeroenvisser101:
Happy to do so :)

PR #181, opened 2016-04-19
[Plugin] Clean up method_missing #trivial

segiddins:
Object#send will call method_missing internally if necessary

PR #182, opened 2016-04-20
Allow checking pr_body for #trivial tag

jeroenvisser101:
Was requested in https://github.com/danger/danger/pull/181#issuecomment-211677536.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -22,7 +22,7 @@ if pr_body.length < 5
   fail "Please provide a summary in the Pull Request description"
 end
 
-declared_trivial = pr_title.include?("#trivial") || !has_app_changes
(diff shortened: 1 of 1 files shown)


Dangerfile
```diff
@@ -22,7 +22,7 @@ if pr_body.length < 5
   fail "Please provide a summary in the Pull Request description"
 end
 
-declared_trivial = pr_title.include?("#trivial") || !has_app_changes
+declared_trivial = (pr_title + pr_body).include?("#trivial") || !has_app_changes
```
orta: hah - that's a nice touch

jeroenvisser101: you like it? :)

orta: it's more elegant than I would have done, good job

Issue #134, opened 2016-03-12
Allow `danger local` to receive a PR reference


jeroenvisser101:
Instead of getting the `git log` and finding the last merge, in the case of a issue like #133, it'd help a lot if we could say `danger local --pr="my/repo#123"` to specify we'd want to test the local `Dangerfile` on a specific PR.

orta:
I agree, I think this makes sense.

jeroenvisser101:
Great! I might be able to take some time to make that work, I think it's fairly trivial :)

jeroenvisser101:
This seems to be a little harder than I initially thought, since we mix and match local git operations with GitHub API calls. Testing it on a specific repo would require you to have that repository locally, so `modified_files` is available.

orta:
ah, I understand, yeah, two options: 
* don't support `danger local` against repos that don't have the same set of commits ( forks are fine for example )
* see if there's github apis for the diff stuff

jeroenvisser101:
@orta I'd like the second the most, but that would also mean you'd have to have a GitHub API token for the `danger local` command

orta:
Yeah, that's not really doable IMO, which makes the former more reasonable

orta:
unless you can find a way to pull it out of a keychain or something?

jeroenvisser101:
Yeah or using [dotenv](https://github.com/bkeepers/dotenv) or something?

orta:
I mean like literal keychain.app

jeroenvisser101:
Oh, yeah, but that would make it Mac specific, having a `.env` (ignored) and `.env.sample` (as template) for `danger local` would allow people from all platforms to work with it

orta:
I think I can live with mac specific, having another dependency and all that setup doesn't really seem worth it for Danger `local` I, not sure I see people using it that much. 

Or maybe that's just how I work

jeroenvisser101:
It wouldn't be an issue for me, I'm on OS X. It would however be an issue for contributors using Linux/Unix/Windows, wanting to test locally.

Another option would be to create a alias locally (`alias danger_local='DANGER_GITHUB_TOKEN="..." bundle exec danger local'`)

orta:
Turns out I'm using `danger local` a lot now, and this is now something I want 👯

PR #195, opened 2016-05-04
Adds a command-line argument to support choosing a specific merged PR

orta:
I'm on the side of  we shouldn't support arbitrary repos e.g. `danger local --pr=artsy/eigen#11` as the file system isn't going to match the expectations of danger, a merged PR at least will have an approximation of the current file system ( though, obviously it may have changed since then, but that should be pretty obvious to the user. ) 

So this just adds a way to pass a single id along via `danger local --use-merged-pr=11`

Fixes https://github.com/danger/danger/issues/134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Master
 
-* Remove me - jeroenvisser101
+* Adds the ability to specify a PR number in `danger local` - orta
+* Ensures local branches are set up with  `danger local` - orta
 
 ## 0.7.3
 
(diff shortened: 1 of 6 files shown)


jeroenvisser101:
This needs documentation, but looks good!

orta:
Agreed, updated

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## Master
+## 0.7.4
```
jeroenvisser101: @orta we should add back # Master?

orta: already have done in the prepare for 0.7.4 stage

jeroenvisser101: Great stuff! :)

Issue #185, opened 2016-04-25
error packfile does not match index


KrauseFx:
```
fatal: Not a valid branch point: '865586f5d84122e04c205402831695f9fcd067c5'.
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
error: refs/remotes/origin/tvos_profiles does not point to a valid object!
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
error: packfile .git/objects/pack/pack-09e7a1fd28e5ceb3bdda92f31d5e8f415e28be98.pack does not match index
/Users/distiller/.fastlane_bundle/ruby/2.0.0/gems/git-1.3.0/lib/git/lib.rb:937:in `command':  (Danger::DSLError)
[!] Invalid `Dangerfile` file: git '--git-dir=/Users/distiller/fastlane/.git' '--work-tree=/Users/distiller/fastlane' diff '--numstat' 'danger_head' 'danger_base'  2>&1:fatal: ambiguous argument 'danger_base': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'. Updating the Danger gem might fix the issue.

 #  from Dangerfile:1
 #  -------------------------------------------
 >  warn("Big PR") if lines_of_code > 500
```

Via https://circleci.com/gh/fastlane/fastlane/1533

No changes from our setup, this just happened.

jeroenvisser101:
[...]

PR #201, opened 2016-05-04
Pass the verbose flag along from the runner command to the git scm

orta:
I don't have the answer to https://github.com/danger/danger/issues/185 - but I think this should help in debugging it.

### DIFF
--- a/lib/danger/environment_manager.rb
+++ b/lib/danger/environment_manager.rb
@@ -3,9 +3,11 @@ require "danger/request_sources/github"
 
 module Danger
   class EnvironmentManager
-    attr_accessor :ci_source, :request_source, :scm
+    attr_accessor :ci_source, :request_source, :scm, :verbose
 
-    def initialize(env)
+    def initialize(env, verbose:verbose = false)
+      self.verbose = verbose
+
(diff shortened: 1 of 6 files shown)


lib/danger/environment_manager.rb
```diff
@@ -3,9 +3,11 @@
 
 module Danger
   class EnvironmentManager
-    attr_accessor :ci_source, :request_source, :scm
+    attr_accessor :ci_source, :request_source, :scm, :verbose
+
+    def initialize(env, verbose:verbose = false)
```
segiddins: What's with the verbose after the `:`?

orta: named parameters, e.g. `e = Danger::EnvironmentManager.new(env, verbose: true)` writing `(env, true)` doesn't really give much context when reading

orta: or can you do them without that declaration?

segiddins: Should be `def initialize(env, verbose: false)`

orta:
Closing, this can be re-built sometime.

Issue #206, opened 2016-05-06
GHE does not seem to work


daniel-beard:
- I've been trying to point to a GHE instance, but even when specifying `DANGER_GITHUB_HOST` for the `local` action, it still tries to point to api.github.com rather than the GHE instance.
- Is this a known issue, or am I missing something here?
- Here's a stack trace, the 401 is because it's pointing to github.com rather than my GHE instance.

```
/Users/dbeard/.rvm/gems/ruby-2.2.4/gems/octokit-4.3.0/lib/octokit/response/raise_error.rb:16:in `on_complete': GET https://api.github.com/repos/myorg/myrepo/pulls/15130: 401 - Bad credentials // See: https://developer.github.com/v3 (Octokit::Unauthorized)
	from /Users/dbeard/.rvm/gems/ruby-2.2.4@global/gems/faraday-0.9.2/lib/faraday/response.rb:9:in `block in call'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4@global/gems/faraday-0.9.2/lib/faraday/response.rb:57:in `on_complete'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4@global/gems/faraday-0.9.2/lib/faraday/response.rb:8:in `call'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4/gems/octokit-4.3.0/lib/octokit/middleware/follow_redirects.rb:73:in `perform_with_redirection'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4/gems/octokit-4.3.0/lib/octokit/middleware/follow_redirects.rb:61:in `call'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4@global/gems/faraday-0.9.2/lib/faraday/rack_builder.rb:139:in `build_response'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4@global/gems/faraday-0.9.2/lib/faraday/connection.rb:377:in `run_request'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4@global/gems/faraday-0.9.2/lib/faraday/connection.rb:140:in `get'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4/gems/sawyer-0.7.0/lib/sawyer/agent.rb:94:in `call'
	from /Users/dbeard/.rvm/gems/ruby-2.2.4/gems/octokit-4.3.0/lib/octokit/connection.rb:154:in `request'
[...]

PR #207, opened 2016-05-06
Add section about GitHub Enterprise #trivial

jeroenvisser101:
Should help people with similar issues as described in #206.

@daniel-beard @dbgrandi can you confirm that with this documentation, one would be able to set this up?


daniel-beard:
Yes, that's what I needed to setup GHE. 👍

orta:
Cool beans

PR #216, opened 2016-05-29
initial work on migrating to new plugins

dbgrandi:
- disabled old style plugins
- create `Danger::Plugin`, that is a base class for plugins (look at https://github.com/dbgrandi/danger-proselint/blob/master/lib/danger_plugin.rb for what it should look like, but be aware that it has not been tested/integrated yet)
- added `init_plugins` method to `Dangerfile` that adds plugins as instance variables to a Dangerfile, allowing them to extend the DSL
- added some specs
- disabled some other specs

Rubocop will fail because some of the commented out specs have emoji, and it does not allow emoji in comments. This is known and should be okay.

Note: this will break all existing plugins.

fde5db0aef0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Considerable under-the-hood changes around the DSL, shouldn't affect end-user Dangerfiles though - orta
 * Add support for Drone CI = gabro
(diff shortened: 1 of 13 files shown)


orta:
Yep, all rubocop over the pizza emoji 👯

KrauseFx:
> Author @dbgrandi is not a contributor

Something we should change: https://github.com/danger/danger/pull/222

PR #217, opened 2016-05-29
Bring back the import Specs, and migrate the internal plugins to use the new API

orta:
Currently `protect_files` sits inside their own scope of `file`, my next PR will be digging into making plugins that support exposing root elements.


dbgrandi:
Looks good, except for the ❌'s

PR #218, opened 2016-05-30
Convert all of the public facing DSL attributes into Danger plugins

orta:
Converts all of the DSL modules into Danger Plugins. This is be really useful in enforcing great documentation, as well as making sure we're dog-fooding correctly.

WIP because all these new plugins need tests


lib/danger/danger_core/dangerfile.rb
```diff
@@ -21,10 +26,51 @@ def to_s
...
+        Danger::DangerfileMessagingPlugin,
+        Danger::DangerfileImportPlugin,
+        Danger::DangerfileGitHubPlugin,
+        Danger::DangerfileGitPlugin
+      ]
+    end
+
+    # http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-warn
+    # http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-fail
+
```
orta: I will add comment here

Rakefile
```diff
@@ -12,7 +12,7 @@ task default: :spec
 task :spec do
   Rake::Task['specs'].invoke
   Rake::Task['rubocop'].invoke
-  Rake::Task['spec_docs'].invoke
+  # Rake::Task['spec_docs'].invoke
```
orta: these rely on certain classes existing, which don't anymore

orta:
Woo! Green!

PR #222, opened 2016-05-30
Added @dbgrandi to Dangerfile #trivial

KrauseFx:
I heard he's a contributor


orta:
I just found this - https://developer.github.com/v3/orgs/members/#check-public-membership  was thinking of migrating to using that API instead 👍

PR #225, opened 2016-05-30
Messages don't need mapping when presenting status reports

orta:
I had thought they were violations, but they weren't. Types.` /shrug`. I should improve the tests around this.
#trivial


lib/danger/danger_core/plugins/dangerfile_messaging_plugin.rb
```diff
@@ -66,7 +66,7 @@ def status_report
         errors: @errors.map(&:message).clone.freeze,
         warnings: @warnings.map(&:message).clone.freeze,
         messages: @messages.map(&:message).clone.freeze,
-        markdowns: @markdowns.map(&:message).clone.freeze
+        markdowns: @markdowns.map.clone.freeze
```
segiddins: why still call `map` then?

orta: was Skyping Danger ( the person)  distracted

PR #234, opened 2016-06-01
Add ruby 2.3 support

### LINKED ISSUE #230: Tests fail on Ruby 2.3
When running `bundle exec rake spec` using Ruby 2.3 several tests fail with error:

```
      TypeError:
        can't create instance of singleton class
```

Failing code:
```ruby
    def refresh_plugins
      plugins = ObjectSpace.each_object(Class).select { |klass| klass < Danger::Plugin }
      plugins.map do |klass|
        plugin = klass.new(self)
        # ... 
      end
    end
```

Full output:
[https://gist.github.com/justMaku/fa813c5b879efa26803188843c35cbf9](url)

segiddins:
Closes #230

### DIFF
--- a/lib/danger/danger_core/dangerfile.rb
+++ b/lib/danger/danger_core/dangerfile.rb
@@ -83,7 +83,8 @@ module Danger
     # a reference to this Dangerfile
     def refresh_plugins
       plugins = ObjectSpace.each_object(Class).select { |klass| klass < Danger::Plugin }
-      plugins.map do |klass|
+      plugins.each do |klass|
+        next if klass.singleton_class?
         plugin = klass.new(self)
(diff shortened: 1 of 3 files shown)


lib/danger/danger_core/dangerfile.rb
```diff
@@ -83,7 +83,8 @@ def initialize(env_manager)
     # a reference to this Dangerfile
     def refresh_plugins
       plugins = ObjectSpace.each_object(Class).select { |klass| klass < Danger::Plugin }
-      plugins.map do |klass|
+      plugins.each do |klass|
+        next if klass.singleton_class?
```
orta: I should read up on this `singleton_class`  👍 thanks

PR #245, opened 2016-06-02
[Plugin] Don't use ObjectSpace to find plugins

segiddins:
#trivial Don't have to use ObjectSpace to find plugins, instead inherited will work better  - orta


orta:
This is a lovely approach 👍

orta:
Needs a rebase though, definitely a 👍 from me

PR #247, opened 2016-06-03
Supporting multiple Dangers

marcelofabri:
This allows multiple Danger instances to be run on the same PR by using the new `--dangerId` param. 

Previously this wasn't possible, so if you tried to run multiple instances you'd end up with only the last one comment.

A possible use case for this would be:

* `basic-danger` runs on a Linux environment (such as Circle CI) and validates the PR itself (title, etc)
* `compilation-danger` runs on a Mac after running unit tests for your iOS app and comments about warnings, test failures, etc
* `uitests-danger` runs on a Mac after running UI Unit tests and comments about test failures


orta:
OK, I'm good with this, but I'd like to see it be `danger_id` instead of `dangerId`, and I'd like a note in the README about this somewhere 👍

marcelofabri:
@orta Done!

orta:
Rockin' thanks

PR #254, opened 2016-06-06
Fix deleting old comment when using custom danger_id #trivial

marcelofabri:
Fixing a small bug that was introduced on #247 when using a custom danger_id 😬


justMaku:
Okay with me, if it passes CI I believe it should be safe to merge.

justMaku:
Okay, merged! Thanks for this fix 👍

PR #255, opened 2016-06-07
[Runner] more friendly error reporting via CLAide

segiddins:
We can eventually make it fancier if we want - #trivial


justMaku:
Perfect 👌🏻 Thanks :)

PR #256, opened 2016-06-07
Use inherited hook for CISource instead of enumerating symbols

justMaku:
Basically the same thing that was done for #221 #trivial


segiddins:
👍🏻

orta:
Rockin'

Issue #226, opened 2016-05-30
Allow CIs to request more details about an existing PR


orta:
CIs run against an arbitrary commit, of all the CI servers I've worked with only one will trigger a new build specifically on the creation of a PR, travis. 

![screen shot 2016-05-30 at 16 13 07](https://cloud.githubusercontent.com/assets/49038/15657714/774b5602-2681-11e6-8072-7e7a08fd2996.png)

In Travis you're guaranteed to have all the right ENV vars setup but only on that `/pr` build, the `/push` one is a _maybe_. 

For example, Circle. A build is triggered from a commit push, then  later it becomes connected to the PR, so you get builds like this: https://circleci.com/gh/artsy/eigen/2632

Which shows the PR connect on the right:
![screen shot 2016-05-30 at 16 21 12](https://cloud.githubusercontent.com/assets/49038/15657804/894c36d6-2682-11e6-8950-34c43e584d1a.png)

where it does't make the ENV vars correct 
![screen shot 2016-05-30 at 16 22 55](https://cloud.githubusercontent.com/assets/49038/15657834/cb1e7ef2-2682-11e6-828b-2c13eab4fdc1.png)

This makes Danger not check
![screen shot 2016-05-30 at 16 24 58](https://cloud.githubusercontent.com/assets/49038/15657857/1acb55a6-2683-11e6-9d8e-c3b0e6c72200.png)

---

Ideally there should be a way to look up, at runtime vs at CI build setup time to pull out the head/base commits.

orta:
https://github.com/danger/danger/pull/292 should help with this

orta:
I'm still seeing this - https://circleci.com/gh/artsy/eigen/3108

orta:
![screen shot 2016-07-26 at 12 10 38 pm](https://cloud.githubusercontent.com/assets/49038/17145849/02f3103a-532a-11e6-885c-2314e3c46165.png)


https://circleci.com/changelog/

might be fixed with this

kulte:
@orta Is it fixed by that? I have an associated PR, and no Danger comment?

orta:
Turns out, there are a few bugs on Circle's side:

![screen shot 2016-08-26 at 3 09 37 pm](https://cloud.githubusercontent.com/assets/49038/18017362/2a22c442-6b9f-11e6-999c-4b4275523cb0.png)

kulte:
[...]

PR #257, opened 2016-06-07
Fail CI Source validation early

justMaku:
There's no point in creating a CI Source only to dismiss it later on because it doesn't contain all of the information required. This PR makes sure we catch those situations early.


orta:
I think this is done so that we can support his kind of setup: https://github.com/danger/danger/issues/226

orta:
so this may not be a good idea

justMaku:
This doesn't check if the information provided is valid, it just checks if it exists. The example from Circle would still pass.

orta:
Awesome, cool, let's do it then 👍

Issue #264, opened 2016-06-10
Update GitHub PR status does not work


valeriomazzeo:
This method always fails because `latest_pr_commit_ref` has never been set.
Also, the generated error is absolutely misleading :(

```
def submit_pull_request_status!(warnings: nil, errors: nil, details_url: nil)
        status = (errors.count == 0 ? 'success' : 'failure')
        message = generate_github_description(warnings: warnings, errors: errors)
        client.create_status(ci_source.repo_slug, latest_pr_commit_ref, status, {
          description: message,
          context: "danger/danger",
          target_url: details_url
        })
      rescue
        # This usually means the user has no commit access to this repo
        # That's always the case for open source projects where you can only
        # use a read-only GitHub account
        if errors.count > 0
          # We need to fail the actual build here
          abort("\nDanger has failed this build. \nFound #{'error'.danger_pluralize(errors.count)} and I don't have write access to the PR set a PR status.")
        else
          puts message
        end
      end
```

I think this is very critical as it's a major loss in functionality. How come no one noticed before? : O

@orta, @KrauseFx, @segiddins

orta:
> I think this is very critical as it's a major loss in functionality. How come no one noticed before?

Could be fallout from https://github.com/danger/danger/pull/221 ?

However this build has been out for a few hours, so please use a less accusatory tone in issues. I'm afraid I don't have time to take a look ATM, but you ( or others ) are welcome to take a quick look at fixing it.  ( And ideally adding a unit test or two so we don't regress again 👍 )

valeriomazzeo:
Easy... I wasn't accusing anyone, just trying to help troubleshooting and pointing you guys to (possibly) the right direction.

>... How come no one noticed before?

It was more intended as: "is this not working only for me?"

I have reason to believe it wasn't working before either..

Unfortunately, I have no way to check where `latest_pr_commit_ref` was previously set without fiddling in the project history. I am not asking anyone to fix it right now, just an heads up...

orta:
> It was more intended as: "is this not working only for me?"

Sorry, I should be assuming [more positive intent](https://github.com/CocoaPods/CocoaPods/wiki/Communication-&-Design-Rules). 

We don't realize that it was broke because most of the projects using Danger are OSS, and we cannot give build access rights to the bot. Doing so would give anyone too much access to make changes via the access token :-/ - a limitation on the access rights for OAuth on GitHub alas.

marcelofabri:
[...]

PR #267, opened 2016-06-10
Fix commit status updating

justMaku:
Fix for #264, I'd say it's #trivial

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Master
+* Fix updating of the commit status after danger check. - justMaku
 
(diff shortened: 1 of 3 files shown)


segiddins:
Since it fixes a bug, there definitely should be a test + a CHANGELOG entry

valeriomazzeo:
Can I also suggest to change the error message 
abort("\nDanger has failed this build...

and print the exception instead?

orta:
> and print the exception instead?

This is the default behavior for most users, so I don't think we should be showing an exception, as it's not really unexpected.

PR #268, opened 2016-06-10
Assume the local HEAD is the PR's head commit

### LINKED ISSUE #266: Private repo with TeamCity issue: Permission denied (publickey)
Here is my setup:

* Private repo on GitHub
* TeamCity uses SSH key to clone repo on build machine
* Danger uses access token to check pull request

During the build I see following error messages:

```
Starting: /bin/sh /Users/teamcity/buildAgent/temp/agentTmp/build4061647261074985101.sh
in directory: /Users/teamcity/buildAgent/work/5b721f8dcbe1cd7a
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
Raising error 'Foo Error'

Danger has failed this build. 
Found 1 error and I don't have write access to the PR set a PR status.
Process exited with code 1
```

Despite _Found 1 error and I don't have write access to the PR set a PR status_ error message Danger was able to left a comment on pull request.

I've checked Danger source code and found [potential source of this problem](https://github.com/danger/danger/blob/master/lib/danger/request_source/github.rb#L49):

```ruby
      def setup_danger_branches
        # [ cut ]
        self.scm.exec "fetch origin +refs/pull/#{pull_id}/merge:#{EnvironmentManager.danger_head_branch}"
      end
```

Fetching origin like that means using `git` binary on a build machine without providing any auth information I specified in TeamCity (SSH key). Moreover, TeamCity always fetches most recent commit from the branch and I can't find any reason to fetch commits again.

To sum up:

* Incorrect error message _Found 1 error and I don't have write access to the PR set a PR status_. Danger actually has write access to the PR, she just don't have fetch access to the repo
* Fetching repo without any auth information

I'd like to discuss possible solutions to this issue :)

justMaku:
Resolves #266

Not really sure we need a changelog entry for that ¯\_(ツ)_/¯

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Master
 * Fix updating of the commit status after danger check. - justMaku
+* Relies on the current git HEAD, instead of pulling a merge branch from GitHub - justMaku
 
 ## 0.8.2
 
--- a/lib/danger/request_source/github.rb
+++ b/lib/danger/request_source/github.rb
@@ -38,15 +38,15 @@ module Danger
 
       def setup_danger_branches
         # we can use a github specific feature here:
-        pull_id = self.ci_source.pull_request_id
-        test_branch = self.pr_json[:base][:sha]
+        base_commit = self.pr_json[:base][:sha]
+        head_commit = self.scm.head_commit
 
         # Next, we want to ensure that we have a version of the current branch at a known location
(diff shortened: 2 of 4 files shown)


orta:
I think it's worth a CHANGELOG because it's a pretty big change in behavior, maybe something like

`- Relies on the current git HEAD, instead of pulling a merge branch from GitHub - justMaku` ?

justMaku:
Right! Updated :)

orta:
aye

Issue #249, opened 2016-06-05
Support a per-org Dangerfile via an `[org]/Danger` repo


orta:
*Problem:* Want to run the same logic on CocoaPods, Xcodeproj, CocoaPods-Core, CocoaPods-App, [etc etc]

*Current solution:* Make a `danger-cocoapods` plugin, refactor logic into that, make each one have a `Dangerfile` that calls something to trigger shared logic.

*Proposed solution:* `bundle exec danger` && `bundle exec danger local` checks to see if there is an existing repo on the org called `Danger` e.g. `CocoaPods/Danger` - if this exists, it clones ( or uses the [import](https://github.com/danger/danger/blob/master/lib/danger/danger_core/plugins/dangerfile_import_plugin.rb#L2) features to pull in a shared `Dangerfile` and run it's logic at the same time. This means we can have org-level consistent checks without even needing to have a `Dangerfile` in the current repo. You just need to set the ENV vars and run `bundle exec danger`.

paulofaria:
Might try this after WWDC if no one beats me to it.

orta:
And support for danger plugins - https://github.com/danger/danger/tree/master/danger_plugins

KrauseFx:
Working on this with https://github.com/danger/danger/pull/325 🚀

orta:
This is in!

KrauseFx:
Oh yeah :rocket:

PR #270, opened 2016-06-11
use claide-plugins to add plugin management commands to danger

dbgrandi:
This depends on https://github.com/dbgrandi/claide-plugins/pull/1 (and probably the publishing of that gem)

This would add:

`danger plugin create`
Uses the template at https://github.com/danger/danger-plugin-template to create a new danger plugin

`danger plugin list`
Pull from a JSON file to list all known danger plugins (I think @orta has been thinking about how to publish these automatically from rubygems updates)

`danger plugin search <foo>`
Pattern matches from the list of available plugins (uses that same JSON file)

It would also give us a place to add additional subcommands for linting/testing/documenting new plugins that people are making.


KrauseFx:
You might want to check out https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/plugins/plugin_search.rb in combination with https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/plugins/plugin_fetcher.rb, I built basically the same thing for `fastlane` 👍

orta:
Yeah, we discussed the idea of using Rubygems for the search, but ended up going with whitelisting - mainly because right now we're advising people use plugins as a way of sharing code. This would mean there'd be quite a lot of "only useful to me" plugins.

https://github.com/danger/danger/issues/249 might mitigate that somewhat - but I think it's enough to not warrant it

dbgrandi:
1755y3

orta:
This looks good to go on green to me 👍

PR #278, opened 2016-06-23
Release for 0.8.3

orta:
#trivial

### DIFF
--- a/danger.gemspec
+++ b/danger.gemspec
@@ -20,10 +20,10 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.0.0"
 
-  spec.add_runtime_dependency 'claide'
-  spec.add_runtime_dependency 'git'
+  spec.add_runtime_dependency 'claide', '~> 1.0'
+  spec.add_runtime_dependency 'git', '~> 0'
   spec.add_runtime_dependency 'colored', '~> 1.2'
-  spec.add_runtime_dependency 'faraday'
+  spec.add_runtime_dependency 'faraday', '~> 0'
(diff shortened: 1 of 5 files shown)


danger.gemspec
```diff
@@ -20,25 +20,25 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.0.0"
 
-  spec.add_runtime_dependency 'claide'
-  spec.add_runtime_dependency 'git'
+  spec.add_runtime_dependency 'claide', '~> 1.0'
+  spec.add_runtime_dependency 'git', '~> 0'
```
segiddins: `~> 0` does nothing

orta: that was the recommendation from `gem build` :-/ - switched them to `> 0`

orta: screen shot 2016-06-23 at 1 01 00 pm

danger.gemspec
```diff
@@ -31,14 +31,14 @@ Gem::Specification.new do |spec|
   
   spec.add_development_dependency "bundler", "~> 1.10"
   spec.add_development_dependency "rake", "~> 10.0"
-  spec.add_development_dependency "rspec", '~> 0'
-  spec.add_development_dependency "webmock", '~> 0'
-  spec.add_development_dependency "pry", '~> 0'
+  spec.add_development_dependency "rspec", '> 0'
```
segiddins: ... this also means exactly the same thing as specifying no requirements at all

danger.gemspec
segiddins: you should probably pin to a specific major version

PR #283, opened 2016-06-26
Parser + Plugin command improvements

orta:
* Adds more metadata for the site
* Initial stab at making a command or two for the parser too


lib/danger/plugin_support/plugin_parser.rb
```diff
@@ -5,29 +5,31 @@ class PluginParser
     attr_accessor :registry
 
     def initialize(path)
-      @path = path
+      raise "Path cannot be empty" if path.empty?
```
orta: Doing an empty path causes CI to take too long and fail

Rakefile
```diff
@@ -17,7 +17,7 @@ end
 
 desc 'Run RuboCop on the lib/specs directory'
 RuboCop::RakeTask.new(:rubocop) do |task|
-  task.patterns = ['lib/**/*.rb', 'spec/**/*.rb']
+  task.patterns = Dir.glob(['lib/**/*.rb', 'spec/**/*.rb']) - Dir.glob(['spec/fixtures/**/*', 'lib/danger/plugin_support/plugin_parser.rb'])
```
dbgrandi: why is `plugin_parser.rb` explicitly removed?

orta: All of the JSON parsing code there isn't passing, and every time I've made it pass it loses a lot of readability

orta: before https://github.com/danger/danger/blob/bffc246a11dac883d76fc6636319bd6c2acd58a3/lib/danger/plugin_support/plugin_parser.rb#L30

after
https://github.com/danger/danger/blob/0988bc7f6230ee280bdbd818780a7187bdbe5104/lib/danger/plugin_support/plugin_parser.rb#L31

lib/danger/plugin_support/plugin_parser.rb
```diff
@@ -50,20 +48,24 @@ def to_dict(classes)
...
+        attribute_meths = klass.attributes[:instance].values.map { |v| v.values }.flatten
+
         {
-        name: klass.name.to_s,
-        body_md: klass.docstring,
-        example_code: klass.tags.select { |t| t.tag_name == "example" }.map(&:text).compact,
-        attributes: klass.attributes[:instance].map do |pair|
-          {
-          pair.first => d_attr.call(pair.last)
+          name: klass.name.to_s,
```
dbgrandi: I think `name` is already a string.

dbgrandi:
Excellent stuff!

lib/danger/plugin_support/plugin_parser.rb
orta: This isn't a `Class` object - it's a Documentation object referring to the class

PR #292, opened 2016-07-01
Allow Circle CI to pass validation without a PR check

orta:
So that we can ask for the metadata a bit later.

I think I'd rather move the API check into the validation, so this is WIP until I get to a computer to do that.

/cc @ashfurrow

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 * Initial work on namespacing existing plugins - orta
 * Notify the user to add the new GitHub account as collaborator to Close Source project
 * Fixes a problem running `danger local` due to a missing dependency on yard - ashfurrow
+* Improvements for CircleCI CI detection - orta
+
 
 ## 0.8.3
 
--- a/lib/danger/ci_source/circle.rb
+++ b/lib/danger/ci_source/circle.rb
@@ -7,7 +7,6 @@ module Danger
     class CircleCI < CI
       def self.validates?(env)
         return false unless env["CIRCLE_BUILD_NUM"]
-        return false unless env["CI_PULL_REQUEST"]
         return false unless env["CIRCLE_PROJECT_USERNAME"]
         return false unless env["CIRCLE_PROJECT_REPONAME"]
 
--- a/spec/lib/danger/ci_sources/circle_spec.rb
+++ b/spec/lib/danger/ci_sources/circle_spec.rb
@@ -29,11 +29,11 @@ describe Danger::CISource::CircleCI do
     expect(t.pull_request_id).to be nil
   end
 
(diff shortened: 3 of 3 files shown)


ashfurrow:
💯

justMaku:
So how will Danger know which PR to update?

orta:
Circle is unique: https://github.com/danger/danger/blob/711d10037165625470079783a5caa7ce49ae7a89/lib/danger/ci_source/circle.rb#L20-L52

justMaku:
Right, sorry for introducing this then 😥

orta:
No worries 👍

orta:
Started moving to something more like this, it's probably better, but I have more important things to work on WRT Danger

``` ruby
# https://circleci.com/docs/environment-variables
require 'uri'
require 'danger/ci_source/circle_api'

module Danger
  module CISource
    class CircleCI < CI
      def self.validates?(env)
        return false unless env["CIRCLE_BUILD_NUM"]
        return false unless env["CIRCLE_PROJECT_USERNAME"]
        return false unless env["CIRCLE_PROJECT_REPONAME"]

        return true if env["CI_PULL_REQUEST"]

        # OK, so we may be in this state: https://github.com/danger/danger/issues/204
        # e.g. could be in a PR, but Circle didn't expose the PR number
        #
        url = pull_request_url(env)
        if url && URI.parse(url).path.split("/").count == 5
          paths = URI.parse(url).path.split("/")
          # The first one is an extra slash, ignore it
          env["CI_PULL_REQUEST"] = paths[4]
        end

        return !env["CI_PULL_REQUEST"].nil?
      end

      def self.fetch_pull_request_url(env, repo_slug, build_number)
        circle_token = env["CIRCLE_CI_API_TOKEN"]
        client = CircleAPI.new(circle_token)

        build_json = client.fetch_build(repo_slug, build_number)
        build_json[:pull_request_urls].first
      end

      def self.pull_request_url(env)
        if env["CIRCLE_PROJECT_USERNAME"] && env["CIRCLE_PROJECT_REPONAME"]
          repo_slug = env["CIRCLE_PROJECT_USERNAME"] + "/" + env["CIRCLE_PROJECT_REPONAME"]
          fetch_pull_request_url(env, repo_slug, env["CIRCLE_BUILD_NUM"])
        else
          nil
        end
      end

      def initialize(env)
        self.repo_url = GitRepo.new.origins # CircleCI doesn't provide a repo url env variable :/
        self.repo_slug = env["CIRCLE_PROJECT_USERNAME"] + "/" + env["CIRCLE_PROJECT_REPONAME"]
        self.pull_request_id = env["CI_PULL_REQUEST"]
      end

      def supported_request_sources
        @supported_request_sources ||= [Danger::RequestSources::GitHub]
      end

    end
  end
end
```

PR #293, opened 2016-07-01
Convert the Danger Dangerfile to the new syntax

orta:
Removes all the implicit plugin usage. Post 1.0 this is the normal syntax.


KrauseFx:
Me being a bot and deleting the branch 😉

PR #301, opened 2016-07-04
Enforce String Literal Styles

### LINKED ISSUE #300: Use consistent quotes
Currently we are specifying 

````
Style/StringLiterals:
  Enabled: false
```

which has resulted in inconsistent string literals all throughout the project. I suggest we enable this cop and update the project to use single quotes for all strings that do not require double quotes(e.g using string interpolation).

cc @orta

DanToml:
closes #300

\c @k0nserv


k0nserv:
Yay :tada: Thanks, I can't merge though that's a up for  @orta

orta:
Cool for me if green

orta:
looks like some emoji special casing

DanToml:
@orta Also looks like Rubocop has a bug in the string literal cop. Autocorrect works fine, but then fails lint :/. Will special case now and see if I can PR Rubocop.

orta:
cool, makes sense 👍

PR #307, opened 2016-07-05
Simplified and updated rubocop rules #trivial

KrauseFx:
Simplified and updated rubocop rules #trivial


KrauseFx:
cc @neonichu @orta 👍

lib/danger/plugin_support/plugin_parser.rb
```diff
@@ -1,36 +1,34 @@
...
+#       usable_methods = methods.select { |m| m.visibility == :public }.reject { |m| m.name == :initialize }
+#
+#
+#   the alternative, is to add
+#
+#       require 'pry'
+#       binding.pry
+#
+#   anywhere inside the source code below.
+#
```
orta: This change isn't wanted, you can't copy and paste these examples anymore

lib/danger/plugin_support/plugin_parser.rb
```diff
@@ -73,6 +71,7 @@ def to_json
       to_dict(plugins).to_json
     end
 
+    # rubocop:disable Metrics/AbcSize
```
orta: 👍  great

KrauseFx:
Updated the PR 🚀

KrauseFx:
👍

.rubocop.yml
```diff
@@ -108,19 +70,40 @@ Metrics/LineLength:
...
   Max: 18
 
-Style/DotPosition:
+# Sometimes it's easier to read without guards
+Style/GuardClause:
   Enabled: false
 
-Style/GuardClause:
+# We allow both " and '
+Style/StringLiterals:
```
DanToml: We want this to be enabled for consistency see https://github.com/danger/danger/issues/300

orta:
Woooooo

.rubocop_todo.yml
```diff
@@ -1,15 +0,0 @@
-# This configuration was generated by
-# `rubocop --auto-gen-config`
-# on 2016-07-04 18:33:07 -0400 using RuboCop version 0.38.0.
-# The point is for the user to remove these configuration records
-# one by one as the offenses are removed from the code base.
-# Note that changes in the inspected code, or installation of new
-# versions of RuboCop, may require this file to be generated again.
-#
-
-Style/StringLiterals:
```
DanToml: We need these to enable StringLiteral checking. (On CI, Rubocop doesn't detect emoji in strings as being special characters, but seems to locally :/)

DanToml:
@orta this just reverted #301 / #300 essentially too

orta:
aww

KrauseFx:
The reason why we should allow both `"` and `'` is to make it easier for non-Ruby developers. The difference between those 2 is not clear, and many people don't know that `#{...}` only works in combination with `"`, I'd also be in favour of `"` for this reason, as there is almost no performance difference between those 2.

orta:
I'd totally be on board with converting to always be double quotes

DanToml:
I'd be fine with having `"` everywhere. The perf difference in Danger won't be noticeable either way. But a nonsensical mix is more confusing to Non-Ruby developers. (because there's no rule at that point)

KrauseFx:
+1 on the `"` always, I should do the same for the `fastlane` code base

PR #310, opened 2016-07-06
Fix #309 Use BuildKite pull request repo environment variable

### LINKED ISSUE #309: BuildKite integration not working with latest BuildKite version when not building for pull request.
I have just been trying to setup Danger on an iOS project that builds on BuildKite. I am not sure if this has changed recently but BuildKite seems to be setting an environment variable with BUILDKITE_PULL_REQUEST=false when the build is not the result of a pull request.

This causes Danger to try the url GET https://api.github.com/repos/.../pulls/false which results in a 404.

I am not sure if BuildKite changed recently but this occurs with the latest version of the BuildKite agent running on OSX.

bentrengrove:
Hello all,

This fixes https://github.com/danger/danger/issues/309. It ensures that danger only runs on pull request builds and not just on any build by using the relevant environment variable.

I am going to be honest, I have no idea what I am doing really when it comes to ruby dev so hopefully this is somewhat correct. I wasn't sure if the empty? check was needed or not but put it in just to be safe. If its not required or there is a better way let me know.

Happy to make implement any suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * More documentation params are exposed to the linter - orta
 * Documentation audit - orta
 * Use proper commits for calculating diff.
+* Update environment variables used by Buildkite - bentrengrove
 
 ## 0.8.4
 
--- a/lib/danger/ci_source/buildkite.rb
+++ b/lib/danger/ci_source/buildkite.rb
@@ -6,14 +6,14 @@ module Danger
     class Buildkite < CI
       def self.validates?(env)
         return false unless env["BUILDKITE"]
-        return false unless env["BUILDKITE_REPO"]
+        return false unless env["BUILDKITE_PULL_REQUEST_REPO"] && !env["BUILDKITE_PULL_REQUEST_REPO"].empty?
         return false unless env["BUILDKITE_PULL_REQUEST"]
 
         return true
       end
 
       def initialize(env)
-        self.repo_url = env["BUILDKITE_REPO"]
+        self.repo_url = env["BUILDKITE_PULL_REQUEST_REPO"]
         self.pull_request_id = env["BUILDKITE_PULL_REQUEST"]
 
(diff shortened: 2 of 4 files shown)


bentrengrove:
Not sure how to fix that travis error. If someone could point me in the right direction, that would be great.

orta:
Hey - nice one, so this test - https://github.com/danger/danger/blob/master/spec/lib/danger/commands/runner_spec.rb#L10-L14 relies on the ordering of accessing the ENV vars to stub them out, I think you just need to change some of that?

And credit yourself in the changelog 💃

bentrengrove:
Sorry @orta I still can't work out what to change. The tests all pass locally.

I can see the test assumes the first CI build source is Buildkite but it looks like on the actual run it ends up trying circle? How does it work out the first build source and should I change the test to the environment variables for Circle instead?

orta:
I don't _strictly_ know (perhaps this makes this a really bad test :/ - b ut I had been struggling to write tests for that section, so `/shrug`), can you please try adding 

```diff
+      allow(ENV).to receive(:[]).with("BUILDKITE_PULL_REQUEST_REPO").and_return("git@github.com:artsy/eigen.git")

```

Twice, as it's accessed twice during startup - if this fails, I'll take over your PR and give it some time

bentrengrove:
No luck 😞

orta:
Looks like I'm wrong - thanks @bentrengrove can you make that test pending (change the `it` on that block to `pending` ) and I'll refactor the test in a separate PR

orta:
`:-/` maybe all in the runner :(

bentrengrove:
Did I disable them correctly?

orta:
Err, yeah,  OK, change those to `xit` which is what I was using for the other tests - sorry, I'm new rspec, only really start using it for Danger. In Specta/Kiwi that would have skipped it.

bentrengrove:
Yay! 🙌

orta:
Woooooo

orta:
Thanks! Sorry for the hassle!

PR #312, opened 2016-07-06
Move generic JSON parsing work into the parser instead of downstream

orta:
WIP, tests fail.

Mainly I wanted to be able to iterate faster on what the method should look like after some discussion on slack, https://dangergem.slack.com/archives/general/p1467726561000070


lib/danger/plugin_support/plugin_parser.rb
```diff
@@ -72,29 +72,56 @@ def to_json
...
+    end
+
+    def attribute_parser(attribute)
+      {
+        read: method_parser(attribute[:read]),
+        write: method_parser(attribute[:write])
+      }
+    end
+
+    def to_dict(classes)
```
segiddins: its `to_h` in ruby rypucally since the class is Hash

orta: yeah - that makes more sense - thanks!

PR #323, opened 2016-07-11
Add CI support for the Surf CI server

paulcbetts:
This PR adds CI support for [Surf](https://github.com/surf-build/surf), a toolkit for folx to create local build servers.

Refs https://github.com/surf-build/surf/pull/32, which adds Danger support for Surf (i.e. projects that have a Dangerfile will automatically be run if CI status will be posted)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,3 +1,10 @@
-gem install rubygems-bundler
 gem install bundler
 bundle exec rake spec
+
+if [-z "$SURF_BUILD_NAME" ]; then
+	## Posting to GitHub
+	bundle exec danger
+else
+	## Local clean build, just print to console
+	bundle exec danger local
+fi
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,3 @@
+gem install rubygems-bundler
+gem install bundler
+bundle exec rake spec
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Warn users not to store GitHub tokens in the repository -- dantoml
 * Crash on load fix for `danger plugins readme` -- orta
+* Add support for Surf CI (https://github.com/surf-build/surf) -- paulcbetts
 
 ## 0.8.5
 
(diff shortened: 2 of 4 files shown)


build.sh
```diff
@@ -0,0 +1,3 @@
+gem install rubygems-bundler
```
segiddins: This shouldn't be necessary since you use bundle exec

paulcbetts: Yeah, this part is a bit tricky to test, since its a chicken-and-egg scenario

orta: What is this file for? I've not seen a `build.sh` in other gem projects?

orta:
Aside form the question about the `build.sh` - this looks great 👍

build.sh
paulcbetts: @orta `build.sh` is one of the ways you can tell Surf how to build your project. I haven't added Ruby support yet (you can see the available platforms at https://github.com/surf-build/surf/tree/master/src/build-discoverers), so I had to add the explicit script

orta:
Alright, this looks cool - this is definitely a reminder that I need to start generating a list of things Danger supports programmatically ASAP in order to ensure the README/danger.systems is always up to date.

WRT to the `build.sh`, I'm not sure we're going to need to run our own Surf instance ( we are currently running Danger on Travis/Circle, as they are most used ( and Circle has custom code, due to some funky workarounds. ) so in the future, I may end up removing that file. However, this is a nice idea, and I'd like to see if people take it up. 

Surf looks nice, well done there.

paulcbetts:
@orta Surf isn't necessarily something you run (like a service), it's just as easy to use it locally:

```sh
npm install -g surf-build

## Man, I wish I had a clean build of what I'm working on locally
git push
surf-build
```

Or

```sh
## Man, I wish I had CI on Windows, or whatever
surf-client

## Open a new console window
git commit
git push      # CI runs in other window
```

orta:
Interesting 👍

paulcbetts:
Surf 0.9.2 is now up on npm and has full Danger support, thanks all!

PR #325, opened 2016-07-12
[Ready] Org-Wide Dangerfile #yolo

KrauseFx:
Ready 🚀 

See in action: https://github.com/Themoji/ios/pull/19 taken from https://github.com/Themoji/danger


lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
```diff
@@ -36,41 +36,53 @@ def self.instance_name
...
-    def import_url(url)
-      raise "URL is not https, for security reasons `danger` only supports encrypted requests" unless url.start_with?("https://")
+    # @param    [String] path
+    #           a local path or a https URL to the Ruby file to import
+    #           a danger plugin from.
+    # @return [String] The path to the downloaded Ruby file
+
+    def download(path_or_url)
+      raise "`download` requires a string" unless path_or_url.kind_of?(String)
+      raise "URL is not https, for security reasons `danger` only supports encrypted requests" unless path_or_url.start_with?("https://")
```
orta: 👍

lib/danger/request_source/github.rb
```diff
@@ -231,6 +231,10 @@ def process_markdown(violation)
         message = match.nil? ? html : match.captures.first
         Violation.new(message, violation.sticky)
       end
+
+      def organisation
```
orta: Probably worth putting in comments here about how it's used inside danger, otherwise it feels arbitrary

KrauseFx: Agreed - it's hard coded and not documented yet, since I wasn't sure if that's the right approac

KrauseFx:
It works for both `local` and on the CI, and was tested in action here: https://github.com/Themoji/ios/pull/19

I’ll do some more testing, and add support for both `danger` and `Danger` as repo name (ok?) and add tests before it’s ready to be merged :rocket:

orta:
> Add support for both danger and Danger as repo name (ok?) 

👍 Yep - good call

KrauseFx:
Add support for `Danger` repo https://github.com/danger/danger/pull/325/commits/1f903dbc5fc689716d42e8f7c38c56e3b9260c03

orta:
Any chance you can special case the Danger org, so that we don't get it running on here?

DanToml:
@orta @KrauseFx could also maybe use `danger-config` or have it passed as an argument? (To prevent conflicting with forks of danger)

KrauseFx:
@orta @DanToml Maybe to make it a more generic and automatic solution, how about I check if the current run is being made on a repo matching the `danger` repo of this org? This way it will work just fine for any org

DanToml:
@KrauseFx Would that handle if someone forked `danger/danger` to `Artsy/danger` for example?

KrauseFx:
@DanToml Yeah, I was thinking of just checking the repo name being equal to `danger`, and if it is don't import it

KrauseFx:
Alright, this is ready from my side, let me know if I can merge and prepare a release 🚀

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 
 * Warn users not to store GitHub tokens in the repository -- dantoml
 * Crash on load fix for `danger plugins readme` -- orta
+* Add support for org-wide Dangerfile -- KrauseFx
```
orta: this is a few releases back ( it's in the changelog for `0.9.0` )

orta:
OK, will give it a look later tonight

orta:
anyone know if rubocop can force the labeled params thing in ruby? I should do this, but I forget

segiddins:
> anyone know if rubocop can force the labeled params thing in ruby? I should do this, but I forget

can you be more specific?

lib/danger/commands/local.rb
```diff
@@ -64,11 +65,19 @@ def run
         dm.env.fill_environment_vars
         dm.env.ensure_danger_branches_are_setup
         dm.env.scm.diff_for_folder(".", from: Danger::EnvironmentManager.danger_base_branch, to: Danger::EnvironmentManager.danger_head_branch)
+
         dm.parse(Pathname.new(@dangerfile_path))
+
+        if dm.env.request_source.organisation && !dm.env.request_source.danger_repo? && (danger_repo = dm.env.request_source.fetch_danger_repo)
```
segiddins: maybe extract this into its own method? `infer_org_plugin`?

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
```diff
@@ -36,41 +36,53 @@ def self.instance_name
     #           a danger plugin from.
     # @return [void]
 
-    def import(path)
-      raise "`import` requires a string" unless path.kind_of?(String)
-      path += ".rb" unless path.end_with?(".rb")
+    def import(path_or_url)
+      raise "`import` requires a string" unless path_or_url.kind_of?(String)
```
segiddins: this should be `is_a` instead of `kind_of` in case someone is using a String subclass

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
```diff
@@ -36,41 +36,53 @@ def self.instance_name
...
-    # @return [void]
-    def import_url(url)
-      raise "URL is not https, for security reasons `danger` only supports encrypted requests" unless url.start_with?("https://")
+    # @param    [String] path
+    #           a local path or a https URL to the Ruby file to import
+    #           a danger plugin from.
+    # @return [String] The path to the downloaded Ruby file
+
+    def download(path_or_url)
+      raise "`download` requires a string" unless path_or_url.kind_of?(String)
```
segiddins: `is_a?`

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
segiddins: imho this should use `URI.parse` and check the `scheme`

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
```diff
@@ -36,41 +36,53 @@ def self.instance_name
...
+
+    # @!group Plugins
+    # Download a remote plugin and use it locally.
+    #
+    # @param    [String] url
+    #           https URL to the Ruby file to use
+    # @return [void]
+    def import_url(url)
+      path = download(url)
+      import_local(path)
```
segiddins: why can't this `eval` instead of writing to a temp file?

lib/danger/request_source/github.rb
```diff
@@ -231,6 +231,49 @@ def process_markdown(violation)
         message = match.nil? ? html : match.captures.first
         Violation.new(message, violation.sticky)
       end
+
+      # @return [String] The organisation name, is nil if it can't be detected
+      def organisation
+        matched = self.issue_json[:repository_url].match(%r{repos\/(.*)\/})
+        return matched[1] if matched
```
segiddins: you could do `matched && matched[1]` instead of both the return and the `nil` (which is already implicit, since an `if` statement not executed also returns `nil`)

lib/danger/request_source/github.rb
```diff
@@ -231,6 +231,49 @@ def process_markdown(violation)
...
+      rescue
+        nil
+      end
+
+      # @return [Hash] with the information about the repo
+      #   returns nil if the repo is not available
+      def fetch_repository(organisation: nil, repository: nil)
+        organisation ||= self.organisation
+        repository ||= self.ci_source.repo_slug.split("/").last
+        return self.client.repo("#{organisation}/#{repository}")
```
segiddins: `return` isn't necessary here, it's the last expression of the method

lib/danger/commands/local.rb
KrauseFx: Good point, I'll update the PR

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
KrauseFx: `kind_of?` and `is_a?` is the same as far as I know (http://stackoverflow.com/a/3893305/445598)

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
KrauseFx: I'm okay with `eval`, this is not new code, I just refactored it. @orta okay to change to `eval`?

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
orta: Rubocop tells us to be `kind_of?` - dunno why personally, but it's that for consistency

spec/lib/danger/import_spec.rb
```diff
@@ -38,7 +38,7 @@
       it 'downloads a remote .rb file' do
         expect { @dm.example_remote.echo }.to raise_error NoMethodError
 
-        url = 'https://krausefx.com/example_remote'
+        url = 'https://krausefx.com/example_remote.rb'
```
orta: This is a breaking change, and should go in the changelog

orta:
I mean:

```ruby
# fails
def method(one, two)
end

#passes
def method(one, two:two)
end
```

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
orta: Fine by me

orta:
Gonna take a quick stab at addressing feedback

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
orta: Eh, I take it back, I like that they both go through the same local code path

orta:
Alllllriiight, let's do it.

segiddins:
```ruby
# fails
def method(one, two)
end

#passes
def method(one, two:two)
end
```
the first takes no named parameters, the second takes one positional argument and a named argument with the label `two` with a default value of (the result of calling the method `two`)

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
segiddins: ah you're right

lib/danger/danger_core/plugins/dangerfile_import_plugin.rb
segiddins: It'd just be `eval(contents, TOP-LEVEL_BINDING, "#{plugin_name}.rb)`

orta:
Aye, I'd like a rubocop linter that forces labeled arguments 👯 👍

segiddins:
There doesn't exist one I know of, but writing one in a plugin shouldn't be too hard

orta:
Does it seem like a good idea to you?

segiddins:
Not particularly, especially since Ruby 2.0 doesn't support required keyword arguments

orta:
Interesting ( https://robots.thoughtbot.com/ruby-2-keyword-arguments ) - thanks @segiddins

spec/lib/danger/import_spec.rb
KrauseFx: Ahhh, great point!

lib/danger/request_source/github.rb
```diff
@@ -263,13 +262,13 @@ def fetch_danger_repo(organisation: nil)
         data = nil
         data ||= fetch_repository(organisation: organisation, repository: DANGER_REPO_NAME.downcase)
         data ||= fetch_repository(organisation: organisation, repository: DANGER_REPO_NAME.capitalize)
-        return data
+        data
```
KrauseFx: I actually prefer returns to be more explicit, but I'm fine with the change in this case

PR #336, opened 2016-07-14
Add urls to the classes for CI types

orta:
Mainly so that I can make the README / danger.systems use these programmatically to link to their respective pages.

#trivial

### DIFF
--- a/lib/danger/ci_source/local_git_repo.rb
+++ b/lib/danger/ci_source/local_git_repo.rb
@@ -5,6 +5,7 @@ require "uri"
 
 module Danger
   module CISource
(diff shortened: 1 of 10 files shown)


lib/danger/ci_source/local_git_repo.rb
```diff
@@ -5,6 +5,7 @@
 
 module Danger
   module CISource
+    # ignore
```
KrauseFx: So sad though 😢

orta: Poor old local git repo

PR #343, opened 2016-07-16
Update documentation for the core plugins 

orta:
* Fixes Danger.Systems
* Ensures that we won't ship broken plugins that could break it, need to move some of this into our linter too.


orta:
Finally.

Issue #198, opened 2016-05-04
Danger local should cache the returned JSON data from GitHub for a PR


orta:
It'd be a common experience to re-iterate again and again when making changes to your `Dangerfile`- currently `danger local` assumes you'll need to do it enough times that you can keep it under the unauthenticated GitHub API limit, but I've hit it once or twice. 

By caching the PR response, we can make less API calls in iterations.

KrauseFx:
Yeah, this is a great addition :+1:

orta:
This is in.

PR #344, opened 2016-07-16
add http caching on GitHub API calls when running danger local [#198]

dbgrandi:
- You can clear the cache with the `--clear-http-cache` command line arg.
- The cache TTL is 5 minutes

Still have 2 or 3 non-cached calls to GitHub, but it used to be around 8 when I was testing locally.

### DIFF
--- a/danger.gemspec
+++ b/danger.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "git", "~> 1"
   spec.add_runtime_dependency "colored", "~> 1.2"
   spec.add_runtime_dependency "faraday", "~> 0"
-  spec.add_runtime_dependency "faraday-http-cache"
+  spec.add_runtime_dependency "faraday-http-cache", "~> 1.0"
   spec.add_runtime_dependency "octokit", "~> 4.2"
   spec.add_runtime_dependency "redcarpet", "~> 3.3"
   spec.add_runtime_dependency "terminal-table", "~> 1"
--- a/danger.gemspec
+++ b/danger.gemspec
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "git", "~> 1"
   spec.add_runtime_dependency "colored", "~> 1.2"
   spec.add_runtime_dependency "faraday", "~> 0"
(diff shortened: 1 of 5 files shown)


dbgrandi:
(We should have a Danger rule that ensures a change to the `.gemspec` has a version restriction.)

orta:
We could use `gem build` and look for warnings

orta:
I'm doing something similar for YARD ATM

orta:
To make that work you need this spec to pass:

``` ruby
      it "a multiline warning with markdown, no errors" do
        warnings = violations(["a markdown [link to danger](https://github.com/danger/danger)\n\n```\ncode\n```\n\nHello"])
        result = @g.generate_comment(warnings: warnings, errors: [], messages: [])
        # rubocop:disable Metrics/LineLength
        expect(result.gsub(/\s+/, "")).to eq(
          '<table><thead><tr><thwidth="50"></th><thwidth="100%"data-kind="Warning">1Warning</th></tr></thead><tbody><tr><td>:warning:</td><tddata-sticky="false"><p>amarkdown<ahref="https://github.com/danger/danger">linktodanger</a></p><p><code>code</code></p><p>Hello</p></td></tr></tbody></table><palign="right"data-meta="generated_by_danger">Generatedby:no_entry_sign:<ahref="http://danger.systems/">danger</a></p>'
        )
        # rubocop:enable Metrics/LineLength
      End
```

danger.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "git", "~> 1"
   spec.add_runtime_dependency "colored", "~> 1.2"
   spec.add_runtime_dependency "faraday", "~> 0"
+  spec.add_runtime_dependency "faraday-http-cache", "~> 1.0"
```
KrauseFx: For our use case, wouldn't it be enough to just cache the GitHub JSON response ourself?

dbgrandi: I went this way to also get caching of GitHub API calls that are made from within the Dangerfile or plugins. Figured there is enough logic in how to properly cache that I didn't want to duplicate it.

orta:
👍

orta:
I've got the multi-line fix in #343

PR #348, opened 2016-07-23
Add `danger local --pry`

dbgrandi:
Drops into a Pry shell after evaluating the Dangerfile. Great way to explore how Danger works and what tools you have to use in your Dangerfile.

Removed a reference that created `@dangerfile_path` because it's a duplicate of logic in the parent class.

### DIFF
--- a/lib/danger/commands/local.rb
+++ b/lib/danger/commands/local.rb
@@ -12,18 +12,23 @@ module Danger
     def initialize(argv)
       @pr_num = argv.option("use-merged-pr")
       @clear_http_cache = argv.flag?("clear-http-cache", false)
-      @should_pry = argv.flag?("pry", false)
 
       super
 
-      if @should_pry && !@dangerfile_path.empty? && validate_pry_available
-        File.delete "_Dangerfile.tmp" if File.exist? "_Dangerfile.tmp"
-        FileUtils.cp @dangerfile_path, "_Dangerfile.tmp"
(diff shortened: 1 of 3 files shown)


lib/danger/commands/local.rb
```diff
@@ -9,16 +10,37 @@ class Local < Runner
     self.command = "local"
 
     def initialize(argv)
-      @dangerfile_path = "Dangerfile" if File.exist? "Dangerfile"
```
dbgrandi: This is duplicate logic from https://github.com/danger/danger/blob/master/lib/danger/commands/runner.rb#L29

lib/danger/commands/local.rb
```diff
@@ -9,16 +10,35 @@ class Local < Runner
...
       @pr_num = argv.option("use-merged-pr")
       @clear_http_cache = argv.flag?("clear-http-cache", false)
+      @should_pry = argv.flag?("pry", false)
+
       super
+
+      if @should_pry && !@dangerfile_path.empty? && validate_pry_available
```
DanToml: Would be nice to move this into a `setup_pry` method?

DanToml:
@dbgrandi Nice word, this will be super useful!

orta:
Great great, I poked dave as he built it so my feedback was I the last few commits. This is useful enough that I want a build now for it.

KrauseFx:
This will make the READMDE instructions on https://github.com/danger/danger#debugging unnecessary, right?

PR #353, opened 2016-07-23
Updated Faraday Builder class name - #trivial

jamtur01:
`Faraday::Builder` is now `Faraday::RackBuilder`.

This stops the error:

```Ruby
Faraday::Builder is now Faraday::RackBuilder
```

From appearing during Danger runs.


orta:
Nice one - thanks

PR #355, opened 2016-07-23
expose the unified diff for PR through the github plugin

champo:
Add a property to access Github's unified diff of the PR:

```
if pr_diff.string.include? "fdescribe"
  warn("Don't include an fdescribe in your tests")
end
```


orta:
Yep - perfect, looks 👍

PR #299, opened 2016-07-04
[WIP] Gitlab

k0nserv:
Opening this PR for tracking purposes and comments from everyone else. There's still some work that needs to be finished up

+ [x] Implement and adopt tests
+ [x] Make sure we get a new version of the `gitlab` gem issued 
+ [ ] For GitLab CI to work we need to convince/implement some changes in [GitLab](https://gitlab.com/gitlab-org/gitlab-ce/issues/14654)

cc @orta, @connorshea, @hjanuschka, @jk

PR #356, opened 2016-07-23
include file and line as possible arguments for messages and violations

champo:
This add the possibility to use `warn`, `message`, `markdown` and `fail` with `file` and `line`parameters like so:

```
warn("You shouldn't use puts in your Dangerfile", file: "Dangerfile", line: 10)
```

This doesn't actually use those parameters for anything useful yet.


orta:
An interesting approach, I hadn't considered this as a possibility. Nice work.

I think we'll need to add `markdown` to that list, while it has similarities with `message` - I bet a lot of people would try doing it. 

Note, we're in the process of supporting more than just GitHub ( #299 ) , which means this will have to go through the `api_source` ideally I think.

orta:
Also, want in on the org?

champo:
I avoided `markdown` because I thought the use case was different, but now that you mention it, it should also be supported.

By `api_source`, do you mean the [RequestSource](https://github.com/danger/danger/pull/299/files#diff-748540fe5943cd0355bdba814e9ada8dR7) classes? If you do, I've got a working version for Github's `RequestSource` with a few TODOs (mostly around making the markup reflect inline comments).

About the org, I do!

orta:
Yeah, sorry, `RequestSource` - so many names. Mainly so that there's a base pattern that others can follow to re-implement ( and maybe echo out if it's not GitHub, and just `puts` out an error message.

lib/danger/danger_core/messages/violation.rb
```diff
@@ -0,0 +1,21 @@
...
+    def to_s
+      extra = []
+      extra << "sticky: true" if sticky
+      extra << "file: #{file}" unless file.nil?
+      extra << "line: #{line}" unless line.nil?
+
+      "Violation #{message} { #{extra.join ', '} }"
+    end
+  end
+end
```
orta: Good move on these two

orta:
Alright, overall structure looks great - if we didn't have such comprehensively auto-generated documentation, I would merge this. ATM it would tell people something that's incorrect on http://danger.systems/reference.html 

Now that the 2.0 milestone is complete as of today - https://github.com/danger/danger/milestone/1  - I expect to be making the release soon, it could be today, it could be tomorrow - it's mostly blocked by work on the docs.

orta:
Depending on how fast you are working, I'd happy take you closing this PR and pushing it to a `develop` branch now that you have access, and would have the freedom to PR new code to that.

champo:
I've good a few more hours of work ahead, but I might not be able to sit down and tackle it until the next weekend. I'll see how to best organize the work then!

champo:
I'm going to close this PR and turn this onto a working branch.

On the other hand, I'm having trouble figuring out how to print markdowns that can't be actually set inline and have to fall back on the global comment. Right now the only option I'm considering is:

```
## <file> line <line>

<< your markdown here! >>
```

https://github.com/champo/phub-test/blob/d6118f0c1bfeb685d9cd67b078721655b46870e5/longfie#L23

I think it'd look kinda ugly and clunky. And if the user includes a heading in his markdown it'll probably confusing.

orta:
^ your example is a private repo ( or deleted? )

champo:
That was actually a stray copy paste! But I am testing on a toy private repo. I might make it public when submiting some of the most complicated parts of this.

PHUBUNIONBD (commented):
তৈরি করে দাও

PR #357, opened 2016-07-23
Clean up the README

orta:
Updates the README to only reflect setting up as a dev. All docs are now in the system.

@champo you've _just_ got set up, do you have any useful advice?

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,239 +8,57 @@ Formalize your Pull Request etiquette.
 -------
 
 <p align="center">
-    <a href="#getting-started">Getting Started</a> &bull;
-    <a href="#usage-on-ci">Usage</a> &bull;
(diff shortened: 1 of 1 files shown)


champo:
maybe something on how to fake a CI environment? `danger local` runs against a closed PR which isn't useful for testing things that need updating the PR as you test. I ended writing a small script that sets the env variables like it's travis.

orta:
This might make more sense as a feature? A way to let `local` run against PRs without relying on the merged PR aspect

champo:
ah, that makes a lot more of sense than writting it down! I've reported an issue on it https://github.com/danger/danger/issues/370. Also, it would be related to https://github.com/danger/danger/issues/368.

PR #359, opened 2016-07-23
Adding html_link to the GitHub plugin

### LINKED ISSUE #351: Add a GitHub html_link function
Look at this example from #350 :

```ruby
unlinted_plists = modified_files.grep(/Localizable.strings$/).reject { |file| system("plutil -lint #{file}") }
fail("#{unlinted_plists.join(", ")} failed to lint with plutil") if unlinted_plists.count > 0
```

Ideally, we want to have richer fail metadata specifically around linking to files. In most of my plugins I end up [writing](https://github.com/dbgrandi/danger-prose/blob/master/lib/danger_plugin.rb#L73) something like: 

``` ruby
github_loc = "/#{current_slug}/tree/#{github.branch_for_head}/#{path}"
message << "#### [#{path}](#{github_loc})\n\n"
```

Wherein what I really want to have happen is that my path string gets replaced by a clickable URL. This makes it easier to do more with Danger.

I'd propose a new method on Danger that takes either a `to_s`, or an array of `to_s`. If it's a string it converts it straight away, if it's an array it loops through them all converting them and adds an `&` at the end.

```ruby
thing = "file.json
github.html_link thing   # -> <a href='https://github.com/danger/danger/blob/my_pr/file.json'>file.json</a>


things = ["file.json", "other.txt", "sure.md"]
github.html_link things   # -> [<a href='https://github.com/danger/danger/blob/my_pr/file.json'>file.json</a>,
<a href='https://github.com/danger/danger/blob/my_pr/other.txt'> other.txt </a> &
<a href='https://github.com/danger/danger/blob/my_pr/sure.md'> sure.md </a>]
```

Then instead I would have wrote:

```ruby
unlinted_plists = modified_files.grep(/Localizable.strings$/).reject { |file| system("plutil -lint #{file}") }
fail("#{github.html_link unlinted_plists} failed to lint with plutil") if unlinted_plists.count > 0
```

Happy to change the name, was also considering `link_to` as I think that's what Rails does?

marcelofabri:
This fixes #351

### DIFF
--- a/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
@@ -104,7 +104,7 @@ module Danger
       pr_json[:head][:sha]
     end
 
-    # @!group GitHub Misca
+    # @!group GitHub Misc
     # The hash that represents the PR's JSON. For an example of what this looks like
     # see the [Danger Fixture'd one](https://raw.githubusercontent.com/danger/danger/master/spec/fixtures/pr_response.json).
     # @return [Hash]
@@ -127,6 +127,29 @@ module Danger
     # @return [String]
     def pr_diff
       @github.pr_diff
+    end
+
+    # @!group GitHub Misc
+    # Returns an HTML link for a file in the head repository. An example would be
+    # `<a href='https://github.com/artsy/eigen/blob/561827e46167077b5e53515b4b7349b8ae04610b/file.txt'>file.txt</a>`
+    # @return String
+    def html_link(paths)
+      paths = [paths] unless paths.kind_of?(Array)
+      commit = head_commit
+      repo = pr_json[:head][:repo][:html_url]
+      paths = paths.map do |path|
+        path_with_slash = "/#{path}" unless path.start_with? "/"
+        create_link("#{repo}/blob/#{commit}#{path_with_slash}", path)
+      end
+
+      return paths.first if paths.count < 2
+      paths.first(paths.count - 1).join(", ") + " & " + paths.last
+    end
+
+    private
+
+    def create_link(href, text)
+      "<a href='#{href}'>#{text}</a>"
     end
   end
 end
--- a/spec/lib/danger/plugins/dangerfile_github_plugin_spec.rb
+++ b/spec/lib/danger/plugins/dangerfile_github_plugin_spec.rb
@@ -1,0 +1,32 @@
+module Danger
+  describe DangerfileGitHubPlugin do
+    describe "dsl" do
+      before do
+        dm = testing_dangerfile
+        @dsl = DangerfileGitHubPlugin.new dm
+        pr_response = JSON.parse(fixture("pr_response"), symbolize_names: true)
+        allow(@dsl).to receive(:pr_json).and_return(pr_response)
+      end
+
+      describe "html_link" do
+        it "works with a single path" do
+          link = @dsl.html_link("file.txt")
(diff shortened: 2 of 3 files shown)


lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
```diff
@@ -104,7 +104,7 @@ def head_commit
       pr_json[:head][:sha]
     end
 
-    # @!group GitHub Misca
```
marcelofabri: I *think* this was a typo

spec/lib/danger/plugins/dangerfile_github_plugin_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+          link = @dsl.html_link(["file.txt", "example.json"])
+          expect(link).to eq("<a href='https://github.com/artsy/eigen/blob/561827e46167077b5e53515b4b7349b8ae04610b/file.txt'>file.txt</a> & " \
+                             "<a href='https://github.com/artsy/eigen/blob/561827e46167077b5e53515b4b7349b8ae04610b/example.json'>example.json</a>")
+        end
+
+        it "works with 3+ paths" do
+          link = @dsl.html_link(["file.txt", "example.json", "script.rb#L30"])
+          expect(link).to eq("<a href='https://github.com/artsy/eigen/blob/561827e46167077b5e53515b4b7349b8ae04610b/file.txt'>file.txt</a>, " \
+                             "<a href='https://github.com/artsy/eigen/blob/561827e46167077b5e53515b4b7349b8ae04610b/example.json'>example.json</a> & " \
+                             "<a href='https://github.com/artsy/eigen/blob/561827e46167077b5e53515b4b7349b8ae04610b/script.rb#L30'>script.rb#L30</a>")
```
marcelofabri: I don't know if we should strip the fragment part from visible text

orta: IMO no 👍

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
```diff
@@ -128,5 +128,28 @@ def api
...
     end
+
+    # @!group GitHub Misc
+    # Returns an HTML link for a file in the head repository. An example would be
+    # `<a href='https://github.com/artsy/eigen/blob/561827e46167077b5e53515b4b7349b8ae04610b/file.txt'>file.txt</a>`
+    # @return String
+    def html_link(paths)
+      paths = [paths] unless paths.kind_of?(Array)
+      commit = head_commit
+      repo = pr_json[:head][:repo][:html_url]
```
orta: we use quotes [in eigen](https://github.com/artsy/eigen/blob/master/Dangerfile#L96) - `  current_branch = env.request_source.pr_json["head"]["ref"]` - this might not work?

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
orta: looks like it, aye

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
marcelofabri: I think it should, since there're other places using symbols (this whole file, actually)

orta: haha - guess the underlying class is agnostic

orta:
Alright, I was already in there writing some examples - so I will merge this and add some examples to the plugin

PR #361, opened 2016-07-24
Improvements to the Parser and the Linter

### LINKED ISSUE #354: The linter's return void checker is wrong
When I run it against this file:

```ruby
module Danger
  # Run Ruby files through Rubocop.
  # Results are passed out as a table in markdown.
  #
  # @example Specifying custom config file.
  #
  #          rubocop.run
  #
  # @see  Moya/Aeryn
  # @tags ruby, rubocop, linter
  #
  class DangerRubocop < Plugin
    # Runs Ruby files through Rubocop. Generates a `markdown` list of warnings.
    #
    # @param   [String] files
    #          A globbed string which should return the files that you want to
    #          run through, defaults to nil. If nil, modified and added files
    #          from the diff will be used.
    # @return  [void]
    #
    def run(files = nil)
      files_to_lint = files ? Dir.glob(files) : (modified_files + added_files)
      files_to_lint.select! { |f| f.end_with? 'rb' }

      offending_files = rubocop(files_to_lint)
      return if offending_files.empty?

      markdown offenses_message(offending_files)
    end

    private

    def rubocop(files_to_lint)
      rubocop_results = files_to_lint.flat_map do |f|
        JSON.parse(`bundle exec rubocop -f json #{f}`)['files']
      end
      rubocop_results.select { |f| f['offenses'].count > 0 }
    end

    def offenses_message(offending_files)
      require 'terminal-table'

      message = "### Rubocop violations\n\n"
      message + Terminal::Table.new(
        headings: %w(File Line Reason),
        rows: offending_files.flat_map do |file|
          file['offenses'].map do |offense|
            [file['path'], offense['location']['line'], offense['message']]
          end
        end
      ).to_s
    end
  end
end
```

I get this
```sh
~/d/r/danger-rubocop (master) ⏛  bundle exec danger plugins lint

[!] Passed

Warnings
  - Return Type - run (method):
    - If the function has no useful return value, use ` @return  [void]` - this will be ignored by documentation generators.
    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L46

```

orta:
It's tough to provide some useful bits of information on danger.system - notably the Plugin <-> Gem  bits. As a gem can contain multiple plugins, and with just the raw class metadata it's not enough. Now each plugin links to a gem, and we have file metadata for all the pieces.

As the file/location data is now in the parser, I've exposed it to the linter, and fixed #354

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -31,12 +31,13 @@ if !git.modified_files.include?("CHANGELOG.md") && !declared_trivial
 end
 
 # Docs are critical, so let's re-run the docs part of the specs and show any issues:
-core_plugins = Dir.glob("lib/danger/danger_core/plugins/*.rb")
-core_plugins_docs = `bundle exec danger plugins lint #{core_plugins.join " "} --warnings-as-errors`
+core_plugins_docs = `bundle exec danger plugins lint lib/danger/danger_core/plugins/*.rb --warnings-as-errors`
 
-# If it failed, fail the build, and include markdown with the output error 
-if core_plugins_docs.include? "Failing due to"
-  markdown(core_plugins_docs)
+# If it failed, fail the build, and include markdown with the output error.
+unless $?.success?
+  # We want to strip ANSI colors for our markdown, and make paths relative
+  colourless_error = core_plugins_docs.gsub(/\e\[(\d+)(;\d+)*m/, "")
+  markdown("### Core Docs Errors \n\n#{colourless_error}")
   fail("Failing due to documentation issues, see below.", sticky: false)
 end
 
--- a/Dangerfile
+++ b/Dangerfile
@@ -30,6 +30,16 @@ if !git.modified_files.include?("CHANGELOG.md") && !declared_trivial
   fail("Please include a CHANGELOG entry. \nYou can find it at [CHANGELOG.md](https://github.com/danger/danger/blob/master/CHANGELOG.md).")
 end
 
+# Docs are critical, so let's re-run the docs part of the specs and show any issues:
+core_plugins = Dir.glob("lib/danger/danger_core/plugins/*.rb")
+core_plugins_docs = `bundle exec danger plugins lint #{core_plugins.join " "} --warnings-as-errors`
+
+# If it failed, fail the build, and include markdown with the output error 
+if core_plugins_docs.include? "Failing due to"
+  markdown(core_plugins_docs)
+  fail("Failing due to documentation issues, see below.", sticky: false)
+end
+
 # Oddly enough, it's quite possible to do some testing of Danger, inside Danger
 # So, you can ignore these, if you're looking at the Dangerfile to get ideas.
 #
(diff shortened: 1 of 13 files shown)


Dangerfile
```diff
@@ -30,6 +30,17 @@ if !git.modified_files.include?("CHANGELOG.md") && !declared_trivial
   fail("Please include a CHANGELOG entry. \nYou can find it at [CHANGELOG.md](https://github.com/danger/danger/blob/master/CHANGELOG.md).")
 end
 
+# Docs are critical, so let's re-run the docs part of the specs and show any issues:
+core_plugins = Dir.glob("lib/danger/danger_core/plugins/*.rb")
+core_plugins_docs = `bundle exec danger plugins lint #{core_plugins.join " "} --warnings-as-errors`
```
segiddins: since you're already invoking a shell here, there's no need to use `Dir.glob`

Rakefile
```diff
@@ -23,3 +23,4 @@ task :spec_docs do
   core_plugins = Dir.glob("lib/danger/danger_core/plugins/*.rb")
   sh "danger plugins lint #{core_plugins.join ' '}"
 end
+
```
segiddins: extra space

lib/danger/plugin_support/plugin_file_resolver.rb
```diff
@@ -17,26 +18,25 @@ def resolve_to_paths
       # When given a list of gems
       elsif @refs and @refs.kind_of? Array
         Bundler.with_clean_env do
-          Dir.mktmpdir do |dir|
-            gem_names = @refs
-            deps = gem_names.map { |name| Bundler::Dependency.new(name, ">= 0") }
+          # We don't use the block syntax as we want it to persist until the OS cleans it on reboot
+          # or whatever, it needs to persist outside this scope.
+          dir = Dir.mktmpdir
```
segiddins: you should use the block version of `mktmpdir` that will ensure the created directory is cleaned up

Dangerfile
orta: Ah yeah, good thinking

lib/danger/plugin_support/plugin_file_resolver.rb
orta: In this case I don't want to do that, as those files get used in the parser outside of this scope

segiddins: in that case, you should store the dir somewhere and make sure to unlink it when you're done using it

orta: I'd put it in another tmpdir, they go in a throwaway place, the OS will clean it up `/shrug` - it'd be little gain for some extra state

segiddins: The OS doesn't typically clean it up until restart -- so that could be days

PR #371, opened 2016-07-25
when outputing the summary, consider file and line

champo:
Print out the file and line on the summary if present.

```markdown
Results:

Warnings:
- [ ] Gemfile line 3: Recursive warn - inline false
- [ ] Gemfile line 1: Recursive warn - inline false
- [ ] Gemfile.lock line 15: Recursive warn 2 - inline false
- [ ] .gitignore line 10: ignore - inline false
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Master
 
+* add `file` and `line` optional parameters to methods on the messaging plugin
+
 ## 2.0
 
 * **BREAKING** Removes a lot of top-level attributes in the DSL, moving them into scoped plugins - orta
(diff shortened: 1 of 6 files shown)


orta:
Maybe it'll look nicer with GitHub link style files?

```
- [ ] Gemfile#L3 - Recursive warn - inline false
- [ ] Gemfile#L1 - Recursive warn - inline false
- [ ] Gemfile.lock#L15 - Recursive warn 2 - inline false
- [ ] .gitignore line#L10 - ignore - inline false
```

orta:
It's because your pledge of Danger allegiance is private, not public. Still not sure what to do about that. Maybe - https://github.com/Moya/Aeryn is the solution instead of Danger.

champo:
yeah, I think it will. Particularly since I'll be using that style for comments that fall back on the table (I'm adding a link to the file in the commit that produces the comment).

I didn't know it was private by default! It is now corrected.

orta:
👍 good to go by me

PR #373, opened 2016-07-25
show a link to the line when setting inline comments

champo:
Add a link to the file & line the message is for in GitHub. Like: `https://github.com/danger/danger/blob/744eefaf2f9338ef373e06d5156dda9d51375390/lib/danger/request_source/github.rb#L5`.

I've added a test around the parsing of this new messages. I'm not sure if more tests would be useful. @orta does it seem right to you?


orta:
Yeah, I think this is fine 👍

Issue #374, opened 2016-07-26
Default Dangerfile reported as "invalid"


epologee:
After going through all steps of the setup process (with the latest version), both CircleCI and `danger local` report that the Dangerfile is invalid. This is the default Dangerfile, as produced by the setup process.

```shell
/usr/local/opt/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/danger-2.0.0/lib/danger/danger_core/dangerfile.rb:60:in `method_missing':  (Danger::DSLError)
[!] Invalid `Dangerfile` file: undefined local variable or method `pr_title' for #<Danger::Dangerfile:0x007fefe21bca90>. Updating the Danger gem might fix the issue.
```

Commenting out the `pr_title` lines will just cause it to fail on the next Danger DSL method. I think the real error is something else, probably something small even, but I don't know how to debug this. It passed all the GitHub-related steps (repo access with the access token, etc.).

Any clues how to resolve this?

orta:
It's `github.pr_title` any idea where you found that docs from? If it's from `danger init` then this https://github.com/danger/danger/pull/372 will fix that, I will make a release so others don't hit the same problem!

epologee:
Yes, it was from `danger init`, after reading the manual (:wink:) at http://danger.systems/guides/getting_started.html

orta:
![screen shot 2016-07-26 at 10 27 07 am](https://cloud.githubusercontent.com/assets/49038/17141784/8401ecf0-531b-11e6-92fd-6b63bfd0ed7c.png)

Shipped, no-one else should have that problem 👯

orta:
Thanks @epologee

PR #376, opened 2016-07-26
Correct formatting of the fdescribe/fit grep calls #trivial

epologee:
I cheered a bit too early in #374, the 2.0.1 release solved the crashers, but still didn't produce an error-free `Dangerfile` when running `danger init`.

```shell
sh: -c: line 0: unexpected EOF while looking for matching `"'
sh: -c: line 1: syntax error: unexpected end of file
```

This PR removes one space and one dquote to get rid of the shell warnings. The output then becomes something more along the lines of:

```shell
Results:

Errors:
- [ ] fit left in tests
```


orta:
Thanks 👍

PR #377, opened 2016-07-27
[WIP] Fix #337. Buildkite should validate but not continue when not in a Pull Request

### LINKED ISSUE #337: Support Question: Should danger only be run when the build is for a pull request?
Hey all, 

Danger is running great for us except for that now when our build server (Buildkite) runs something that is not a pull request the danger step fails.

`Could not find a valid pull request within the known CI sources`

Is this the expected behaviour? Are we only meant to run danger if we are building for a pull request?

Thanks for the help, 
Ben

bentrengrove:
This is my attempt at fixing the issue identified in #337. Buildkite currently doesn't validate at all if its not in a PR. I moved some logic into initialise after comments from @orta to fix this.

Just wanted to check if this is correct before removing my [WIP].

Cheers, Ben

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Master
 
+* Fixes Buildkite not validating when not in a PR - bentrengrove
+
 ## 2.0.1
 
 * Updates the `danger init` template to 2.0 syntax - orta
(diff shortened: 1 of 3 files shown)


orta:
Yep looks good - wonder what's up with the fail

orta:
I had a think about it, and as I needed to make some changes to add docs around CI sources, I've also added what we talked about in #337 which should fix your problem and make it not really happen again.

orta:
So thanks for the PR! But it's negated by #380 👍

PR #380, opened 2016-07-28
Support validating CI and PRs separately, and sanely. 

orta:
And adds a command for outputting CI docs.

screen shot 2016-07-28 at 2 39 56 pm

This comes from two pain points. 

1. It was unintuitive about the connection between detecting the CI, and detecting the fact that it was on a PR. A detected CI should not fail the build if a PR isn't happening. This negates #377 👍  and the root cause that created the need for the fix.

2. I want the guide for getting started to cover all the CI sources, but I want that documentation in the code. This adds a linter that means anytime anyone else adds a new CI then they have to add docs. It means that we can be sure we're not breaking danger.systems.


orta:
Alright 👍

orta:
Windows builds.

PR #388, opened 2016-07-30
\n now works in HTML tables

### LINKED ISSUE #321: \n does not work in HTML tables
We should probably replace '\n' in warn/fail with `<br />` or similar?

screen shot 2016-07-09 at 14 38 11

https://github.com/CocoaPods/CocoaPods/pull/5607#issuecomment-231547745

marcelofabri:
This should fix #321


orta:
Great work @marcelofabri

PR #390, opened 2016-07-30
Adds an option for only showing the filename for a path with `html_link`

orta:
Also fixes a bug - it wouldn't work if you gave a path which began with a `/`


orta:
Windows builds.

PR #391, opened 2016-07-30
Shows attributes correctly in `danger plugins readme`.

orta:
There was an extra `<tr>`left over
Also there is now a check for the methods before showing the headline. Can't _easily_ see a use for attributes only, but sure.


orta:
👍 winders builds.

PR #395, opened 2016-07-31
Fix: removed extra have.

dblock:
I don't think this is very English, but I could have given be wrong.


orta:
awesome

orta:
Moya contributor guidelines say you can get in on the org, no pressure, want in?

dblock:
No thanks, I have too many responsibilities :)

PR #396, opened 2016-07-31
Enforce single quotes for non-interpolated strings.

dblock:
This discussion can only be a step up from tabs vs. spaces, so feel free to close without merging :) This PR saves 3 lines of code in `.rubocop.yml`. Removing special handling in Rubocop rules has been a general goal that I've been trying to pursue. If this goes through I'll probably remove everything else in another PR and use `rubocop --auto-gen-config` to generate exceptions.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,10 +6,6 @@ AllCops:
     - 'spec/fixtures/**/*'
     - 'lib/danger/plugin_support/plugin_parser.rb'
 
-Style/StringLiterals:
-  EnforcedStyle: double_quotes
-  Enabled: true
-
 # kind_of? is a good way to check a type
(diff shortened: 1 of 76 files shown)


orta:
We opted for enforcing double quotes on purpose, https://github.com/danger/danger/pull/307#issuecomment-230520772

Effectively it adds extra confusion for new contributors for little gain. I'm interested in the other though, this one was just well discussed over time.

dblock:
Personally, I think it's much easier to stick to Rubocop defaults and call it a day. This means your workflow becomes:

1) Write code.
2) `rubocop -a`
3) `rubocop --auto-gen-config`
4) Profit.

No more arguing over tabs or spaces :)

Issue #394, opened 2016-07-31
JRuby support


dblock:
Looks like danger use redcarpet only for markdown parsing, this not possible with jruby. 

Workaround for Gemfiles:

```ruby
unless RUBY_PLATFORM == 'java'
  gem 'danger', '~> 2.0', require: false
end
```

orta:
Hrm, yeah, I'm not sure that we can live without redcarpet, it's the only markdown rendered that does github flavoured markdown right. Think we should document this 👍

Bugagazavr:
As i know kramdown can support GFM http://kramdown.gettalong.org/parser/gfm.html

LeFnord:
@orta … yeap I would also prefer `kramdown` over `redcarpet`

orta:
Shipped as 2.1.1 👍

PR #397, opened 2016-07-31
replaces redcarpet with kramdown, to avoid jruby foo

LeFnord:
relates to: https://github.com/danger/danger/issues/394

### DIFF
--- a/lib/danger/helpers/comments_helper.rb
+++ b/lib/danger/helpers/comments_helper.rb
@@ -1,10 +1,10 @@
-require "redcarpet"
+require "kramdown"
 
 module Danger
   module Helpers
     module CommentsHelper
-      def markdown_parser
-        @markdown_parser ||= Redcarpet::Markdown.new(Redcarpet::Render::HTML.new(hard_wrap: true), no_intra_emphasis: true)
+      def markdown_parser(text)
+        Kramdown::Document.new(text, input: "GFM")
       end
 
       def parse_tables_from_comment(comment)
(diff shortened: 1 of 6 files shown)


LeFnord:
… its so cool 😄

lib/danger/helpers/comments_helper.rb
```diff
@@ -1,10 +1,11 @@
-require "redcarpet"
+# require "redcarpet"
+require "kramdown"
```
orta: Cool, yeah, this looks good - can you remove this moment then we're good to go 👍

lib/danger/helpers/comments_helper.rb
```diff
@@ -1,10 +1,11 @@
...
 
 module Danger
   module Helpers
     module CommentsHelper
-      def markdown_parser
-        @markdown_parser ||= Redcarpet::Markdown.new(Redcarpet::Render::HTML.new(hard_wrap: true), no_intra_emphasis: true)
+      def markdown_parser(text)
+        Kramdown::Document.new(text, input: "GFM")
```
DanToml: We probably still want to cache this?

orta: Yeah - good call, can't hurt 👍

orta:
( and a rebase - sorry about that 👯 )

lib/danger/helpers/comments_helper.rb
LeFnord: normally yeap, but Kramdown behaves a bit different to Redcarpet, it needs a string for initialising, so by caching the object it has always the same string to render, think this is not what would be expected

orta:
oh damn, err, sorry, I should have merged this before I merged #399 :/
can I get another rebase

orta:
Thanks @LeFnord  - we have [policy of offering](https://github.com/Moya/contributors) org access after a PR, would you like in? No pressure.

LeFnord:
thanks for the offer @orta, but I think I can also contribute without to be a member 😉 
and yeah maybe think about these policy and differ between small changes – such as this one, and more significant one … similar to the grape membership

orta:
Thanks for the recommendation, but we've thought a long time about the policies for running larger OSS projects and are pretty happy with what we've come up with 👍

Issue #384, opened 2016-07-29
Proposal: Support `Dangerfile.js`


orta:
Conceptually the `Dangerfile` is in Ruby because everything else is in Ruby, it makes for something that is easy to `eval` and inject your own code, and we have lots of history to work with. It's a solid mature language.

What if we also supported the most popular programming language.

In theory we would need to be able to expose any `Danger::Plugin` as a global javascript object in the scope. This can be done via something like [ruby_ffi-javascriptcore](https://github.com/ppibburr/ruby_ffi-javascriptcore) or [commonjs.rb](https://github.com/cowboyd/commonjs.rb) then we eval the `Dangerfile.js` and it has all the same powers as the `Dangerfile`.

Three main advantages: 
1. More people can read/write Dangerfiles
2. Could in theory also support node modules as danger plugins
3. With effective sandboxing, it may be possible to look deeper into danger as a service

orta:
an example of issue 1 - https://github.com/bnb/awesome-hyperterm/pull/68

orta:
Danger can probably run Python code to with this: http://rubypython.rubyforge.org

orta:
This kind of thing makes me wish there were subspecs for rubygems, rubypython support would add 3 dependencies

orta:
Looks like I can use this pattern: http://www.visuality.pl/posts/handling-optional-dependencies-in-gems

And then in the docs, we can say you need to have both `danger` and `rubypython` installed in order to have python Dangerfiles.

samdmarshall:
would we be able to write the dangerfile completely in JavaScript or Python or whatever language, or would it require we add the boilerplate code to get it to be handled by ruby?

orta:
For the JS Dangerfile, it would be completely in JS, I'd expose all of the ruby objects and their public methods to the JS runtime before the `Dangerfile.js` is eval'd and you should be able to write the exact same kind of Dangerfile but in JS.

For other environments it's harder to say outright without research - I looked at [rubypython](http://rubypython.rubyforge.org) but I'm not sure that I can expose Danger plugins to the python parser 

> There is no support for passing complicated (non-basic) Ruby types to Python.

samdmarshall:
[...]

PR #403, opened 2016-08-01
Separate the responsibilities of the Danger::Dangerfile

orta:
Currently the Dangerfile class handles: Parsing, Keeping track of all plugins, Printing all the plugin information.

This spits those responsibilities into different classes.

In order to make #384 feasible, this really should happen so that the responsibilities of a dangerfile are consistent regardless of the underlying Dangerfile implementation.

WIP cause tests need uncommenting


orta:
Given that this would cause merge conflicts with both #299 and #412 - I'm willing to pause on it for a day or two

orta:
Pushed to the `js` branch.

PR #407, opened 2016-08-01
[Project] Add .editorconfig

k0nserv:
#trivial

Improve consistency of styles and conventions
from all contributors. See http://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset=utf-8
+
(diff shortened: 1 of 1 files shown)


orta:
Works for me, I'll need to see if VS Code supports this

k0nserv:
Not sure about VS code specifically, but it works with a bunch of editors and there are extensions/plugins for most others. Should help with consistency

k0nserv:
Not sure about VS code specifically, but it works with a bunch of editors and there are extensions/plugins for most others. Should help with consistency

PR #411, opened 2016-08-04
[circle] improve doc

ReadmeCritic:
I believe the update is required to expose the github token for Circle CI.

### DIFF
--- a/lib/danger/ci_source/circle.rb
+++ b/lib/danger/ci_source/circle.rb
@@ -20,7 +20,7 @@ module Danger
   #
   # ### Token Setup
   #
-  # There is no difference here for OSS vs Closed, add your `DANGER_GITHUB_API_TOKEN` to the Environment variable settings page.
+  # There is no difference here for OSS vs Closed, add your `DANGER_GITHUB_API_TOKEN` to the Environment variable settings page. You must however turn on **Permissive building of fork pull requests**..
   #
   class CircleCI < CI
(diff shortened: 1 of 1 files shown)


ReadmeCritic:
Join? Sure why not

orta:
That's not true, permissive is only needed for forks, for same-repo branches it will work, I'm happy to have it changed to something like:

>  # There is no difference here for OSS vs Closed, add your `DANGER_GITHUB_API_TOKEN` to the Environment variable settings page. If you want to support running Danger on PRs from forks you  need to turn on **Permissive building of fork pull requests**.

What do you think?

ReadmeCritic:
@orta is it possible something has changed? or am I missing something

Before changing the setting:
https://circleci.com/gh/dkhamsing/scan-urls/73
> Suppressing export of environment variables DANGER_GITHUB_API_TOKEN fork PR builds

After changing the setting:
https://circleci.com/gh/dkhamsing/scan-urls/74
> Exporting env vars from project settings

orta:
> Suppressing export of environment variables DANGER_GITHUB_API_TOKEN fork PR builds

> ... fork PR builds

Your original text implied everyone had to do it, but you don't, you only have to do it for fork builds. That's what I changed it to 👍

ReadmeCritic:
Except I got this problem NOT on a fork. Anyway, closing. Keep this in mind if someone reports an issue. Thx

Issue #410, opened 2016-08-03
Improve Circle CI ENV Validation


DanToml:
Currently, there are many issues caused by bad/missing data from Circle CI ENV Vars. 
For example:
- https://circleci.com/gh/samdmarshall/pyconfig/238 (rebuild success: https://circleci.com/gh/samdmarshall/pyconfig/239)
- https://github.com/danger/danger/issues/408

To at least raise better errors, we should also validate the presence of:
- `CIRCLE_PROJECT_USERNAME`
- `CIRCLE_PROJECT_REPONAME`
- `CIRCLE_CI_API_TOKEN`

We could do this in `validates_as_ci?`, `validates_as_pr?` or an extra, whole environment validation check.
I think a whole environment step would allow us to give a more detailed error (I.e 'This is a problem with the provider', vs 'This is not a pull request/not running on CI')

\c @orta

samdmarshall:
This is still happening for me on non-PR builds and causes CI to fail consistently.

orta:
I've not shipped a release - as  I figured we had three big PRs coming in

samdmarshall:
Gotcha. Sorry, I should have clarified i was trying to give more information as this was previously happening on a PR branch and telling Circle to rebuild resolved it, but now it happens on primary my branch and PR branches and rebuilding doesn't resolve the issue has it previously had. I wanted to make sure those data-points were captured in this issue.

orta:
Just shipped 2.1.2 - wanna see if that fixes it?

samdmarshall:
After updating to 2.1.2, i stopped getting failures on my PR branches, but I am still getting failures on my primary branch: https://circleci.com/gh/samdmarshall/pyconfig/267

rpowelll:
I can confirm I'm also still seeing failures on non-PR branches. I can't link the build output because it's from a private repo, but the `bundle exec danger` step gives the following output:

```
bundler: failed to load command: danger (/usr/local/bin/danger)
Octokit::InvalidRepository: Invalid Repository. Use user/repo format.
  /usr/local/lib/ruby/gems/2.3.0/gems/octokit-4.3.0/lib/octokit/repository.rb:87:in `raise_invalid_repository!'
  /usr/local/lib/ruby/gems/2.3.0/gems/octokit-4.3.0/lib/octokit/repository.rb:31:in `initialize'
[...]

Issue #408, opened 2016-08-01
Danger is trying to run on master


rynbyjn:
This seems to have been introduced with v `2.1.x`. Our setup also includes having a `Dangerfile` for our organization, which is where this seems to be blowing up when trying to access `github.pr_json` or more specifically `github.pr_json[:base][:repo][:full_name]` in order to make commands through `octokit` that require a repo. It seems like the check that once printed out "Not a Pull Request - skipping `danger` run" might not be working as expected in the new versions. 

This can be temporarily fixed by locking down danger to `2.0.1`.

orta:
What type of CI server are you using?

rynbyjn:
Circle CI...

orta:
Should be better in the next release due to #413

PR #413, opened 2016-08-05
Add some more validations to the circle CI checking

orta:
https://github.com/danger/danger/issues/410 and https://github.com/danger/danger/issues/408

Lets it use just the environment variables if they exist, instead of always relying on the API too.

### DIFF
--- a/lib/danger/ci_source/circle.rb
+++ b/lib/danger/ci_source/circle.rb
@@ -28,6 +28,7 @@ module Danger
     end
 
     def self.validates_as_pr?(env)
+      # This will get used if it's available, instead of the API faffing.
       return true if env["CI_PULL_REQUEST"]
 
       # Real-world talk, it should be worrying if none of these are in the environment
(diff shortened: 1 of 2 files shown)


lib/danger/ci_source/circle.rb
```diff
@@ -30,6 +30,9 @@ def self.validates_as_ci?(env)
     def self.validates_as_pr?(env)
       return true if env["CI_PULL_REQUEST"]
```
DanToml: This can be present even when the subsequent validations are missing

lib/danger/ci_source/circle.rb
```diff
@@ -62,14 +65,17 @@ def pull_request_url(env)
     def initialize(env)
       self.repo_url = GitRepo.new.origins # CircleCI doesn't provide a repo url env variable :/
 
-      @circle_token = env["CIRCLE_CI_API_TOKEN"]
-      url = pull_request_url(env)
+      pr_url = env["CI_PULL_REQUEST"]
+      unless pr_url
```
orta: This could be `unless URI.parse(pr_urll)` and it would better reflect the idea

Issue #400, opened 2016-07-31
Refactor out code in spec/danger/requests_sources/github_spec.rb related to comments


k0nserv:
With the introduction of [CommentsHelper](https://github.com/danger/danger/blob/master/lib/danger/helpers/comments_helper.rb) there're now a set of tests in the [GitHub request source spec](https://github.com/danger/danger/blob/master/spec/lib/danger/request_sources/github_spec.rb) that can be move to [comments_helper_spec](https://github.com/danger/danger/blob/master/spec/lib/danger/helpers/comments_helper_spec.rb)

Namely

+ [generate_comment](https://github.com/danger/danger/blob/master/spec/lib/danger/request_sources/github_spec.rb#L173-L315)
+ [comment parsing](https://github.com/danger/danger/blob/master/spec/lib/danger/request_sources/github_spec.rb#L414-L453)

PR #412, opened 2016-08-05
set inline comments directly on the line when possible

champo:
Use GitHub's API to set inline comments. Comments can only be set on lines that are part of the unified diff by way of the [position in the diff](https://developer.github.com/v3/pulls/comments/#create-a-comment). There's a few tricky cases when lines go in and out of the diff. These cases mostly complicate the previous violations detection. There's probably a lot of room for improvement there.

PR #414, opened 2016-08-07
Hk move comments specs #trivial

hanneskaeufler:
This shovels around some specs to a more fitting place as described in #400.
It allowed me to remove the before block for the `#generate_comment` examples. I also split the example on deduplicating previous errors into two specs, one concerning the deduplication and one concerning the random compliment. Because of not wanting to stub the subject under test I asserted on a more general `\w` pattern.


orta:
Thanks for the first PR @hanneskaeufler - would you like to join the org? 
This fixe https://github.com/danger/danger/issues/400 and would likely require a rebase for #412

hanneskaeufler:
Thanks for the merge. Great first-commit experience btw. An easy to pick up issue, and I just love it when I can `git clone && bundle && rake` and be confident in the tests right from the get-go. Would be stoked to join the org ❤️

k0nserv:
Top notch 👍  Thanks for the PR

PR #417, opened 2016-08-08
fix rubocop issues excluded through .rubocop_todo #trivial

hanneskaeufler:
All of theme were double/single quote issues. Remove rubocop_todo.yml
because all the issues are fixed.


orta:
Rockin'

Issue #415, opened 2016-08-07
Proposal: Replace implicit dangerfile


orta:
The work being discussed in #384 would mean that the implicit Dangerfile support would need to look for every type of `Dangerfile` - which isn't optimal. I propose removing that, and maybe adding another root DSL attribute:

```ruby
run_dangerfile "repo/name"
```

Would implicitly run a Dangerfile:

```ruby
run_dangerfile "repo/name", file: "Dangerfile.js"
```

Would let you define the file, if it's not a ruby one.

This is a breaking change, and requires a version bump.

champo:
Would this need to be a breaking change? The original behaviour could be kept, only for ruby files. I'm inclined to the breaking change though, the limited behaviour would be unintuitive.

orta:
Yeah, I think so, I know there are people using this feature in production. Plus, I'd rather be on the side of more major versions than less, realistically. 

I don't want to try and split code-paths for JS/Ruby any more than is current, making it easy to add another language option in the something I'd like to do. I was debating having the JS danger files class as a major release anyway, so the two can be "released" at the same time, even if it comes out in a patch release before, as it doesn't break things

orta:
Alright, this is happening in #418

PR #418, opened 2016-08-10
Remove implicit org Dangerfiles

orta:
It's a bandage over a bug for https://github.com/danger/danger.systems/pull/81 - deeper discussion of this is in #415 

---

* **BREAKING** Removes the implicit Dangerfile support. - orta

  The implicit support was a feature that would check for an `[org]/danger`
  repo, and automatically parse that Dangerfile. Think it was a bit too magic,
  and the only person who I know is using it, said they think it should have 
  been this way from the start. I'm cool with this.

  To handle the replacement, we've added a new object to the DSL.
  This is the `danger` API. It has two responsibilities at the moment,
  downloading a Dangerfile:

  ```ruby
  danger.import_dangerfile = "artsy/danger"
  ```

  and importing local plugins:

  ```ruby
  danger.import_plugin [path_or_url]
  ```

  Which the astute may have remembered used to be the purview of `plugin.import`.
  Which is now removed, in favour of the `danger.import_plugin`. Think there's more
  space for improvement inside the context of `danger` than `plugin`.

  I also removed `plugin.download` - couldn't see a use for it inside a Dangerfile. Happy
  to change that, if it's being used it.


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Master
 
+* **BREAKING** Removes the implicit Dangerfile work. This is a WIP. The feature will be back in some form. - orta
```
KrauseFx: From this it's not 100% clear on what feature got removed

orta: yeah, it looks like this PR will have to provide all the docs and the replacement - so this will get a real changelog entry

orta:
I'm seeing:

```
NoMethodError: undefined method `danger_repo?' for #<Danger::RequestSources::GitHub:0x007fef42bd3b00>
  /Users/orta/.gem/ruby/2.1.3/bundler/gems/danger-c3c17f273728/lib/danger/commands/local.rb:120:in `check_and_run_org_dangerfile'
  /Users/orta/.gem/ruby/2.1.3/bundler/gems/danger-c3c17f273728/lib/danger/commands/local.rb:110:in `run'
  /Users/orta/.gem/ruby/2.1.3/gems/claide-1.0.0/lib/claide/command.rb:334:in `run'
  /Users/orta/.gem/ruby/2.1.3/bundler/gems/danger-c3c17f273728/bin/danger:5:in `<top (required)>'
  /Users/orta/.gem/ruby/2.1.3/bin/danger:23:in `load'
  /Users/orta/.gem/ruby/2.1.3/bin/danger:23:in `<top (required)>'
```

orta:
Danger is also dying on circle CI master - https://circleci.com/gh/danger/danger.systems/22?utm_campaign=build-failed&utm_medium=email&utm_source=notification

PR #422, opened 2016-08-12
Support JavaScript Dangerfiles

orta:
This adds rudimentary support for JavaScript Dangerfiles. 

- [x] Decide on the Rubygems situation
- [x] Executor should decide which Dangerfile type to run

Then I can start looking at using it in production on [Emission](https://github.com/artsy/emission/) to test out some of the more complex interactions.


orta:
Think I will include a separate gem, `danger-js` which only has two dependencies `danger` and `therubyracer`. I'll recommend that JS users have that as their dependency.  

This means less gems for anyone currently using it, including v8 which is a native module.

orta:
https://rubygems.org/gems/danger-js

Issue #424, opened 2016-08-14
Take into account the max size of a comment


orta:
trollolol.
Re [artsy/artsy.github.io/builds/152228051](https://travis-ci.org/artsy/artsy.github.io/builds/152228051).

```
POST https://api.github.com/repos/artsy/artsy.github.io/issues/252/comments: 422 - Validation Failed
Error summary:
  resource: IssueComment
  code: custom
  field: body
  message: body is too long (maximum is 65536 characters) // See: https://developer.github.com/v3/issues/comments/#create-a-comment (Octokit::UnprocessableEntity)
```

orta:
Done.

PR #425, opened 2016-08-14
Add a limit to the number of shown violations #trivial

hanneskaeufler:
Github imposes a limit on the length of the comment body, which might be
exceeded when having an enormous amount of violations. Truncate after a
pretty arbitrary amount of violations to keep the body length in check. This might or might not be a solution to #424.

As far as I see it, the comment length can become big for 
* large number of violations ✅ 
* large number of resolved violations
* large number of pure markdown messages
* long mardown messages
* very long violations/resolved violation messages

This PR would only address the large number of violations. I chose to add the limit in the view because this is likely different for different apis?


orta:
I can't rebuild your CI builds, as they're from a fork, but if you want to skip the CHANGELOG requirement, you need to add `#trivial` in the title 👍

k0nserv:
Should probably include a description that the limit was exceeded?

hanneskaeufler:
👍 yeah makes sense. I added something like that. Still I'm not entirely happy with this as uses some arbitrary magic number to truncate and disregards most of the causes as listed above 😐

hanneskaeufler:
Is there a way to see changes like these in a test environment? The comment likely looks very meh.

orta:
You could run with `bundle exec danger local` after changing the Dangerfile to show 100,000 characters?

orta:
> violations with this PR, too keep the size down the number shown has been truncated to 

-> 

> violations with this PR. Due to GitHub's max issue comment size, the number shown has been truncated to 

Shows that it's not just us deciding on a random number for no reason

orta:
Lets do it 👍

hanneskaeufler:
bildschirmfoto 2016-08-15 um 16 21 46

This is what it looks like when I slap the html into this pr in the inspector. I had to add a second `td`, missed that one ...

hanneskaeufler:
I moved the commit of what I missed into https://github.com/danger/danger/pull/427

Issue #426, opened 2016-08-15
Invalid Repository Error


bentrengrove:
Getting a invalid repository error (for a private repo so I can't share all the details unfortunately) when trying to run danger against a new project. Any pointers at working out what is causing this?

    /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/octokit-4.3.0/lib/octokit/repository.rb:87:in `raise_invalid_repository!': Invalid Repository. Use user/repo format. (Octokit::InvalidRepository)
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/octokit-4.3.0/lib/octokit/repository.rb:31:in `initialize'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/octokit-4.3.0/lib/octokit/repository.rb:55:in `new'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/octokit-4.3.0/lib/octokit/repository.rb:55:in `path'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/octokit-4.3.0/lib/octokit/client/pull_requests.rb:31:in `pull_request'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/danger-2.1.2/lib/danger/request_source/github.rb:55:in `fetch_details'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/danger-2.1.2/lib/danger/danger_core/environment_manager.rb:39:in `fill_environment_vars'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/danger-2.1.2/lib/danger/danger_core/executor.rb:31:in `run'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/danger-2.1.2/lib/danger/commands/runner.rb:56:in `run'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/claide-1.0.0/lib/claide/command.rb:334:in `run'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/gems/danger-2.1.2/bin/danger:5:in `<top (required)>'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/bin/danger:23:in `load'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/bin/danger:23:in `<main>'
	from /Users/StripySock/.rvm/gems/ruby-2.3.0/bin/ruby_executable_hooks:15:in `eval'
[...]

PR #428, opened 2016-08-15
Require BUILDKITE_PULL_REQUEST_REPO to not be the empty string

cysp:
Check that `BUILDKITE_PULL_REQUEST_REPO` is both present and not the empty string, as noted [here](https://github.com/danger/danger/issues/426#issuecomment-239886660).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Master
 
+* Improve detection of Buildkite's PR context - cysp
+
 ## 2.1.2
 
 * Improvements to `danger init` - orta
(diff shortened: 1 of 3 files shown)


orta:
Yep - this looks _perfect_ @cysp - just a README letting people know the change and it's done - great work

cysp:
Cools, take a squiz at that wording. Would you prefer the technical detail of the environment variable to be exposed in the changelog?

orta:
Nah, that's fine 👍

orta:
Want in on the org @cysp ?

cysp:
Sure. 😀 
Haha, it's hard to say no when you're invited to something named *Danger*.

PR #429, opened 2016-08-16
Hk tiny GitHub fixes #trivial

hanneskaeufler:
Just removed a tiny bit of duplication and fixed an insignificant typo 😇 I don't quite get the `issues` naming in there, aren't we talking about `comments` here?

### DIFF
--- a/lib/danger/request_source/github.rb
+++ b/lib/danger/request_source/github.rb
@@ -133,7 +133,7 @@ module Danger
             # We need to fail the actual build here
             is_private = pr_json[:base][:repo][:private]
             if is_private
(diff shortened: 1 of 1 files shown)


hanneskaeufler:
> Tests were not updated

Btw isn't that the point of a decent test-suite? Refactor some code and NOT having to touch the tests, they still pass?

```ruby
has_app_changes = !git.added_files.grep(/lib/).empty? # added instead of modified files 
has_test_changes = !git.modified_files.grep(/spec/).empty?

if has_app_changes && !has_test_changes
  warn "Tests were not updated"
end
```
Maybe?

orta:
> Btw isn't that the point of a decent test-suite? Refactor some code and NOT having to touch the tests, yet they still pass?

Yeah definitely! However, most PRs are not refactors, so it's aimed at the most often case

orta:
Looks good 👍

Issue #316, opened 2016-07-07
Erroneous error message when Danger fails to find branch-able point in git tree


samdmarshall:
When running CI today I encountered an error where Danger was reporting an error due to unable to find the branch point of the current working copy but ultimately reported the error as a problem with the DSL and that I should attempt to update the gem to see if that resolved anything. It seems to me that the error was that I was in a detached head state in the git repo and not an expected checkout state for the CI to correctly run. This caused some of the git commands that Danger uses to fail before reporting the DSL error. This is kinda tricky to get to reproduce since I think it was a bug in CircleCI running two different builds of the same branch concurrently. 

<details>
 <summary>Error Log</summary>
```
$ make ci
Checking for which... yes
Checking for cat... yes
Checking for cp... yes
Checking for touch... yes
Checking for find... yes
Checking for xargs... yes
Checking for rm... yes
Checking for python... yes
Checking for python3... yes
Checking for pip... yes
Checking for tox... yes
Checking for coverage... yes
Checking for gem... yes
Checking for danger... yes
============================
/home/ubuntu/virtualenvs/venv-2.7.10/bin/tox
GLOB sdist-make: /home/ubuntu/pyconfig/setup.py
py27 create: /home/ubuntu/pyconfig/.tox/py27
py27 installdeps: coverage
py27 inst: /home/ubuntu/pyconfig/.tox/dist/pyconfig-1.0.2.zip
py27 installed: coverage==4.1,pyconfig==1.0.2,pyparsing==2.1.5
py27 runtests: PYTHONHASHSEED='2212423871'
py27 runtests: commands[0] | coverage run --source=./pyconfig/ setup.py test
running test
running egg_info
writing entry points to pyconfig.egg-info/entry_points.txt
writing dependency_links to pyconfig.egg-info/dependency_links.txt
writing pyconfig.egg-info/PKG-INFO
writing top-level names to pyconfig.egg-info/top_level.txt
reading manifest file 'pyconfig.egg-info/SOURCES.txt'
writing manifest file 'pyconfig.egg-info/SOURCES.txt'
running build_ext
test_comments (tests.pyconfig_test.pyconfigTestCases) ... ok
test_conditionals (tests.pyconfig_test.pyconfigTestCases) ... ok
test_direct_assignment_automatic (tests.pyconfig_test.pyconfigTestCases) ... ok
test_direct_assignment_specific (tests.pyconfig_test.pyconfigTestCases) ... ok
[...]

Issue #160, opened 2016-03-29
Danger's commit range may include last PR's merge commit when there is only 1 commit in PR


orta:
See Danger's notes from #159 - I've seen a similar thing in `artsy/Eigen` - this is just a guess, but it fits.

daniel-beard:
I'm seeing this as well, the `modified_files` includes files from the PR's destination branch's last commit.

jamtur01:
Also seeing this.

orta:
I've not seen this, and I don't think it's fixed, if anyone sees it on an OSS project, let me know and I'll dig in with `danger local --pry --use-merged-pr=XX`

chrismcg:
I'm seeing something similar and I think it might be down to commits being made to the base branch while the PR is open. I haven't had time to delve too deeply into what's happening but am seeing warnings on a project that could be explained by this.

* staging is at `aaaaaa`
* make feature branch PR and add some commits `111111`
* someone else merges a PR so staging is now at `bbbbbb`
* add some commits to the feature branch PR so it's a `222222`

The feature branch PR modified_files is now between `222222` and `bbbbbb`, not `222222` and `aaaaaa` where it branched off from.

Will hopefully have time to check this in the next couple of days.

jamtur01:
Hey @chrismcg Did you puzzle this out at all? I'm stuck. :(

yimingtang:
I came across the same problem. I the diff option that danger use is not `...`. However, GitHub uses 
`git diff commitA...commitB` usually.

Ref: https://git-scm.com/docs/git-diff

yimingtang:
```
master:           ---a--------b--------c--------d-----e
                      \
                       \
feature-branch:         e------f----g
```

A developer sends a PR from feature-branch to master. GitHub uses `git diff e...g` command and shows the diff on web. Danger uses `ruby-git` gem which only supports `git diff e..g` command.

orta:
Interesting, I think this [image sums it up quite nicely](http://stackoverflow.com/questions/7251477/what-are-the-differences-between-double-dot-and-triple-dot-in-git-dif)

![git-diff-help](https://cloud.githubusercontent.com/assets/49038/17720357/4e296a38-63ef-11e6-89d0-3a8dcbbf8928.png)

jamtur01:
Oh nice catch.

jamtur01:
[...]

PR #432, opened 2016-08-17
Find the common merge commit, before looking for a diff in git

orta:
Re: #160 #316

This applies the fix in https://github.com/jamtur01/ruby-git/commit/6f0b644d24d9e061a852ba2f18a3c78c4ec1a0e4 / inline. 

Considering the amount of tests we have where we create a real repo and exec some commits, I'm happy this isn't breaking anything, what I'm not 100% sure of is whether it fixes the problem. I rarely would see it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Master
 
 * Improve detection of Buildkite's PR context - cysp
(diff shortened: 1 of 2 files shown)


k0nserv:
Looks good and yeah I agree that having a set of integration test that run against a proper git repository would be nice

orta:
I mean we already have that, that's how we test all of the diff-related parts of the git integration

PR #434, opened 2016-08-17
Add Danger on circle

orta:
In part because this seems to be fragile.


KrauseFx:
👍

marcelofabri:
Should we pass another `danger_id` to avoid one Danger overwriting the other one?

orta:
_assumably_ they should all give the same results @marcelofabri - so for now, no

orta:
If it acts odd, thats a bug

PR #449, opened 2016-08-21
Gitlab test is missing an expectation #trivial

hanneskaeufler:
This test does not have an expectation. Funny how that test even works, magic of lazily evaluated rspec `let`s.


orta:
👍

k0nserv:
Woops :) Nice catch

orta:
@hanneskaeufler afraid it needs a rebase ( you're welcome to use local branches so that we can trigger those too - if you'd like )

orta:
Thanks

PR #457, opened 2016-08-21
Spelling and grammar

connorshea:
Feel free to squash this down into just one commit.

### DIFF
--- a/lib/danger/ci_source/jenkins.rb
+++ b/lib/danger/ci_source/jenkins.rb
@@ -13,8 +13,7 @@ module Danger
   #
   # With that set up, you can edit your job to add `bundle exec danger` at the build action.
   #
-  # ### GitLab
-  #
+  # #### GitLab
   # You will want to be using the [GitLab Plugin](https://github.com/jenkinsci/gitlab-plugin)
   # in order to ensure that you have the build environment set up for MR integration.
   #
@@ -22,7 +21,11 @@ module Danger
   #
   # ### Token Setup
   #
+  # #### GitHub
   # As you own the machine, it's up to you to add the environment variable for the `DANGER_GITHUB_API_TOKEN`.
+  #
+  # #### GitLab
+  # As you own the machine, it's up to you to add the environment variable for the `DANGER_GITLAB_API_TOKEN`.
   #
   class Jenkins < CI
     def self.validates_as_ci?(env)
--- a/lib/danger/ci_source/jenkins.rb
+++ b/lib/danger/ci_source/jenkins.rb
@@ -5,7 +5,7 @@ module Danger
   # https://jenkins-ci.org
 
   # ### CI Setup
-  # Ah Jenkins, so many memories. So, if you're using Jenkins, you're hosting your own environment. You
+  # Ah Jenkins, so many memories. So, if you're using Jenkins, you're hosting your own environment.
   #
   # #### GitHub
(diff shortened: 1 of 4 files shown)


orta:
👯

lib/danger/ci_source/jenkins.rb
```diff
@@ -15,14 +15,14 @@ module Danger
...
-  # in order to ensure that you have the build environment set up for PR integration.
+  # You will want to be using the [GitLab Plugin](https://github.com/jenkinsci/gitlab-plugin)
+  # in order to ensure that you have the build environment set up for MR integration.
   #
   # With that set up, you can edit your job to add `bundle exec danger` at the build action.
   #
   # ### Token Setup
   #
-  # As you own the machine, it's up to you to add the enviroment variable for the `DANGER_GITHUB_API_TOKEN`.
+  # As you own the machine, it's up to you to add the environment variable for the `DANGER_GITHUB_API_TOKEN`.
```
connorshea: Also wondering if this should mention a `DANGER_GITLAB_API_TOKEN` if you're using GitLab?

k0nserv: Yeah that seems reasonable, should probably be under each instruction block

connorshea: Would you like that here or in a separate PR?

k0nserv: Here seems fine

lib/danger/ci_source/jenkins.rb
```diff
@@ -5,24 +5,27 @@ module Danger
...
+  # #### GitLab
+  # You will want to be using the [GitLab Plugin](https://github.com/jenkinsci/gitlab-plugin)
+  # in order to ensure that you have the build environment set up for MR integration.
   #
   # With that set up, you can edit your job to add `bundle exec danger` at the build action.
   #
   # ### Token Setup
   #
-  # As you own the machine, it's up to you to add the enviroment variable for the `DANGER_GITHUB_API_TOKEN`.
+  # #### GitHub
```
connorshea: This may not be what you meant, but I figured this was better than going to the 5th header level, I can change it if you disagree.

orta: Seems 👍 to me

spec/lib/danger/ci_sources/jenkins_spec.rb
```diff
@@ -86,7 +86,7 @@
         expect(described_class.validates_as_ci?(valid_env)).to be true
       end
 
-      it "doesnt validate when require env variables are not set" do
+      it "doesn't validate when require env variables are not set" do
```
k0nserv: I believe the convention is to no use contractions for rspec examples. At least that's what I've been trying to do

connorshea: Do we have a style guide or anything establishing this? I saw a few different misspellings of doesn't/does not so I wasn't sure which to use.

k0nserv: Other than what's enforced by rubocop, no. 

cc @orta

orta: No, nothing further than anything inside `rubocop.yml` - personally, I have no problem with anyone's choices, so long as they're readable and easily understood. 

I'd be hesitant to add anything that cannot be automated, given that I write most tests in the app - and I'm the least likely to conform :P

Issue #458, opened 2016-08-21
Some references to GitHub exclusivity still lingering


connorshea:
There are a few places where GitHub is implied to be the only supported platform so I wanted to note them in an issue to see if I'm misunderstanding and to get them fixed otherwise.

- [`xcode_server.rb`](https://github.com/danger/danger/blob/master/lib/danger/ci_source/xcode_server.rb) includes GitHub as the only supported request source.
- [`init.rb`](https://github.com/danger/danger/blob/master/lib/danger/commands/init.rb#L48) tells the user a GitHub account is necessary.

And on the flipside, [`github_helper.rb`](https://github.com/danger/danger/blob/master/spec/support/github_helper.rb) seems to include a bunch of GitLab references/terminology (e.g. merge requests vs. pull requests).

orta:
In the case of Xcode server, I'm afraid it only supports [Github & BitBucket](https://github.com/czechboy0/Buildasaur#hammer-requirements)

For `danger init`- yeah, that definitely needs a run through, it can probably figure out whether you're on GH/GL based on the remotes and offer better docs. 

WRT the github_helper I'd guess they overwrite each other ATM, and are redundant :)

k0nserv:
I might have copied and pasted a bit for the github_helper so that seems reasonable

connorshea:
@orta that'll work for gitlab.com and github.com, but not self-hosted GitLab or GitHub Enterprise. If it's not obvious what the platform is, I guess the user should be asked to choose from a selection of options.

orta:
Options: booooooo

But, I think that might make sense for the self-hosted/ enterprises. 👍

jk:
I just want to mention another place: I'm doing my first steps with Danger and GitLab+Jenkins right now. I'm just through the `bundle init+install` and `bundle exec danger init` process. When I test it via `bundle exec danger local` without setting any environment variables I get:

```
Danger local requires a repository hosted on GitHub.com or GitHub Enterprise.
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/danger-3.0.1/lib/danger/ci_source/local_git_repo.rb:52:in `initialize': No recent pull requests found for this repo, danger requires at least one PR for the local mode. (RuntimeError)
       	from /usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/danger-3.0.1/lib/danger/danger_core/environment_manager.rb:20:in `new'
[...]

PR #459, opened 2016-08-21
[Specs] Fix carless copying #trivial

k0nserv:
As pointed out in #458


spec/support/github_helper.rb
```diff
@@ -24,34 +24,6 @@ def stub_ci
       def stub_request_source
         Danger::RequestSources::GitHub.new(stub_ci, stub_env)
       end
-
-      def stub_merge_request(fixture, slug, merge_request_id)
```
k0nserv: We should try an reintroduce this to make the stubbing less verbose for GitHub though

orta:
👍

PR #475, opened 2016-08-26
Fix typo in gitlab.branch_for_merge documentation #trivial

jk:
The examples list a `gitlab.branch_for_base` which failed for us and it should be `gitlab.branch_for_merge`.


orta:
Thanks - @jk want in on the org?

jk:
Sure why not. I've a couple of gitlab issues coming :)

PR #481, opened 2016-08-27
Bitbucket Server

HeEAaD:
# Description
Adds Bitbucket Server aka Stash support:
* request source
* Jenkins integration
* dangerfile plugin

# Required Environment variables
See spec/support/bitbucket_server_helper.rb :
* `DANGER_BITBUCKETSERVER_HOST`, e.g. "stash.example.com"
* `DANGER_BITBUCKETSERVER_USERNAME`, e.g. "a.name",
* `DANGER_BITBUCKETSERVER_PASSWORD`, e.g. "a_password",
* `JENKINS_URL`, e.g. "http://jenkins.example.com/job/ios-check-pullrequest/",
* `GIT_URL`, e.g. "ssh://git@stash.example.com:7999/ios/fancyapp.git",
* `ghprbPullId`, e.g. "2080"

# Notes
* Bitbucket Server does not support :emojis:. We need to use UTF-8 emojis.
* Parsing and updating of old comments is not supported. Old comments will be deleted.
* I use the plain REST API due to lack of libraries.


justMaku:
Niiiiice! Thanks for this contribution 👍

Just one thing though: does Bitbucket finally allows to use emoji in the comments? I remember having issues with that back when I used Bitbucket Server (aprox. 4-5 months ago).

lib/danger/request_source/bitbucket_server.rb
```diff
@@ -0,0 +1,113 @@
...
+      def delete_old_comments(danger_id: 'danger')
+        uri = URI("#{pr_api_endpoint}/activities?limit=1000")
+        comments = get_json(uri)[:values]
+          .select { |v| v[:action] == "COMMENTED" }
+          .map { |v| v[:comment] }
+          .select { |c| c[:text] =~ /generated_by_#{danger_id}/ }
+          .each { |c| delete(URI("#{pr_api_endpoint}/comments/#{c[:id]}?version=#{c[:version]}")) }
+      end
+
+      def get_json(uri)
```
k0nserv: Might wanna move the logic around talking to the API to a BitBucketServerClient. Violates SRP a bit now when it's doing both the low level HTTP calls and the high level danger stuff

spec/lib/danger/plugins/dangerfile_bitbucket_server_plugin_spec.rb
```diff
@@ -0,0 +1,30 @@
+# coding: utf-8
+
+module Danger
```
k0nserv: I don't think we are using modules in specs instead we use

```ruby
describe Danger::DangerfileBitbucketServerPlugin
```

spec/support/bitbucket_server_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+      def stub_request_source
+        Danger::RequestSources::GitLab.new(stub_ci, stub_env)
+      end
+
+      def stub_pull_request
+        raw_file = File.new("spec/fixtures/bitbucket_server_api/pr_response.json")
+        url = "https://stash.example.com/rest/api/1.0/projects/ios/repos/fancyapp/pull-requests/2080"
+        WebMock.stub_request(:get, url).to_return(raw_file)
+      end
+
```
k0nserv: Redundant newline

spec/lib/danger/request_sources/bibucket_server_spec.rb
```diff
@@ -0,0 +1,34 @@
+# coding: utf-8
+require "danger/request_source/request_source"
+
+describe Danger::RequestSources::BitbucketServer, host: :bitbucket_server do
+  let(:env) { stub_env }
+  let(:bs) { Danger::RequestSources::BitbucketServer.new(stub_ci, env) }
+  
+  describe "the bitbucket server host" do
+    it "allows the set the host" do
```
k0nserv: I'd recommend using this style for the tests. See http://betterspecs.org/

```
describe "#host" do
  it "sets the host specified by `DANGER_BITBUCKETSERVER_HOST` do
    expect(bs.host).to eql("stash.example.com")
  end
end
```

lib/danger/danger_core/plugins/dangerfile_bitbucket_server_plugin.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module Danger
+  class DangerfileBitbucketServerPlugin < Plugin
+    def self.new(dangerfile)
+      return nil if dangerfile.env.request_source.class != Danger::RequestSources::BitbucketServer
+      super
+    end
+
+    def self.instance_name
+      "bitbucket_server"
```
orta: I don't know the terminology (as I've not used BitBucket as a use) but I'm willing to bet that this should probably be `bitbucket`, e.g. I would expect to write `bitbucket.pr_json` instead of `bitbucket_server.pr_json`

lib/danger/helpers/comments_helper.rb
```diff
@@ -102,6 +102,19 @@ def random_compliment
...
+        if emoji == "no_entry_sign"
+          "🚫"
+        elsif emoji == "warning"
+          "⚠️"
+        elsif emoji == "book"
+          "📖"
+        elsif emoji == "white_check_mark"
+          "✅"
+        end
+      end
```
orta: 👍

lib/danger/request_source/bitbucket_server.rb
```diff
@@ -0,0 +1,113 @@
...
+      include Danger::Helpers::CommentsHelper
+      attr_accessor :pr_json
+
+      def initialize(ci_source, environment)
+         self.ci_source = ci_source
+         self.environment = environment
+
+         @username = @environment["DANGER_BITBUCKETSERVER_USERNAME"]
+         @password = @environment["DANGER_BITBUCKETSERVER_PASSWORD"]
+      end
```
orta: Just in case, there are no access tokens in bb server?

orta:
> I use the plain REST API due to lack of libraries.

🎉 this is great, less dependencies

lib/danger/danger_core/plugins/dangerfile_bitbucket_server_plugin.rb
BenchR267: There is actually a difference between Bitbucket and Bitbucket Server. Bitbucket Server is just the new name for Stash. Bitbucket is the Github-like service. So it's like Github and Github Enterprise.

orta: Would this support both Bitbucket and Bitbucket Server? If so then it probably should be just `bitbucket` ( for example we don't have different Dangerfile syntax for both GitHub and GitHub Enterprise ( or gitlab.com / self hosted GitLab )

HeEAaD: Unfortunately not. Bitbucket and Stash (Bitbucket Server) are two separate products and they have different APIs.
https://developer.atlassian.com/static/rest/stash/3.11.6/stash-rest.html vs https://confluence.atlassian.com/bitbucket/use-the-bitbucket-cloud-rest-apis-222724129.html

lib/danger/request_source/bitbucket_server.rb
HeEAaD: https://developer.atlassian.com/static/rest/stash/3.11.6/stash-rest.html#authentication says:
> The prefered authentication methods are HTTP Basic (when using SSL) and OAuth.

orta:
@HeEAaD - let me know when you're happy, and I'll merge this into a branch on Danger, then give it some documentation to pass the "Core Docs Errors" above

HeEAaD:
@orta Wait a bit. I'll extract the network stuff out of bitbucket_server.rb (see https://github.com/danger/danger/pull/481#discussion_r76518367)

HeEAaD:
@orta extracted the network calls in aa60dbe. Now we have a much cleaner `BitbucketServer`.

I guess its ok to merge this PR now.

orta:
you sound so confident @HeEAaD  ;)

lib/danger/danger_core/plugins/dangerfile_bitbucket_server_plugin.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def pr_title
+      @bs.pr_json[:title].to_s
+    end
+
+    def pr_description
+      @bs.pr_json[:description].to_s
+    end
+
+    def pr_author
```
k0nserv: Can we get more data from bitbucket so that this plugin is closer to gitlab/github? labels, the diff, base commit, head commit, base branch, head branch for example?

lib/danger/request_source/bitbucket_server_api.rb
```diff
@@ -0,0 +1,70 @@
+# coding: utf-8
+require "danger/helpers/comments_helper"
+
+module Danger
+  module RequestSources
+    class BitbucketServerAPI
```
k0nserv: 👍

k0nserv: Strictly speaking this isn't a request source so having in in `lib/danger/request_source` might be confusing. Can we put it somewhere else?

spec/lib/danger/plugins/dangerfile_bitbucket_server_plugin_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  before do
+    stub_pull_request
+  end
+
+  describe "plugin" do
+    before do
+      dangerfile.env.request_source.fetch_details
+    end
+
+    describe "#pr_json" do
```
k0nserv: 💯

lib/danger/request_source/bitbucket_server_api.rb
HeEAaD: @k0nserv Any suggestions where to put this file?

orta: I would expect this in here, it's the main way in which Danger communicates with BitBucket server? It's the pattern we use for github/gitlab too

lib/danger/danger_core/plugins/dangerfile_bitbucket_server_plugin.rb
HeEAaD: I added pr_link, branch_for_base, branch_for_head, base_commit and head_commit in 20c79b8. I'm unsure about the `diff`. This will need another REST request and you can simply use `git. diff_for_file ` in the Dangerfile.

orta:
👍  I'm not sure how much time I'll have tomorrow, but I might be able to find time to get this merged into another branch and then documented up this evening, looking like a solid improvement. Thanks 👯

lib/danger/danger_core/plugins/dangerfile_bitbucket_server_plugin.rb
k0nserv: Great :+1:

lib/danger/request_source/bitbucket_server_api.rb
k0nserv: To me this is just a supporting class for the request source. This class doesn't inherit from the base request source so it doesn't seem right for it to live next to the other request sources. I would put it in `lib/danger/api_clients` or something like that

orta: Ah yeah, I wasn't looking at the full diff, @k0nserv is right, but that's a nitpick, I'm god with this getting merged into another branch so we can do a documentation run 👍

orta:
Great work @HeEAaD

PR #484, opened 2016-08-29
fixes crash when running `bundle exec danger`

### LINKED ISSUE #472: Undefined Local Variable or method 'ci_klass'
I'm having issues getting danger up and running on some build machines at work. I'll attach the output I'm seeing when I run `bundle exec danger`. Apologies if this is the wrong place.|

```Moby-Dick:danger-playground bilue$ bundle exec danger
bundler: failed to load command: danger (/Users/bilue/.gem/bin/danger)
NameError: undefined local variable or method `ci_klass' for #<Danger::Executor:0x007ffd8c342538>
  /Users/bilue/.gem/gems/danger-3.0.3/lib/danger/danger_core/executor.rb:16:in `run'
  /Users/bilue/.gem/gems/danger-3.0.3/lib/danger/commands/runner.rb:56:in `run'
  /Library/Ruby/Gems/2.0.0/gems/claide-1.0.0/lib/claide/command.rb:334:in `run'
  /Users/bilue/.gem/gems/danger-3.0.3/bin/danger:5:in `<top (required)>'
  /Users/bilue/.gem/bin/danger:22:in `load'
  /Users/bilue/.gem/bin/danger:22:in `<top (required)>'```

Has anyone seen something similar or have I missed something obvious with my ruby setup?

Gemfile looks like this:

```
source 'https://rubygems.org'

gem 'fastlane'
gem 'danger'
```

And if it helps I'm running danger through fastlane just by simply adding the stock danger lane.

hanneskaeufler:
When not in a CI environment, this would previously fail because
of an undefined variable. Fixes #472 

- [x] I'm kind of guessing this will not pass CI because there actually will be CI env vars set 🔥 

Testing on these classes is not a solved story yet I think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Show appropriate error message when GitHub repo was moved - KrauseFx
 * `danger plugins json [gem]` will now give gem metadata too - orta
+* Crash fix for `bundle exec danger` - hanneskaeufler
 
 ## 3.0.3
 
--- a/lib/danger/commands/runner.rb
+++ b/lib/danger/commands/runner.rb
@@ -53,7 +53,7 @@ module Danger
     end
 
     def run
-      Executor.new.run(base: @base,
+      Executor.new(ENV).run(base: @base,
                        head: @head,
                        dangerfile_path: @dangerfile_path,
                        danger_id: @danger_id)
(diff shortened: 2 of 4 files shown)


hanneskaeufler:
Well my guess was right ... second commit works around this.

orta:
Yep, looks great - I've never seen a test like that - nice work

Issue #476, opened 2016-08-26
Danger breaks on non-english locale systems (Jenkins)


jk:
### Problem
Danger breaks in https://github.com/danger/danger/blob/master/lib/danger/scm_source/git_repo.rb#L26 if the Jenkins job isn't running with an english locale set in `LANG` OR `LC_ALL` since it's expecting git responds in english and matches `Fetch URL`.

```bash
bundler: failed to load command: danger (/var/lib/jenkins/jobs/playground-danger-php-test/workspace/vendor/bundle/ruby/2.2.0/bin/danger)
NoMethodError: undefined method `split' for nil:NilClass
  /var/lib/jenkins/jobs/playground-danger-php-test/workspace/vendor/bundle/ruby/2.2.0/gems/danger-3.0.1/lib/danger/scm_source/git_repo.rb:26:in `origins'
  /var/lib/jenkins/jobs/playground-danger-php-test/workspace/vendor/bundle/ruby/2.2.0/gems/danger-3.0.1/lib/danger/request_source/request_source.rb:23:in `validates_as_ci?'
  /var/lib/jenkins/jobs/playground-danger-php-test/workspace/vendor/bundle/ruby/2.2.0/gems/danger-3.0.1/lib/danger/danger_core/environment_manager.rb:26:in `block in initialize'
  /var/lib/jenkins/.rbenv/versions/2.2.0/lib/ruby/2.2.0/set.rb:283:in `each_key'
  /var/lib/jenkins/.rbenv/versions/2.2.0/lib/ruby/2.2.0/set.rb:283:in `each'
  /var/lib/jenkins/jobs/playground-danger-php-test/workspace/vendor/bundle/ruby/2.2.0/gems/danger-3.0.1/lib/danger/danger_core/environment_manager.rb:22:in `initialize'
  /var/lib/jenkins/jobs/playground-danger-php-test/workspace/vendor/bundle/ruby/2.2.0/gems/danger-3.0.1/lib/danger/danger_core/executor.rb:26:in `new'
  /var/lib/jenkins/jobs/playground-danger-php-test/workspace/vendor/bundle/ruby/2.2.0/gems/danger-3.0.1/lib/danger/danger_core/executor.rb:26:in `run'
  /var/lib/jenkins/jobs/playground-danger-php-test/workspace/vendor/bundle/ruby/2.2.0/gems/danger-3.0.1/lib/danger/commands/runner.rb:56:in `run'
[...]

PR #485, opened 2016-08-29
[Git] Set env variable to en_US-UTF8

bogren:
Proposed change form #476. Need to test the change though, I'll give it a try tomorrow on a Travis machine.

### DIFF
--- a/lib/danger/scm_source/git_repo.rb
+++ b/lib/danger/scm_source/git_repo.rb
@@ -16,7 +16,7 @@ module Danger
 
     def exec(string)
       require "open3"
-      Open3.popen3(default_env, "git #{string}") do |_in, stdout, _err, _wait|
+      Open3.popen2(default_env, "git #{string}") do |_stdin, stdout, _wait_thr|
         stdout.read.rstrip
       end
     end
--- a/lib/danger/scm_source/git_repo.rb
+++ b/lib/danger/scm_source/git_repo.rb
@@ -15,7 +15,9 @@ module Danger
     end
 
     def exec(string)
-      `LANG=en_US.UTF-8 git #{string}`.strip
+      # See #476
+      ENV["LANG"] = "en_US.UTF-8"
+      `git #{string}`.strip
     end
 
     def head_commit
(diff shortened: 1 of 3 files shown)


orta:
@bogren this looks great - wanna add a CHANGELOG to credit yourself then we're good to go 👍

k0nserv:
@orta Is there a way to test backtick command line calls? Noticed that this doesn't have any tests

orta:
This is used in a lot of places, check out the windows CIs, but alas this change does not work on windows, which we need to keep supporting, might need to think of another way of doing it

bogren:
I'm looking for new possible solutions 🔎

orta:
If it comes to it, we can also use `system` which allows setting env vars, and return the output of that

lib/danger/scm_source/git_repo.rb
```diff
@@ -15,7 +15,9 @@ def diff_for_folder(folder, from: "master", to: "HEAD")
     end
 
     def exec(string)
-      `LANG=en_US.UTF-8 git #{string}`.strip
+      # See #476
+      ENV["LANG"] = "en_US.UTF-8"
+      `git #{string}`.strip
```
JuanitoFatas: How about?

```ruby
Open3.popen3({"LANG" => "en_US.UTF-8"}, "git #{string}") { |_stdin, stdout, _stderr, _wait_thr| stdout.read.strip }
```

[Open3](http://ruby-doc.org/stdlib-2.0.0/libdoc/open3/rdoc/Open3.html#method-c-popen3)

orta: is that synchronous?

orta: ah yeah, looks like it, aye, I'd be happy with that instead

JuanitoFatas:
Sorry, do we care about the error? Otherwise we can use popen2.

> ::popen2 is similer to ::popen3 except it doesn't make a pipe for the standard error stream.
> [Open3#popen2](http://ruby-doc.org/stdlib-2.0.0/libdoc/open3/rdoc/Open3.html#method-c-popen2)

orta:
I think we were already throwing it away, so no

JuanitoFatas:
Updated and I added a test to make sure 💪

orta:
Great -yeah - this looks 💯  - thanks @bogren and @JuanitoFatas 👍

PR #488, opened 2016-08-30
Adding support for Bitrise

kylejm:
:tada::warning: **This is my first ever Ruby PR** :warning::tada:

We use Bitrise for our CI, and we’d love to use Danger!

I’m a Ruby n00b. In fact this is my first go at writing Ruby code for something real, so please let me know if I missed anything or something seems wrong :smile:.

### DIFF
--- a/lib/danger/ci_source/bitrise.rb
+++ b/lib/danger/ci_source/bitrise.rb
@@ -1,0 +1,37 @@
+# http://devcenter.bitrise.io/docs/available-environment-variables
+
+module Danger
+  # ### CI Setup
+  #
+  # Add a script step to your workflow:
+  # ``` yml
+  # - script@1.1.2:
+  #     inputs:
+  #    - content: |-
+  #        bundle install
+  #        bundle exec danger
+  # ```
+  # ### Token Setup
+  #
+  # Add the `DANGER_GITHUB_API_TOKEN` to your workflow's Secret Env Vars
+  #
+  class Bitrise < CI
+    def self.validates_as_ci?(env)
+      env.key? "BITRISE_BUILD_URL"
+    end
+
+    def self.validates_as_pr?(env)
+      return true if env["BITRISE_PULL_REQUEST"] && !env["BITRISE_PULL_REQUEST"].empty?
+    end
+
+    def supported_request_sources
+      @supported_request_sources ||= [Danger::RequestSources::GitHub]
+    end
+
+    def initialize(env)
+      self.repo_slug = env["BITRISE_BUILD_SLUG"]
+      self.pull_request_id = env["BITRISE_PULL_REQUEST"]
+      self.repo_url = env["GIT_REPOSITORY_URL"]
+    end
+  end
+end
--- a/lib/danger/ci_source/bitrise.rb
+++ b/lib/danger/ci_source/bitrise.rb
@@ -17,21 +17,23 @@ module Danger
   #
   class Bitrise < CI
     def self.validates_as_ci?(env)
-      env.key? "BITRISE_BUILD_URL"
+      env.key? "BITRISE_IO"
     end
 
     def self.validates_as_pr?(env)
-      return true if env["BITRISE_PULL_REQUEST"] && !env["BITRISE_PULL_REQUEST"].empty?
+      return !env["BITRISE_PULL_REQUEST"].to_s.empty?
     end
 
     def supported_request_sources
-      @supported_request_sources ||= [Danger::RequestSources::GitHub]
+      @supported_request_sources ||= [Danger::RequestSources::GitHub, Danger::RequestSources::GitLab]
     end
 
     def initialize(env)
-      self.repo_slug = env["BITRISE_BUILD_SLUG"]
       self.pull_request_id = env["BITRISE_PULL_REQUEST"]
       self.repo_url = env["GIT_REPOSITORY_URL"]
+
+      repo_matches = self.repo_url.match(%r{([\/:])([^\/]+\/[^\/.]+)(?:.git)?$})
+      self.repo_slug = repo_matches[2] unless repo_matches.nil?
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/danger/ci_source/bitrise.rb
```diff
@@ -0,0 +1,37 @@
...
+    def self.validates_as_ci?(env)
+      env.key? "BITRISE_BUILD_URL"
+    end
+
+    def self.validates_as_pr?(env)
+      return true if env["BITRISE_PULL_REQUEST"] && !env["BITRISE_PULL_REQUEST"].empty?
+    end
+
+    def supported_request_sources
+      @supported_request_sources ||= [Danger::RequestSources::GitHub]
```
k0nserv: Are other request sources supported? Specifically GitLab? A quick google search seems to indicate it might be https://bitrise.uservoice.com/forums/235233-general/suggestions/9836094-built-in-gitlab-support?

k0nserv:
Please include tests in `specs/danger/ci_sources` Feel free to clone one of the existing CI Sources and modify it.

lib/danger/ci_source/bitrise.rb
```diff
@@ -0,0 +1,37 @@
...
+  #
+  # Add the `DANGER_GITHUB_API_TOKEN` to your workflow's Secret Env Vars
+  #
+  class Bitrise < CI
+    def self.validates_as_ci?(env)
+      env.key? "BITRISE_BUILD_URL"
+    end
+
+    def self.validates_as_pr?(env)
+      return true if env["BITRISE_PULL_REQUEST"] && !env["BITRISE_PULL_REQUEST"].empty?
```
viktorbenei: could be shorter:

```
return true if !env["BITRISE_PULL_REQUEST"].to_s.empty?
```

or just

```
return !env["BITRISE_PULL_REQUEST"].to_s.empty?
```

k0nserv: Nice catch, most other sources use this approach though. We can update it later

lib/danger/ci_source/bitrise.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def supported_request_sources
+      @supported_request_sources ||= [Danger::RequestSources::GitHub]
+    end
+
+    def initialize(env)
+      self.repo_slug = env["BITRISE_BUILD_SLUG"]
```
viktorbenei: hmm.. I don't think we have this as an env var right now..

Build Slug is the app's Bitrise.io build slug

viktorbenei: you could use something like this instead https://github.com/danger/danger/blob/master/lib/danger/ci_source/buildkite.rb#L41

lib/danger/ci_source/bitrise.rb
viktorbenei: not yet

spec/lib/danger/ci_sources/bitrise_spec.rb
```diff
@@ -0,0 +1,64 @@
+require "danger/ci_source/xcode_server"
+
+describe Danger::XcodeServer do
```
orta: If you change this, and the line above to use the bitrise, then the tests should start to work.

You're doing great so far 👍

orta: You can make your terminal auto run the tests as you make changes by using `bundle exec guard`

lib/danger/ci_source/bitrise.rb
```diff
@@ -0,0 +1,37 @@
...
+  #        bundle install
+  #        bundle exec danger
+  # ```
+  # ### Token Setup
+  #
+  # Add the `DANGER_GITHUB_API_TOKEN` to your workflow's Secret Env Vars
+  #
+  class Bitrise < CI
+    def self.validates_as_ci?(env)
+      env.key? "BITRISE_BUILD_URL"
```
viktorbenei: or you could use `BITRISE_IO` - set to `"true"` on bitrise.io VMs ; although that's true for our Docker images as well, if you'd use it locally https://github.com/bitrise-docker/bitrise-base/blob/master/Dockerfile#L15

spec/lib/danger/ci_sources/bitrise_spec.rb
k0nserv: Cool I didn't know about that either :)

spec/lib/danger/ci_sources/bitrise_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    it "does not validate when the required env variables are not set" do
+      expect(described_class.validates_as_pr?(invalid_env)).to be false
+    end
+  end
+
+  describe ".validates_as_ci?" do
+    it "validates when the required env variables are set" do
+      expect(described_class.validates_as_ci?(valid_env)).to be true
+    end
```
k0nserv: Typically nice to include a case where you set only the variables required for `validates_as_ci?` to be valid, but that would not be the case for `validates_as_pr?`. For example see this [example](https://github.com/danger/danger/blob/master/spec/lib/danger/ci_sources/travis_spec.rb#L25-L35) in the Travis CI tests

spec/lib/danger/ci_sources/bitrise_spec.rb
```diff
@@ -35,6 +40,11 @@
     it "does not validate when the required env variables are not set" do
       expect(described_class.validates_as_ci?(invalid_env)).to be false
     end
+
+    it "validates even when there is no PR" do
+      valid_env["BITRISE_PULL_REQUEST"] = nil
```
k0nserv: 💯

k0nserv:
LGTM, great work on your first Ruby PR 💯

orta:
Yep agreed with @k0nserv - great first Ruby PR

Would you like to join the org?

kylejm:
Thank you all so much for your patience and kind support! 

giphy

Sooo good to see those green ticks! 😁💃🎉

orta:
I'll try and get a minor Danger release out this evening, though my time is relatively short today.

Issue #487, opened 2016-08-30
bug in danger running imported Dangerfile


samdmarshall:
I came across this issue today while trying to run Danger on a new PR for one of my projects, while the Dangerfile in the project repo has been updated for the 3.0 syntax, I don't know if the org-level Dangerfile has. 

```
Running danger (PR)... 
/tmp/d20160829-19865-1c6uqzd/temporary_danger.rb:2:in `<top (required)>':  (Danger::DSLError)
[!] Invalid `Dangerfile` file: undefined local variable or method `git' for main:Object. Updating the Danger gem might fix the issue.

 #  from Dangerfile:30
 #  -------------------------------------------
 #  # Run the shared Dangerfile with these settings
 >  danger.import_dangerfile "samdmarshall/danger"
 #  -------------------------------------------
	from /opt/circleci/ruby/ruby-2.2.4/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /opt/circleci/ruby/ruby-2.2.4/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /opt/circleci/.rvm/gems/ruby-2.2.4/gems/danger-3.0.3/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb:116:in `block in import_local'
	from /opt/circleci/.rvm/gems/ruby-2.2.4/gems/danger-3.0.3/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb:114:in `each'
	from /opt/circleci/.rvm/gems/ruby-2.2.4/gems/danger-3.0.3/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb:114:in `import_local'
	from /opt/circleci/.rvm/gems/ruby-2.2.4/gems/danger-3.0.3/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb:103:in `import_url'
	from /opt/circleci/.rvm/gems/ruby-2.2.4/gems/danger-3.0.3/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb:63:in `import_dangerfile'
	from Dangerfile:30:in `block in parse'
	from /opt/circleci/.rvm/gems/ruby-2.2.4/gems/danger-3.0.3/lib/danger/danger_core/dangerfile.rb:194:in `eval'
[...]

PR #491, opened 2016-08-30
Ensure that danger.import_dangerfile evals inside the right context

orta:
Fixes - #487 

Ensures that external dangerfiles are ran and can use the plugins of the outer Dangerfile.


spec/lib/danger/danger_core/plugins/dangerfile_danger_plugin_spec.rb
```diff
@@ -52,4 +52,23 @@
       end
     end
   end
+
+  describe "#import_dangerfile" do
+    it "runs a fixtured Dangerfile" do
+      url = "https://raw.githubusercontent.com/example/example/master/Dangerfile"
+      stub_request(:get, url).to_return(status: 200, body: "")
+      @dm.danger.import_dangerfile("example/example")
+    end
```
hanneskaeufler: Is there an expectation missing in here?

orta: ah yeah, it used to crash, now it doesn't 👯

orta:
I agree with @hanneskaeufler that test was not doing anything, happy to merge on 🍏

Issue #184, opened 2016-04-24
Update error message


KrauseFx:
> Updating the Danger gem might fix the issue.

This message should only be shown if the user actually uses an outdated version of the gem.

hanneskaeufler:
How would we know if the installed danger is the current version?

```
is_outdated = `bundle outdated danger`.include?("* danger")
```

comes to mind but ist that worth doing every time an error is raised? Danger might not even be installed through bundler all the time but as a gem directly.

Maybe remove the "Updating the Danger gem ..." text and not bother with it?

marcelofabri:
We can use RubyGems API: https://rubygems.org/api/v1/versions/danger.json

hanneskaeufler:
Cool, did not know that. Seems like a lot of work + code for networking / networking error handling etc. for something pretty minor?

orta:
Aye, the concept comes from CocoaPods, where there is a guarantee of a local copy of the latest version as a string anytime an action is called. So it's easy there.

JuanitoFatas:
I'm working on this 😄

JuanitoFatas:
Implemented in #503.

Issue #382, opened 2016-07-28
Proposal: Major changes to `danger local`


orta:
lol @ this being a major SemVer change already. This is Danger 3.0, and command line changes like this is a major release.

These https://github.com/danger/danger/issues/381 https://github.com/danger/danger/issues/370 https://github.com/danger/danger/issues/368 issues all touch on the current cognitive dissonance around what you expect `danger local` to be and what it is.

After talking with @dbgrandi and @dblockdot - I think I've come to a conclusion.

* `danger local`- runs the current changes since the last commit in Danger against the current Dangerfile. 

* `danger pr [pr]` - runs the given PR against the current Dangerfile. This means you can PR against any PR, regardless of forks in an intuitive way.

---

The hardest thing here is `danger local` will need to be able to handle the ambiguity of _not_ having a PR to work against. Something @justMaku brought up here https://github.com/danger/danger/pull/299#issuecomment-231165513 - so good on you for being three weeks ahead of me.

I'm not 100% on the ideal here actually I see one of two patterns:

* Stub out with example data
* don't initialize a request source

This pattern will need to work with the work in #299 on supporting multiple request sources too. So maybe pulling in the commits @k0nserv mentioned here https://github.com/danger/danger/pull/299#issuecomment-235624279 is a pre-requisite to doing this work. Wherein we define a Random request source, ideally one that submits data based on Danger (the app)'s [music influences](http://danger.systems/guides/faq.html#danger-keeps-getting-refered-to-as-quot-her-she-quot-what-gives ).

marcelofabri:
👍 for improving `danger local`. Another thing that I'd add is making sure to support forks as well (in case of `danger pr`).

orta:
the [pr] bit would be `danger pr https://github.com/danger/danger/pull/362` so it can have any URL, which can then be used to support gitlab too

marcelofabri:
Awesome, I assumed that it was the PR number

ashfurrow:
It'd be cool to have an option to run danger against the diff from the current branch and `master`.

justMaku:
Since I was thinking about this recently I believe ideal solution to the problem with `danger local` would be to drastically change Danger's architecture. 

Here's a concept I have in mind:

* "Raw" danger executes a Dangerfile on a specified changeset: `danger master HEAD`. It returns the output using standard output.
* We have a new entity called `OutputSomething` (can't decide what to call it) which handles all of the PR updating staff we do in the `RequestSource` now.
[...]

PR #503, opened 2016-09-02
Feature/only print update message when local danger is outdated

JuanitoFatas:
Hi 👋 , This PR implements [Update error message #184](https://github.com/danger/danger/issues/184).

When an `Danger::DSL::Error` is raised:

* When local danger is outdated, prints the update danger message.
* When local danger is not outdated / latest version file not found, not print the update message.
* Outdated is checked with https://rubygems.org

<details>
<summary>staled information</summary>

Outdated? By looking at the latest danger version file (save at `~/.danger_version `).
The latest danger version file is added / updated when the danger library is loaded (`Danger.setup!`).
</details>

--

Some questions 🙋

* [x] When user's home directory is not writable 🤔 / not a directory (😱)
  ❗ will raise an `Danger::HomeDirectoyError` (**BREAKING BEHAVIOUR** 🚸).
  ~~- [ ] Should we fail?~~

  - [x] Or not fail, skip silently? => Should not fail for this tiny feature

~~* [ ] Should we only update the latest danger version file in a period of time?~~

~~* [ ] The latest danger version file location is okay?~~
  ~~- [ ] `~/.danger_version `~~

~~* [ ] The source file location is okay?~~
  - [x] `lib/danger/clients/rubygems_client.rb`

  ~~- [ ] `lib/danger/services/home_keeper.rb`~~

~~* [ ] [cross-platform home directory](https://ci.appveyor.com/project/orta/danger-47jiu/build/1.0.363) (dig rubygems/rubygems source code, maybe?)~~

Thanks! 🌈


orta:
> When user's home directory is not writable 🤔 / not a directory (😱)
> ❗ will raise an Danger::HomeDirectoryError (BREAKING BEHAVIOUR 🚸).

Seems fine, Danger's nearly always running on CI, [ish](https://github.com/danger/danger/issues/382) 

> Should we fail?
> Or not fail, skip silently?
> Should we only update the latest danger version file in a period of time?

We should just continue, this feature is definitely not important enough (in the grand scheme of why someone would use Danger ) to warrant breaking someone's build

>  The latest danger version file location is okay?

`~/.danger_version` seems fine to me - yes

> The source file location is okay?
> - lib/danger/clients/rubygems_client.rb
> - lib/danger/services/home_keeper.rb

Yep

>  cross-platform home directory (dig rubygems/rubygems source code, maybe?)

Yep

hanneskaeufler:
Looking epic, I wondered how one would implement this. Still seems like a looot of code for a very minor (?) change in error message that has to be maintained.

orta:
Hrm, I think a simpler approach might be better here.

I think we should probably skip the saving to the home and just always use the API? I doubt we'd be causing enough traffic to be a problem to rubygems. This greatly simplifies the windows support, and the complexity tradeoff for the feature

orta:
Considering I'm releasing multiple danger versions every week, there's _always_ a new version if you're using a `Gemfile`.

JuanitoFatas:
> I think we should probably skip the saving to the home and just always use the API? I doubt we'd be causing enough traffic to be a problem to rubygems. This greatly simplifies the windows support, and the complexity tradeoff for the feature

Sounds good to me.

hanneskaeufler:
Nice this looks like a great tradeoff!

JuanitoFatas:
Apart [from 3 offesens](https://travis-ci.org/danger/danger/jobs/157203747#L957):

screenshot 2016-09-03 05 00 36

This PR is good to go.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Fixes for `danger.import_dangerfile "org/repo"` - orta re:#487
 * Remove a variable assignment in comments_helper.rb - JuanitoFatas
+* Prints updated danger message when installed danger is outdated - JuanitoFatas
```
orta: can you move this up into the master section please ( https://github.com/dblock/danger-changelog/issues/9 )

orta:
Alright looks good, I'd be alright with removing the assignment in `if` rubocop if you have strong opinions ( I've wrote that kind of code then sighed and removed it )

The exceptions can be easily switched that that should be good to go IMO

orta:
Woooooooo

orta:
Awesome, let's do it - thanks for the contribution!

PR #504, opened 2016-09-02
Feature: import Dangerfile from path or gem.

dblock:
### Problem

I've extracted code from https://github.com/ruby-grape/danger/blob/master/Dangerfile into the https://github.com/dblock/danger-changelog gem. If I update the https://github.com/ruby-grape/danger project's Dangerfile to use that via `changelog.check` I will break pull requests in all projects that use Danger in that org. I must update half a dozen repositories at once. For each repository I have to modify `Gemfile` to include `danger-changelog`. If I don't do that, it just breaks because the org projects don't know anything about `danger-changelog`.

Pain and suffering. Chicken and egg.

### Solution

Allow one to package Dangerfile in a versioned gem and be able to load it from another Dangerfile like this:

```ruby
import_dangerfile gem: 'ruby-grape-danger'
```

### Side Effects in this PR

The `import_dangerfile(String)` method is deprecated with a warning. It now supports `path`, `github` or `gem`. In the future it can take a lot more options, making `import_dangerfile` much more extensible and more compatible with syntax used in, say, Gemfile.

### TODO

* Potentially add `import file: ...` and `import: url: `, so we can just load a file and a `url` without having to imply `Dangerfile`, this way the Dangerfiles can be named differently.
* Support the `gem` and `github` pair, where the .gemspec is loaded from Github, a familiar Gemfile syntax. Really in this case it's the same as `github: org/repo`. 

What else?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 ## master
 
+* Deprecated `import_dangerfile(String)` in favor of `import_dangerfile(Hash)` - dblock
+
+  The new `import_dangerfile` method supports loading Dangerfile from Github.
+
+  ```ruby
+  danger.import_dangerfile github: 'ruby-grape/danger'
+  ```
+
+  You can package a Dangerfile in a gem, add it to Gemfile and import it.
+
+  ```ruby
+  danger.import_dangerfile gem: 'ruby-grape-danger'
+  ```
+
+  Use a local path for testing Dangerfile changes.
+
+  ```ruby
+  danger.import_dangerfile path: '/Users/me/source/ruby-grape/danger'
+  ```
+
+  See [#504](https://github.com/danger/danger/pull/504) for details.
+
 ## 3.1.1
 
 * Fixes for `danger.import_dangerfile "org/repo"` - orta re:#487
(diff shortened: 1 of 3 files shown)


dblock:
It may be wise not to deprecate `import_dangerfile` because there's also `import_plugin`. I could make `import_dangerfile` support both `Hash` and `String`, I am not sure what's best? @orta

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 ## master
 
+* Deprecated `import_dangerfile` in favor of `import` - dblock
+
+  The new import supports loading Dangerfile from Github or another gem.
+
+  ```ruby
+  danger.import github: 'ruby-grape/danger'
```
orta: this is nice, as we can add the same for gitlab too

orta: well, strictly speaking, this supports gitlab too, so we can add that

orta:
I'm definitely in favour of this - this is awesome.

> may be better not to deprecate import_dangerfile and not add import, because there's also import_plugin. I could make import_dangerfile support both Hash and String

Yep, I'm in favour of killing the new `import` and moving that functionality back into `import_dangerfile` - deprecating the string support ( support the param, but log that it will be removed in a future version ) 

Mainly because I prefer the clarity in the API.  This makes me want the same thing for `import_plugin` now.

dblock:
This is ready to be merged AFAIK.

orta:
Yep - shipping

Giphy

dblock:
Here's an example of how this is used: https://github.com/ruby-grape/grape/pull/1480

PR #505, opened 2016-09-02
local_git_repo ci source must implement method #trivial

hanneskaeufler:
validates_as_pr? must be implemented by each ci source, even the local
git repo.

This fixes a possible crash when running with `DANGER_USE_LOCAL_GIT` env var. Considering as #trivial since this is not happening to anyone?


orta:
Works for me 👍

PR #507, opened 2016-09-03
Incremented version for next development iteration.

dblock:
Version 3.1.1 has already been released. I'd like to increment master to 3.2.0 as #504 is a major change with a backward compatible deprecation, so I can lock the minimum danger version to something in https://github.com/ruby-grape/danger/pull/5.

#trivial

### DIFF
--- a/lib/danger/version.rb
+++ b/lib/danger/version.rb
@@ -1,4 +1,4 @@
 module Danger
-  VERSION = "3.1.1".freeze
+  VERSION = "3.2.0".freeze
   DESCRIPTION = "Like Unit Tests, but for your Team Culture.".freeze
 end


dblock:
But please don't release it yet! I want to run the whole thing end-to-end first.

orta:
Hrm, this is not normally how I do gem versioning - I usually just switch it when it's time for a release but I'm alright for this. Will be likely making a big release over the weekend.

dblock:
This PR is a good example why you want to increment a version right after you release a gem. In this case I want `ruby-grape-danger` to depend on `danger` code that's on master, and I want to make sure nobody can accidentally uses `ruby-grape-danger` with an older version of danger wondering why it doesn't work.

orta:
Yeah, but I don't know what type of release I want the next one to be. Though I guess you're saying "just make it anything above x" which would still give me the space to say, "eh lets make this a minor instead of a patch"

dblock:
Right. I usually increment the patch number after release. And I change the major number with the first pull request that breaks things, for example.

And I just made this mistake forgetting to add a dependency on danger master and it rightfully complained with:

```
Bundler could not find compatible versions for gem "danger":
  In Gemfile:
    ruby-grape-danger was resolved to 0.1.0, which depends on
      danger (~> 3.2.0)

Could not find gem 'danger (~> 3.2.0)', which is required by gem 'ruby-grape-danger', in any of the sources.
```

PR #510, opened 2016-09-03
Fix an CHANGELOG item [ci skip]

JuanitoFatas:
Follow up of https://github.com/dblock/danger-changelog/issues/9#issue-174820591.


JuanitoFatas:
in favor of #511

PR #512, opened 2016-09-03
Add back the master section of CHANGELOG.md [ci skip]

JuanitoFatas:
Mistakenly removed, this PR adds it back 😊

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## master
+
 ## 3.2.0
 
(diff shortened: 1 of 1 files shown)


orta:
Nice one

orta:
Your commits don't seem to run in travis very often, wonder what's up with that?

orta:
Ah, you use CI skip - but we use travis as a "must be green" so the skip means that I have to use admin powers to merge

PR #517, opened 2016-09-03
Add a check for the merge inflection points existing

orta:
It is possible that the commit that we want too use to generate a github-like diff may _not_ be in the local CI's repo. Danger now takes a quick look, and will fetch things from the remote.


spec/lib/danger/ci_sources/git_spec.rb
```diff
@@ -1,167 +0,0 @@
-require "danger/scm_source/git_repo"
```
orta: this was in the wrong folder

lib/danger/scm_source/git_repo.rb
```diff
@@ -4,20 +4,25 @@
...
-    attr_accessor :diff, :log
+    attr_accessor :diff, :log, :folder
 
     def diff_for_folder(folder, from: "master", to: "HEAD")
-      repo = Git.open folder
+      self.folder = folder
+      repo = Git.open self.folder
 
       merge_base = repo.merge_base(from, to)
+      ensure_merge_base_exists!(merge_base)
```
cysp: When I saw this happen, it was actually the `from` argument to `merge_base` that did not exist locally, so I think that we'll already have raised an error by this point?

orta: interesting, in my case https://github.com/danger/danger.systems/pull/102 / https://circleci.com/gh/danger/danger.systems/89 the commit isn't in the list of commits at all

I don't think it's unreasonable to have the same checks for both `from` and `to`, then the `merge_base`

cysp: Isn't it this commit (https://github.com/danger/danger.systems/commit/3e956f142ee4f9b5da9fb26f0c2e041b199c6c7a)? It looks like it would have been the HEAD of master at the time that that CI build was running.
I suspect that the test for the presence of the commit that `git merge-base` tells us about can be omitted, as I'd be quite surprised if git was able to produce a merge base that it does not have details of.

orta: Looks like it 👍

orta:
Thanks @cysp

Issue #209, opened 2016-05-17
Allow to lint issues


mrackwitz:
It would be super cool if Danger could assist linting issues. Even though GitHub introduced the `ISSUE_TEMPLATE.md`, I see overall on projects I work on, that these are not fully understood or filled correctly.

Some of them could use some heuristics, indicators or measures to figure out whether an issue is meaningful or not. Feedback from Danger is likely easier to handle for all involved than from humans.

Also Danger could help classifying issues, which need revision by their author by tagging them as such to streamline the workflow for contributors and maintainers who triage issues.

orta:
To pull this off, we need to be able to support danger as a service - https://github.com/danger/danger/issues/42

As there would be no CI process to run

jeroenvisser101:
Tagging could happen within CI right? It'd only run when commits are pushed though, not when comments are placed (unless you manually change the CirceCI/Travis hook settings in GitHub, but then all would run when a comment is placed)

orta:
@mrackwitz means actual issues like this one, thing that don't or won't have related commits and thus no connection to CI

jeroenvisser101:
Oh, actual issues would be no-go until we introduce 'DaaS', yeah

orta:
Peril does this now 👍

PR #518, opened 2016-09-03
New Danger Local - Fixes #384

### LINKED ISSUE #384: Proposal: Support `Dangerfile.js`

orta:
Full context: #382 

This will have a long way to go - see all [the Local](https://github.com/danger/danger/issues?q=is%3Aissue+is%3Aopen+label%3ALocal) issues

### Decisions

- [x] Discuss with Danger what "scopes" should be called. - This didn't go too well, so far `scopes` is my fav
- [ ] Decide on whether `danger` should call all scopes, likely to be yes
- [ ] Decide on what the command for `danger local` looks like.
  - My guess is that if it sees scopes, and you have provided no input it lists out the examples
  - Perhaps we can take ideas from rake (how they do docs), and bundler [groups](http://bundler.io/v1.12/groups.html)
 - [ ] Decide on whether we can re-use executor?

### Implementation

- [ ] Data from a request source needs to be lazy
- [ ] needs tests, current local has 0 tests
- [ ] Improve error reporting?

- [ ] `danger local` 
   - [ ] should be entirely local, comparing changes to another local branch, defaults to master - 
   - [ ] supports scopes that allows you to move you api source work into a separate scope
   - [ ] scopes should be flexible enough that _if_ we get to a point where we can [do issues](https://github.com/danger/danger/issues/209) then it should feel natural to use that
   - [ ] think about what could come _in_ from a scope as an arg

- [ ] `danger local pr`
  - [ ] has the expectation of running against an exist PR, must use networking for data
  - [ ] needs to check if it has the required commits in the range, perhaps even setting up a remote?
  - [ ] should ideally support all 3 types of request sources, perhaps those classes can claim a URL?
  - [ ] May need to support including that it is a `gitlab` url for example, may not be possible to derive from url
  - [ ] alias `local mr` to `local pr`?

- [ ] Offer an easy way to create a `pre-commit` hook? `sh $(danger local pre-commit --with=linting,local)` ?


orta:
Killing this for now, when this comes up ( Danger 5.0 ) that can be a new PR

PR #524, opened 2016-09-04
Remove impossible Bitbucket example, fix some terms in GitLab docs #trivial

connorshea:
No API method, so that Bitbucket example doesn't actually work.


orta:
Awesome, looks good 👍

PR #526, opened 2016-09-04
Mask password when inspect BitbucketServerAPI

JuanitoFatas:
I saw [octokit.rb did this](https://github.com/octokit/octokit.rb/blob/efceb79b2d1db92d908abbe62d4fb0f436240002/lib/octokit/client.rb#L122), and think this is a good feature to add 🔒, just in case someone did the `inspect`🕵.


orta:
Yeah, definitely!

Wanna give this a rebase & merge yourself?

JuanitoFatas:
i merged it when 5/5 checks ok but now 4/5? Hmm...

PR #496, opened 2016-08-31
Bitbucket Server

orta:
I didn't do most of the work here - @HeEAaD did - @HeEAaD I've invited you to the org, you can use that access to push to this branch if you'd like too 👍  

This PR is mainly for writing docs, and wrapping up any minor issues at integration level - similar to what we did for GitLab in https://github.com/danger/danger/pull/446

PR #527, opened 2016-09-04
[Core] Provide a way to check the active scm host

k0nserv:
Adds `scm_provider` to the DSL which will let users and plugins check
what the active scm_provider is. Example

```ruby
  if scm_provider == :gitlab do
    message("Running under gitlab")
  end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Improved the width handling for the output of the `danger local` table - orta
 * Update comment & remove unused regexp name in request_source.rb - JuanitoFatas
 * Mask password on BitbucketServerAPI object - JuanitoFatas
+* Add `scm_provider` to the DSL allowing users and plugins ot check which scm provider is being used when running danger - K0nserv
 
 ## 3.2.0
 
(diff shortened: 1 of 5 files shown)


lib/danger/danger_core/dangerfile_dsl.rb
```diff
@@ -13,6 +13,19 @@ module DSL
...
+      # @return [Symbol] The name of the SCM Provider used for the active repository.
+      def scm_provider
+        if env.request_source && env.request_source.class == Danger::RequestSources::GitHub
+          :github
+        elsif env.request_source && env.request_source.class == Danger::RequestSources::GitLab
+          :gitlab
+        else
+          :unknown
+        end
+      end
```
orta: this this might be best on the `danger` plugin, it'll then get documented in the reference 👍

JuanitoFatas:
We also support [bitbucket server](https://github.com/danger/danger/blob/master/CHANGELOG.md#320)? :)

k0nserv:
Is the bitbucket support merged?

lib/danger/danger_core/dangerfile_dsl.rb
k0nserv: Will do :)

JuanitoFatas:
> Is the bitbucket support merged?

Yea merged in https://github.com/danger/danger/pull/496, shipped yesterday in 3.2.0. :)

k0nserv:
👌 will update

orta:
10/10 👌 
awesome

PR #536, opened 2016-09-09
Link to https://danger.systems rather than http://example.com

HeEAaD:
Link to https://danger.systems/ in Bitbucket Server comments.


orta:
Awesome thanks!

PR #537, opened 2016-09-09
Add `markdown_link` for the Bitbucket Server plugin

HeEAaD:
I check if Bitbucket server supports inline HTML in comments: Unfortunately not.
For that reason I added `markdown_link` for the Bitbucket Server plugin.


orta:
Thank you @HeEAaD 👍

PR #539, opened 2016-09-10
[Proposal] Replace URL of danger gem with danger.systems #trivial

KrauseFx:
@orta Feel free to merge if you think it makes sense 👍


orta:
I'm more hesitant on this, will close it to keep it being open for 90 days. Once we're sure about the redirection from http -> https on the gitlab pages - as we can't automate the deployment of new certs and I'm not going to force myself into doing a rote task that often - see  https://github.com/danger/danger.systems/issues/12#issuecomment-244643577 for more diets

Issue #493, opened 2016-08-30
Not setting DANGER_GITHUB_API_TOKEN doesn't lead to intended error


kylejm:
Not setting the `DANGER_GITHUB_API_TOKEN` lease to the following error:

`Could not find a Request Source for Danger::Bitrise`

from [environment_manager.rb](https://github.com/danger/danger/blob/8f50e957cbd065a411fa12ecfa82c3b2f06b82fe/lib/danger/danger_core/environment_manager.rb#L31), where as the following error:

`No API token given, please provide one using "DANGER_GITHUB_API_TOKEN"`

from [github.rb](https://github.com/danger/danger/blob/ed790d856587be5a00b1d42ffa7ecefe9acea89b/lib/danger/request_source/github.rb#L44) should have been printed.

Using version 3.1.0. Adding the token resolved this issue. Something to bare in mind for Bitrise is that Secret Env Vars are not exposed in PR builds, so you need to add your token to App Env Vars. Therefore the comments in [bitrise.rb](https://github.com/danger/danger/blob/master/lib/danger/ci_source/bitrise.rb#L16) need to be updated.

There's a [user voice ticket for Bitrise](https://bitrise.uservoice.com/forums/235233-general/suggestions/11701587-make-secret-env-variables-available-for-prs-from-t) on this matter. @viktorbenei should be able to tell us more about this ❤️.

orta:
@kylejm this might be worth adding to the documentation above the class [here](https://github.com/danger/danger/blob/master/lib/danger/ci_source/bitrise.rb#L4) - they can get [pretty long](https://github.com/danger/danger/blob/master/lib/danger/ci_source/travis.rb#L5) and that's just fine. They get turned into this guide:  http://danger.systems/guides/getting_started.html#setting-up-danger-to-run-on-your-ci

JuanitoFatas:
@kylejm Now danger ⚡ correctly inform you, my friend 😊 Please upgrade to v3.4.0, thanks!

[<img width="1033" alt="screen shot 2016-09-30 at 4 29 24 pm" src="https://cloud.githubusercontent.com/assets/1000669/18985551/1360dcd8-872b-11e6-9a07-a1524218a537.png">](https://travis-ci.org/JuanitoFatas/testdanger/builds/163950117#L224)

PR #541, opened 2016-09-11
Updating bitrise docs

kylejm:
Correction to using App Env Vars as apposed to Secret Env Vars, as mentioned in #493.


orta:
Nice, yeah, that would be your issue - thanks for the docs fix

orta:
Ace 👍

PR #542, opened 2016-09-13
Add per-file info (insertions, deletions, before/after contents)

tikitu:
This is very closely modeled on the `diff_for_file` support (read: code search and copy/paste).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## master
 
+* Add `git.info_for_file("some/file.txt")`: line counts and before/after file contents - tikitu
+
 ## 3.2.2
 
 * Link to https://danger.systems/ in Bitbucket Server comments - HeEAaD
--- a/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
@@ -120,7 +120,15 @@ module Danger
     # @return [Hash] with keys `:insertions`, `:deletions` giving line counts, or nil if the file has no changes or does not exist
     #
     def info_for_file(file)
-      modified_files.include?(file) ? @git.diff.stats[:files][file] : nil
+      return nil unless modified_files.include?(file)
+      stats = @git.diff.stats[:files][file]
+      diff = @git.diff[file]
+      {
+        :insertions => stats[:insertions],
+        :deletions => stats[:deletions],
+        :before => diff.blob(:src).contents,
+        :after => diff.blob(:dst).contents
+      }
     end
   end
 end
--- a/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
@@ -119,7 +119,7 @@ module Danger
     # Statistics for a specific file in this diff
     # @return [Hash] with keys `:insertions`, `:deletions` giving line counts, or nil if the file has no changes or does not exist
     #
-    def stats_for_file(file)
+    def info_for_file(file)
       modified_files.include?(file) ? @git.diff.stats[:files][file] : nil
     end
   end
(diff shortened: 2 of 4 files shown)


lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
```diff
@@ -114,5 +114,13 @@ def commits
     def diff_for_file(file)
       modified_files.include?(file) ? @git.diff[file] : nil
     end
+
+    # @!group Git Metadata
+    # Statistics for a specific file in this diff
+    # @return {:insertions => [Fixnum], :deletions => [Fixnum]} or nil
```
tikitu: I'm not a rubyist and am only guessing at the appropriate format for the `@return` annotation.

orta:
Interesting idea - I wonder if we should expand this even further - think you can add the before and after contents of the file into that dict? Currently I'm doing this: https://github.com/danger/danger.systems/blob/master/Dangerfile#L8-L11

spec/lib/danger/danger_core/plugins/dangerfile_git_plugin_spec.rb
```diff
@@ -99,6 +99,27 @@ module Danger
           end
         end
       end
+
+      describe "getting stats for a specific file" do
+        it "returns nil when specific stats do not exist" do
+          run_in_repo_with_diff do |git|
+            diff = git.diff("master")
+            allow(@repo).to receive(:diff).and_return(diff)
```
k0nserv: If the repo where this runs is a proper repo with diffs do we need to stub out responses?

lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
JuanitoFatas: The format is something like `@return [Hash] description` 😊

tikitu:
@orta Surely can, using the same diff information you do. I'll rename it to `info_for_file` in that case. I'm only wondering if it overlaps too much with `diff_for_file` (since that info is already available in the `Git::Diff::DiffFile` object) -- if you reckon it's worth exposing it more explicitly, though, I'm happy to go ahead with it.

lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
tikitu: Thanks, fixed

spec/lib/danger/danger_core/plugins/dangerfile_git_plugin_spec.rb
tikitu: Good call, that was cargo-cult code. Now that I'm adding file contents I probably will end up stubbing though. Thanks for the sharp eyes.

tikitu: On second thoughts: I think we do still need the stubbing, or at least, removing it is beyond the shallow understanding I have so far. It's giving us access to a pre-prepared diff (see `run_in_repo_with_diff` at the head of the file) which we otherwise wouldn't have because `@repo` doesn't really have a diff to report on.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 * Link to https://danger.systems/ in Bitbucket Server comments - HeEAaD
 * Add `markdown_link` for the Bitbucket Server plugin - HeEAaD
 * Updated Bitrise documentation about env vars - kylejm
+* Add `git.stats_for_file("some/file.txt")` to get line insertions/deletions counts - tikitu
```
hanneskaeufler: Missing the rename here too? Great PR!

orta:
@tikitu - I'm falling somewhat in favour of a more explicit API, a lot of my _current_ (and I'm flip-flopping a lot here) long-term thinking is around having to move to something where we can't pass through "rich" objects like a `Git::Diff` - so having something like this makes me feel better

We should get danger-rubocop on here, so that we know when that's what's failing.

orta:
Other than the rename at the top and a `bundle exec rubocop -a` this should be good to go 👍

CHANGELOG.md
tikitu: Oops... thanks, good catch.

orta:
👍

orta:
Thanks @tikitu - would you like to join our org?

tikitu:
@orta Sure, hook me up. (I could have replied to Danger, I suppose, sorry.)

PR #544, opened 2016-09-14
Added danger gem command.

dblock:
Added a `danger gem` command that creates a new `dangerfile-` gem using https://github.com/danger/dangerfile-gem-template. This closes 
https://github.com/danger/danger.systems/issues/106.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,15 @@
 ## master
+
+* Add `danger gem` - dblock
 * Link to https://danger.systems/ in Bitbucket Server comments - HeEAaD
 * Add `markdown_link` for the Bitbucket Server plugin - HeEAaD
 * Updated Bitrise documentation about env vars - kylejm
 
 ## 3.2.1
 
-* Fix for markdown accidentally crashing Danger when sending a message #520 - orta 
-* Improved the width handling for the output of the `danger local` table - orta 
-* Added a check for the merge inflection point inside the diff, and will fetch it if needed - orta 
+* Fix for markdown accidentally crashing Danger when sending a message #520 - orta
+* Improved the width handling for the output of the `danger local` table - orta
+* Added a check for the merge inflection point inside the diff, and will fetch it if needed - orta
 * Improved the width handling for the output of the `danger local` table - orta
 * Update comment & remove unused regexp name in request_source.rb - JuanitoFatas
 * Mask password on BitbucketServerAPI object - JuanitoFatas
--- a/lib/danger/commands/plugins/plugin_gem.rb
+++ b/lib/danger/commands/plugins/plugin_gem.rb
@@ -1,0 +1,92 @@
+require "claide/command/plugins_helper"
+require "claide/executable"
+
+module Danger
+  class PluginGem < CLAide::Command::Plugins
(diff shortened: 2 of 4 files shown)


orta:
Hrm, I'm not sure this is a good idea - this is basically a C&P of https://github.com/dbgrandi/claide-plugins/blob/claide-plugins/lib/claide/command/plugins/create.rb

We should do it properly by making the create command from claide work on both

dblock:
Claid is a CocoaPods plugin which shows info about available CocoaPods plugins or helps you get started developing a new plugin. It has been reused in Danger but not updated to be a generic thing. I don't think I am motivated enough to go down the rabbit hole of refactoring that thing to add a command to danger :)

dblock:
This test that runs the plugin fails on Windows. AFAIK `danger plugin` will fail on Windows too. What should I do about that?

orta:
Only the README hasn't been changed, inside it's 100% generic - we've just not taken those changes back to CocoaPods, might never end up doing that - hard to say. 

WRT windows, not really sure about that, if it doesn't have fork we'd have to move to `system` maybe?

orta:
Hrm, it is using `system` - that sucks

dblock:
I've conditioned the test not to run on Windows and opened https://github.com/dbgrandi/claide-plugins/issues/4.

dblock:
If someone refactors claide-plugins to be reusable, I'll gladly update this PR.

CHANGELOG.md
```diff
@@ -1,13 +1,15 @@
...
+* Add `danger gem` - dblock
 * Link to https://danger.systems/ in Bitbucket Server comments - HeEAaD
 * Add `markdown_link` for the Bitbucket Server plugin - HeEAaD
 * Updated Bitrise documentation about env vars - kylejm
 
 ## 3.2.1
 
-* Fix for markdown accidentally crashing Danger when sending a message #520 - orta 
-* Improved the width handling for the output of the `danger local` table - orta 
-* Added a check for the merge inflection point inside the diff, and will fetch it if needed - orta
```
k0nserv: These spaces should stick around right?

dblock: My editor removed them. I can do surgery to put them back, but why? :)

JuanitoFatas: Why [we exclude markdown files not trim trailing whitespaces](https://github.com/danger/danger/blob/master/.editorconfig#L16-L17)? :/

dblock: I love me a good tabs vs. spaces debate :) +1 @JuanitoFatas to make a PR to remove that.

orta: It's CocoaPods' fault - the CHANGELOG there is whitespace EOL dependent and so I try keep some constency

orta: Personally I don't care, it's meant to be human readable, happy to have them removed etc

orta:
I've made the changes in claide-plugins for this

orta:
I've shipped claide-plugins 0.9.2 with the API for this now - https://github.com/dbgrandi/claide-plugins

Will try and re-create this PR with that now in the next week

orta:
@dbgrandi ^

orta:
Awesome, I've taken this and made #608 with a bit more infrastructure 👍

PR #552, opened 2016-09-15
Remove unknown passings of nils

JuanitoFatas:
Reading danger/danger code before goes to 😴 , pondering what these `nil`s are. So let's set nil-defaults for them and remove some calls 🔪


orta:
rock

PR #555, opened 2016-09-16
Adds support for bitbucket cloud

WolfgangD:
- Duplicates bitbucket_server RequestSource and Plugin and adapts them for bitbucket cloud REST API.
- Dangerfile support is the same as for bitbucket_server but uses bitbucket_cloud name.
- Basic Auth via username/password is used and needs to be set via DANGER_BITBUCKETCLOUD_USERNAME and DANGER_BITBUCKETCLOUD_PASSWORD environment vars

### DIFF
--- a/lib/danger/request_source/bitbucket_cloud.rb
+++ b/lib/danger/request_source/bitbucket_cloud.rb
@@ -1,0 +1,78 @@
+# coding: utf-8
+require "danger/helpers/comments_helper"
+
+module Danger
+  module RequestSources
+    class BitbucketCloud < RequestSource
+      include Danger::Helpers::CommentsHelper
+      attr_accessor :pr_json
+
+      def initialize(ci_source, environment)
+        self.ci_source = ci_source
+        self.environment = environment
+
+        project, slug = ci_source.repo_slug.split("/")
+        @api = BitbucketCloudAPI.new(project, slug, ci_source.pull_request_id, environment)
+      end
+
+      def validates_as_ci?
+        # TODO: ???
+        true
+      end
+
+      def validates_as_api_source?
+        @api.credentials_given?
+      end
+
+      def scm
+        @scm ||= GitRepo.new
+      end
+
+      def host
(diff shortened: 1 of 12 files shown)


lib/danger/request_source/bitbucket_cloud.rb
```diff
@@ -0,0 +1,78 @@
...
+
+        comment = generate_description(warnings: warnings, errors: errors)
+        comment += "\n\n"
+        comment += generate_comment(warnings: warnings,
+                                     errors: errors,
+                                   messages: messages,
+                                  markdowns: markdowns,
+                        previous_violations: {},
+                                  danger_id: danger_id,
+                                   template: "bitbucket_server")
```
JuanitoFatas: `bitbucket_cloud` 😊 ?

WolfgangD: if I understood this correctly this is the template for the comment that get's posted to bitbucket. As the systems should be similar and the comment looks okay; I decided to stay with the same template.

orta: This makes sense

lib/danger/request_source/bitbucket_cloud.rb
```diff
@@ -0,0 +1,78 @@
...
+      def initialize(ci_source, environment)
+        self.ci_source = ci_source
+        self.environment = environment
+
+        project, slug = ci_source.repo_slug.split("/")
+        @api = BitbucketCloudAPI.new(project, slug, ci_source.pull_request_id, environment)
+      end
+
+      def validates_as_ci?
+        # TODO: ???
```
orta: I love how this comment of mine is propagating itself around the system 👯

JuanitoFatas:
LGTM, this is so cool 😎

orta:
Yep, looks all 👍

orta:
@WolfgangD would you like to join the org?

PR #557, opened 2016-09-16
Hk spec improvements #trivial

hanneskaeufler:
`with_git_repo` was kinda duplicates across different spec_helpers. pulled it up and made the origin more specific in the tests where it actually mattered.
Also removed all of the `switched to branch 'new'` logs in the spec output. That backwards compatibility test in `github_spec.rb` was passing without invoking anything, because the url would be set from the previous test already.


orta:
nicceee the `--quiet` is so needed 10/10

JuanitoFatas:
💯 :shipit:

PR #561, opened 2016-09-17
Add a spec for Executor#post_results #trivial

JuanitoFatas:
Add a spec for `Executor#post_results` 🍬

### DIFF
--- a/lib/danger/danger_core/executor.rb
+++ b/lib/danger/danger_core/executor.rb
@@ -52,12 +52,18 @@ module Danger
       end
     end
 
-    def post_results(dm, danger_id)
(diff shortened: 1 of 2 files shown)


lib/danger/danger_core/executor.rb
```diff
@@ -52,12 +52,18 @@ def run(env: nil,
       end
     end
 
-    def post_results(dm, danger_id)
-      gh = dm.env.request_source
-      violations = dm.violation_report
-      status = dm.status_report
```
orta: awww, poor Danger ( whose nickname is `dm` )

PR #563, opened 2016-09-17
Fix danger/ci_source/ci_source spec #trivial

JuanitoFatas:
The old spec does not even run (without `_spec`), this PR fixes the file name + add one spec for it.


orta:
Nicceee

PR #565, opened 2016-09-17
Update site references to https://danger.systems #trivial

JuanitoFatas:
Since now [our site](https://gitlab.com/danger-systems/danger.systems) is served by GitLab pages with https 🔒 .


orta:
I'm still a bit hesitant to do this ( _I_ would have to manually re-create the certs every 90 days, which is too short a timeframe for me to say I will do for as long as Danger exists ) so - once this current cert expires, lets see how the redirect is handled. If it works fine, and https links resolve to http - I'm totally cool with doing this 👍  - till then, no point keeping a PR open for ~70 days

JuanitoFatas:
Understood, thanks!

ge-fa:
Hi @orta ,

> I would have to manually re-create the certs every 90 days, which is too short a timeframe for me to say I will do for as long as Danger exists 

Why do you have to re-create the certs manually? Could you elaborate? 

Thanks!

orta:
https://letsencrypt.org only does 90 day certs, and there's no API on GitLab to update the keys - so I would need to download and run certbot again, and upload the new key to GitLab every 90 days

ge-fa:
Alright I see, thanks. I wasn't aware that the URL points to GitLab.

orta:
👍 no worries, it's an interesting constraint

PR #559, opened 2016-09-17
Improves "danger local" Pull Request finding

JuanitoFatas:
Before this PR,

`danger local` can only find PR like `Merge pull request #42` from **merge commits**.

After this PR,

`danger local` can find PR like `Merge pull request #42` and `Summary (#42)` from [50 recent **commits**](https://github.com/danger/danger/blob/ccfbd83f289e716766161dd9802f0ff979fc51f3/lib/danger/ci_source/local_git_repo.rb#L49).

--

##### Misc.

- Add pry-byebug for debugging.
- Target Rubocop for >= Ruby 2.0

--

<details>
<summary>Merge commits v.s. commits</summary>
The commit logs given to find merged Pull Request changed from:

`git log --merges --oneline` to `git log --oneline -50`
</details>

<details>
<summary>Merge pull request v.s. squash and merge</summary>
`Merge pull request #42` is great old git merge, which we are using at [danger](https://github.com/danger/danger/commits/master):

screenshot 2016-09-17 14 47 10

`Summary (#42)` is GitHub's [new *squash and merge* type of merge](https://help.github.com/articles/about-pull-request-merges/), which is employed by [SwiftWeekly/swiftweekly.github.io](https://github.com/SwiftWeekly/swiftweekly.github.io/commits/master):

screenshot 2016-09-17 14 47 15
</details>

PR #569, opened 2016-09-18
Increase number of commits danger local can search for merged PR

JuanitoFatas:
This size affects how many commits can `danger local` search for `--use-merged-pr`.

50 is too small (my fault in #559 😭). Let's increase to 1000000.

I did memory experiments against some repos to see if this size is too big:

```ruby
# measure.rb

require "objspace"
ObjectSpace.trace_object_allocations_start

pid = Process.pid; puts "Process: #{pid}"
before = Time.now
before_rss = `ps -eo pid,rss | grep #{pid} | awk '{print $2}'`.to_i

commits = `git log --oneline -n1000000`

after_rss = `ps -eo pid,rss | grep #{pid} | awk '{print $2}'`.to_i
puts "Elapsed time: #{Time.now - before} seconds"
difference = after_rss - before_rss

if difference >= 1024
  puts "Memory consumed: #{(after_rss-before_rss)/1024.0} MB"
else
  puts "Memory consumed: #{after_rss-before_rss} KB"
end

commits = nil
```

Use `ruby measure.rb` in the repo root.

- danger/danger (1281 commits as of 19 Sep 2016):

  ```
  Process: 5859
  Elapsed time: 0.109446 seconds
  Memory consumed: 180 KB
  ```

- rails/rails (59600 commits as of 19 Sep 2016)

  ```
  Process: 5853
  Elapsed time: 1.286322 seconds
  Memory consumed: 4.3515625 MB
  ```

- torvalds/linux (617821 commits as of 19 Sep 2016)

  ```
  Process: 5834
  Elapsed time: 8.914288 seconds
  Memory consumed: 36.8671875 MB
  ```

I think 1000000 is quite ok, because normally a project just few thousand commits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## master
 
+* Increase the number of commits `danger local` to search for merged PR - Juanito Fatas
+
 ## 3.3.1
 
 * "danger local" can find squash-and-merge-type Pull Request - Juanito Fatas
--- a/lib/danger/ci_source/local_git_repo.rb
+++ b/lib/danger/ci_source/local_git_repo.rb
@@ -47,7 +47,7 @@ module Danger
 
       pull_request_id, sha = MergedPullRequestFinder.new(
         env["LOCAL_GIT_PR_ID"] || "",
-        run_git("log --oneline -50".freeze)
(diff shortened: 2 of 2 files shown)


orta:
Yeah, that makes sense 👍

PR #577, opened 2016-09-19
Simplify initialization of Octokit client in request_sources/github.rb

JuanitoFatas:
This PR simplifies the init of Octokit client 🎋

### DIFF
--- a/lib/danger/request_sources/github.rb
+++ b/lib/danger/request_sources/github.rb
@@ -15,11 +15,7 @@ module Danger
         self.environment = environment
         self.support_tokenless_auth = false
 
-        Octokit.auto_paginate = true
         @token = @environment["DANGER_GITHUB_API_TOKEN"]
-        if api_url
-          Octokit.api_endpoint = api_url
-        end
       end
 
       def validates_as_api_source?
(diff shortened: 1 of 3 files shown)


lib/danger/request_sources/github.rb
```diff
@@ -34,16 +30,23 @@ def host
...
+      # backwards compatibility. `DANGER_GITHUB_API_BASE_URL` is the new
+      # correctly named variable.
       def api_url
-        # `DANGER_GITHUB_API_HOST` is the old name kept for legacy reasons and
-        # backwards compatibility. `DANGER_GITHUB_API_BASE_URL` is the new
-        # correctly named variable.
-        @environment["DANGER_GITHUB_API_HOST"] || @environment["DANGER_GITHUB_API_BASE_URL"]
+        @environment.fetch("DANGER_GITHUB_API_HOST") do
+          @environment.fetch("DANGER_GITHUB_API_BASE_URL") do
+            "https://api.github.com/".freeze
```
JuanitoFatas: This is Octokit's default value:

```
require "octokit"
=> true
2.3.1 (main)> Octokit.api_endpoint
=> "https://api.github.com/"
```

orta: interesting, so if the lookup fails then it uses the return value from the closure? Smart

PR #578, opened 2016-09-20
Run our specs randomly #trivial

JuanitoFatas:
Run specs randomly may catch some bugs in the future.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
-
-# This should move to the next release, see https://github.com/CocoaPods/Cork/pull/10
-gem "cork", git: "https://github.com/CocoaPods/Cork.git"
 gemspec
+
+# This should get removed when > 3.7.0 comes out
+gem "gitlab", git: "https://github.com/NARKOZ/gitlab.git", branch: "master"
 
 gem "danger-junit", "~> 0.5"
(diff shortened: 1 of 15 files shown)


JuanitoFatas:
A bug is being caught 😍, working on a fix 💪

orta:
Awesome

orta:
I simplified the failing test ( and merged in #582 so I had a fixed gemfile for `bundle exec guard` )

orta:
problem was that multiple tests used the same class for testing plugins, ut the plugin could realistically only be registered once because of that

orta:
Taking this, I want the Gemfile changes to do other work

Gemfile
```diff
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
-
-# This should move to the next release, see https://github.com/CocoaPods/Cork/pull/10
-gem "cork", git: "https://github.com/CocoaPods/Cork.git"
 gemspec
 
+# This should get removed when > 3.7.0 comes out
+gem "gitlab", git: "https://github.com/NARKOZ/gitlab.git", branch: "master"
```
connorshea: Is this intentional?

orta: Yes, the official gem uses JSON 1.7 ( because it's stored as json in rubygems), which conflicts with _something_ which makes my testing tooling not work - so I made https://github.com/NARKOZ/gitlab/pull/243 and used something which would use the ruby gemspec, and not hit the same problems )

PR #579, opened 2016-09-20
Improving the no token error during setup

orta:
///////////////////⚡///////////////////

Danger is sleeping, so I don't have too much time, but this is the start of improving the no token setup error messaging. Will finish tomorrow.

### DIFF
--- a/lib/danger/danger_core/environment_manager.rb
+++ b/lib/danger/danger_core/environment_manager.rb
@@ -70,44 +70,51 @@ module Danger
     end
 
     def raise_error_for_no_request_source(env, ui)
-      repo = ci_source.repo_url
-      source = nil, title = "", subtitle = ""
+      title, subtitle = extract_title_and_subtitle_from_source(ci_source.repo_url)
+      subtitle += travis_note if env["TRAVIS_SECURE_ENV_VARS"] == "true"
 
-      if repo =~ /github/
-        source = RequestSources::GitHub
-      elsif repo =~ /gitlab/
-        source = RequestSources::GitLab
-      elsif repo =~ /bitbucket.org/
-        source = RequestSources::BitbucketCloud
-      else
-        source = nil
+      ui_display_no_request_source_error_message(ui, env, title, subtitle)
+
+      exit(1)
+    end
+
+    private
+
+    def get_repo_source(repo_url)
+      if repo_url =~ /github/i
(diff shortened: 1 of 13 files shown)


k0nserv:
What about optional variables that you might want to set? Do we wanna raise any info about those too?

orta:
It wouldn't fail here if they were missing

k0nserv:
Nope, but it could be a good opportunity to surface them as well?

lib/danger/danger_core/environment_manager.rb
```diff
@@ -67,5 +67,41 @@ def self.danger_head_branch
...
+    def raise_error_for_no_request_source
+      title = ""
+      subtitle = ""
+      repo = ci_source.repo_url
+      case repo
+      when repo =~ /github/
+        title = "For your GitHub repo, you need to expose:" + RequestSources::GitHub.env_vars.join(", ").yellow
+      when repo =~ /gitlab/
+        title = "For your GitLab repo, you need to expose:" + RequestSources::GitLab.env_vars.join(", ").yellow
+
```
JuanitoFatas: ✂️

lib/danger/danger_core/environment_manager.rb
```diff
@@ -67,5 +67,41 @@ def self.danger_head_branch
...
+        " - #{klass}, #{klass.env_vars.join(', ').yellow}"
+        end
+        title = "For Danger to run on this project, you need to expose some of following the ENV vars\n#{available.join('\n')}"
+      end
+
+      if ci_source.class == Danger::Travis
+        subtitle = "If you have an open source project, you should ensure 'Display value in build log' enabled, so that PRs from forks work."
+      end
+
+      repo_url.
```
JuanitoFatas: This looks weird 😕❓

JuanitoFatas:
Do we also want to add `.env_vars` to the [base class](https://github.com/danger/danger/blob/master/lib/danger/request_sources/request_source.rb) and raise an not implemented error? 😊

orta:
allllright, this should be good for review now. Danger is out all day, so I get some other Danger time.

orta:
All feedback agreed with and fixed

JuanitoFatas:
🛳⛵🚢⛴🛥🚤

orta:
works for me

PR #583, opened 2016-09-25
Support for failing on errors

orta:
# ///////////////////⚡///////////////////

OK, so, Danger 🙆  is almost done up getting changed, don't have time to finish this - it actually is _done_ feature wise, but I'm starting to get my head wrapped around what it takes to actually make the executor testable - feel like it's getting there.

WIP till tests.

PR #589, opened 2016-09-28
The post_results method belongs to Dangerfile #trivial

JuanitoFatas:
# !!! This PR merges into #583

This method used `dangerfile` 3 times, it is jealous of `Dangerfile` ([feature envy](https://sourcemaking.com/refactoring/smells/feature-envy)), so this method should belong to `Dangerfile`.


orta:
Makes sense to me 👍

PR #592, opened 2016-09-28
Remove duplicate env check in TeamCity's GitHub env check #trivial

JuanitoFatas:
Probably copy-pasta 🍝, this PR removes this duplicate check.


orta:
Yep - looks like it

PR #596, opened 2016-09-29
Raises error if plugin file doesn't contain any plugin

JuanitoFatas:
This PR adds a feature that if plugin file doesn't contain any class inherits from `Plugin` then raises an error message: `"danger-orta-plugin.rb doesn't contain any valid danger plugin."`


orta:
Nice - this is a lovely feature 👍

PR #602, opened 2016-09-30
Not raises error if plugin that inherits from Danger::Plugin

JuanitoFatas:
Some users may not use `module Danger` on outer namespace. This PR fixes it.

If the regexp is not reliable, we can use Ruby [parser](https://github.com/whitequark/parser) gem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Added option to run a Dangerfile from a `branch` and/or `path` when using a remote repo - felipesabino
 * Add specs for `Danger::EnvironmentManager` - Juanito Fatas
+* Fixes plugin that inherits from `Danger::Plugin` should not raise error - Juanito Fatas
 
 ## 3.4.0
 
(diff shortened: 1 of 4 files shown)


orta:
I had a thought about 

```
/home/ubuntu/danger/vendor/bundle/ruby/2.1.0/gems/rspec_junit_formatter-0.2.3/lib/rspec_junit_formatter.rb:29:in `block in xml_dump_examples': undefined method `xml_dump_' for #<RSpecJUnitFormatter:0x0000000112f858> (NoMethodError)
```

it occurs when an _unnamed_ test is found - can you think of any tests we may have with no names?

JuanitoFatas:
> it occurs when an unnamed test is found - can you think of any tests we may have with no names?

RSpec gives this spec:

```ruby
  it "" do
    expect(1+1).to eq(2)
  end
```

a default name:

```
Danger::Plugin
  should forward unknown method calls to the dangerfile
  creates an instance name based on the class name
  should eq 2

Finished in 0.00228 seconds (files took 0.4083 seconds to load)
3 examples, 0 failures
```

orta:
> a default name:

of course they do, RSpec is so comprehensive ☀️

JuanitoFatas:
> of course they do, RSpec is so comprehensive ☀️

So the failure is actually intermittent. Restart circle build now passed 😓

PR #607, opened 2016-10-02
Multiple remotes support for Jenkins

### LINKED ISSUE #474: Jenkins GIT_URL environment variable is not always available
If one configures Jenkins GitLab-plugin according to their [documentation](https://github.com/jenkinsci/gitlab-plugin/blob/master/README.md#jenkins-job-configuration) you end up with two git repositories which were merged before anything is executed in that job. One consequence of the configuration is that you do have `GIT_URL_n` (in this case `GIT_URL_1` and `GIT_URL2`) environment variables instead of just `GIT_URL`.

Problem is https://github.com/danger/danger/blob/master/lib/danger/ci_source/jenkins.rb#L45 only checks for `GIT_URL`. This should handle those `GIT_URL_n` variables as well, but I don't know which `GIT_URL_n` var is to be chosen in case there are many of them.

JuanitoFatas:
Capture `GIT_URL_1` if user's Jenkins has multiple remotes.

Fixes #474.

https://wiki.jenkins-ci.org/pages/viewpage.action?pageId=69273545 (`⌘-F` for `GIT_URL_N` docs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 * Adds backwards compatibility for plugin testing API - sorry - @orta
+* Multiple remotes support for Jenkins - Juanito Fatas
 
 ## 3.4.1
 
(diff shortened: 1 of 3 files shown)


orta:
Great 👍

Issue #90, opened 2016-02-28
The default template should inspect the current folder to offer smarter defaults


orta:
We should have templates that adapt to the repo `init` was run in:

- [ ] A Gem
- [ ] An npm module
- [ ] Rails apps
- [ ] Sinatra projects
- [ ] Node apps
- [ ] React specific apps
- [ ] A Pod / SPM module
- [ ] iOS App

Note: This is a living document and I'll keep updating with ideas here.

orta:
Got some ideas from @jamtur01 today - thanks

marcelofabri:
I think we should be careful about where to draw the line. IMO some of these ideas are not about "Pull Request etiquette" but about linting the project. They could be implemented as plugins, though.

orta:
Agree, linting shouldn't but changing `*.[pod|gem]specs` probably should. Interesting, still worth some more thought.

orta:
in https://github.com/artsy/artsy.github.io/pull/217 - I wondered about a thing specific for blogging where Danger can check prose and links

orta:
Swift package manager: https://github.com/VeniceX/HTTPServer/pull/6/files#diff-16417461de376720abfe05ef4cb7d73aR4

orta:
A gem: https://github.com/realm/jazzy/pull/572/files#diff-16417461de376720abfe05ef4cb7d73aR1

PR #608, opened 2016-10-02
[WIP] Gem based generator for Dangerfiles

orta:
This adds three new commands:

The first is an abstract parent:

```sh
$ bundle exec danger dangerfile

Usage:

    $ danger dangerfile COMMAND

      Easily create you Dangerfiles.

Commands:

    + gem    Create a gem-based Dangerfile quickly.
    + init   Create an example Dangerfile.

```

* It adds `danger dangerfile init`, which shares the `Dangerfile` creation work with `danger init`.
* It also adds `danger dangerfile gem` which uses https://github.com/danger/dangerfile-gem-template to set up a new gem based project. Based on the code sharing in https://github.com/dbgrandi/claide-plugins/pull/5 👍  

This is blocked on some work on the template, as the last one I just tried didn't work by default, and I need to have a think about what testing a Dangerfile even is, and whether it's worth offering so much infrastructure for something you probably won't do.

I also need to add some tests.


lib/assets/PluginTemplate.rb.template
```diff
@@ -1,20 +0,0 @@
...
-        def self.description
-          [
-            "Describe what this plugin does",
-            "and how the user can use it"
-          ].join(" ")
-        end
-      end
-    end
-  end
-end
```
orta: wasn't being used, and wouldn't work anyway 👍

lib/danger/danger_core/dangerfile_generator.rb
```diff
@@ -0,0 +1,11 @@
+module Danger
+  class DangerfileGenerator
+    # returns the string for a Dangerfile based on a folder's contents'
+    def self.create_dangerfile(_path, _ui)
+      # Use this template for now, but this is a really ripe place to
+      # improve now!
+      dir = Danger.gem_path
+      File.read(File.join(dir, "lib", "assets", "DangerfileTemplate"))
+    end
```
orta: Been wanting to do https://github.com/danger/danger/issues/90 for half a year, it should be pretty easy for someone to do now

orta:
I gave this a run though - looks good 👍

Issue #609, opened 2016-10-03
PR commit race condition


fwal:
# Report
## What did you do?
Run `bundle exec danger --dangerfile=PreBuild.dangerfile --danger_id=prebuild` 
Run build scripts
Run `bundle exec danger --dangerfile=PostBuild.dangerfile --danger_id=postbuild`  

## What did you expect to happen?
Run Danger twice, one before builds scripts and one after, without problems.

## What happened instead?
The second danger crashed.
```
fatal: Not a valid branch point: '<hash>'.
fatal: ambiguous argument 'danger_head': unknown revision or path not in the working tree.
```

The `<hash>` in the error message points to a commit that was sent to the PR branch _after_ the build started and as such doesn't exist in the current build git repo when danger runs.

The following build (that's executed on the `<hash>` commit) will succeed without problem.

## Your Environment
* GitHub
* Jenkins 2.7.2
* Danger 3.3.2

orta:
Weird, Danger [should be fetching commits](https://github.com/danger/danger/blob/master/lib/danger/request_sources/github.rb#L64) when GitHub is saying use a commit which doesn't exist in the current git repo, could that be failing somehow?

fwal:
When trying to run the git commands locally I get:
`git rev-parse --quiet --verify <hash>` returns the hash ✅ 
`git branch test_head <hash>` gives me `fatal: Not a valid branch point:` 💀 
So that's probably what happens.

fwal:
`git fetch` solves it

orta:
o.0 - hrm, interesting, if you ( or anyone else who is) in  this position again, can you please try see if `-git -no-pager show <hash>` and see if that returns an empty value

We just switched from that to `rev-parse`

fwal:
Alright, will see if I can reproduce 👍

fwal:
`git --no-pager show <hash>` gives me `fatal: bad object` on commit that hasn't been fetched yet and the content of the commit if it has.

fwal:
Ah now I get it. It should actually be `git rev-parse --verify --quiet "<hash>^{commit}"` otherwise it just verifies that it's a valid sha1 😆

fwal:
So now I'm wondering; why does danger need the latest commit instead of just using the current commit?

fwal:
Made a PR to fix this but I'm still curious about the reasoning for doing this at all :)

orta:
> So now I'm wondering; why does danger need the latest commit instead of just using the current commit?

[...]

PR #610, opened 2016-10-03
Fix commitish exists check

fwal:
The current check just verifies that a commit hash is a valid hash rather than verifying that the commit exists in the repo. This may create problems as described in #609.


orta:
Nice one, there's some tests that verify the commands called - they should be trivial to update 👍

orta:
And would you like to join the org @fwal ?

fwal:
@orta sure!

fwal:
Thanks @orta 🎉

orta:
👍

orta:
lets get this shipped

PR #613, opened 2016-10-05
Fix ability to have false positives specs #trivial

JuanitoFatas:
✨ 💖 

Notice when running the specs:

```
WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives,
since literally any other error would cause the expectation to pass, including those raised
by Ruby (e.g. NoMethodError, NameError and ArgumentError), meaning the code you are
intending to test may not even get reached. Instead consider using `expect {}.not_to raise_error`
or `expect { }.to raise_error(DifferentSpecificErrorClass)`. This message can be suppressed by
setting: `RSpec::Expectations.configuration.on_potential_false_positives = :nothing`. 
```


orta:
Heh, Circle - works for me 👍

orta:
Fixed!

orta:
well, not fixed, but I rebuilt - same same

PR #617, opened 2016-10-06
remove stout output in spec and be more specific #trivial

hanneskaeufler:
There was a bunch of json being outputted while running specs. Rspec makes it easy to capture that.
The assertion is pretty unspecific on purpose because the actual json is verified in another spec.


orta:
Rockin'

PR #619, opened 2016-10-06
use `with_git_repo` helper where applicable #trivial

hanneskaeufler:
Felt like a lot of duplication even where the spec didn't specificly care about the exact commits/files in the repo.


orta:
nice one - thanks @hanneskaeufler  👍

hanneskaeufler:
my pleasure, thanks for reviewing

PR #620, opened 2016-10-07
Adds Chandler to run on every tag

orta:
I don't want more process on releases, so this let's travis and Chandler do everything for us 👍

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,3 +24,6 @@ script:
   - bundle exec rake spec
   - bundle exec danger --verbose
   - bundle exec danger init --mousey --impatient
+
+after_success:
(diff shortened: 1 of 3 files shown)


orta:
Chandler is what turns the CHANGELOG into GitHub Release Notes 👍

JuanitoFatas:
@orta Do you use `rake release` to release a new version? Then we can [hook chandler with rake release](https://github.com/mattbrictson/chandler#2-modify-the-rakefile) and we don't need travis to do for us.

orta:
I do, but then I have to expose my ( or another ) github access token in my ENV, this just automates that away for us all

orta:
Cool, update looks good to me

PR #621, opened 2016-10-07
Replace eql to eq in specs #trivial 

JuanitoFatas:
We probably want `eq` and it is 1 character less.


orta:
Haha - all the space savings :P

PR #623, opened 2016-10-07
Feature/pr argument

### LINKED ISSUE #622: Convert danger pr to use arguments instead of options for the PR address
# Report

## What did you do?

Ran `bundle exec danger pr https://github.com/artsy/eigen/pull/1899`

It would run my PR. This is a breaking change, but IMO, this too new to be worth a 4.0 for ;)

## What happened instead?

screen shot 2016-10-07 at 17 32 30

## Your Environment

Just local, on 3.5.1

--- 

I also need to look into the encoding issue too, I guess maybe there's emoji in the PR or something.

JuanitoFatas:
`danger pr` takes argument instead of passing via `--use-pr` option.

Fixes #622.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 ## master
 
+* `danger pr` takes an argument as PR URL instead of `--use-pr` option - Juanito Fatas
+
+  Work with Open PR:
+
+  ```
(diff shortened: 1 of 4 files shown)


JuanitoFatas:
@orta merge and release v3.5.2?

orta:
Yep! Let's do it

PR #625, opened 2016-10-08
Ensure that PR/MR diffs don't break the verbose table output

orta:
Also relates to #622 

-- 

Simplifies the table'd output when Danger shows the table, the diff can be really long and not too useful in that table - so I've removed it, and to add some extra paranoia, all metadata from plugins is not encoded into utf8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Add your own contribution below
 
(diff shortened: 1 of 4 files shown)


orta:
I've thrown in my fork of Junit rspec, it should not crash and give some error messaging https://github.com/sj26/rspec_junit_formatter/compare/master...orta:errors?expand=1

Issue #618, opened 2016-10-06
Proposal: Argument to make Danger post new comments instead of editing it's first one


bruno-rocha-movile:
I personally don't like how multiple builds of Danger makes it edit it's first comment instead of posting new ones, because it makes the comment chain rather confusing. It's specially worse to my team since we use 'Test it!' to trigger it, so usually the comments section consist of the same guy posting 'Test it!' a few times, with a single Danger comment on the top of the chain. Visually, it doesn't look nice.. It would be really cool if it had an option to post new comments instead of editing the first one.

orta:
That does sound like a useful option 👍

connorshea:
@orta --immutable-comment, --static-comment, --no-modify-comments?

orta:
`--replace-comment`?

bruno-rocha-movile:
Since this would be a result of multiple runs of danger, perhaps `--new-comment` ?

orta:
Hrm, if this will result in a new comment _every_ time, then we should re-use existing infrastructure - we already support a `    --danger_id=<id>                    The identifier of this Danger instance` option, so this should be `--unique-danger-id` which would create a timestamp based id, which means it will create a new comment each time

bruno-rocha-movile:
That worked @orta ! Awesome!

I do believe this functionality needs an argument with a more obvious name though, so I'll definitely take a shot at this :)

JuanitoFatas:
Implemented in #632.

PR #632, opened 2016-10-10
Added the '--new-comment ' argument

bruno-rocha-movile:
Implementing [#618](https://github.com/danger/danger/issues/618) - Currently, users that would like to have Danger post new comments on every build (instead of editing it's first one) need to generate new danger_ids on every build. This pull request wraps this process into a new argument.

I decided to use a formatted time string as the ID for visual purposes, since GitHub shows the instance's danger_id on the PR's status grid.

An obvious upgrade for this would be making the default danger_id ("danger") update according to the usage of this argument, so that the following process:
```
bundle exec danger
bundle exec danger --new-comment
bundle exec danger
```
..would make Danger update the comment created at the second call, instead of the first one. But since most CI servers wipe their workspaces before the builds, I could not find a decent way to implement this. This could also be used to remove the previous build's check results, if the user would like only one check to be displayed. If you guys have any suggestion for this, I would be happy to add this functionality to the PR.


lib/danger/commands/runner.rb
```diff
@@ -56,7 +57,8 @@ def self.options
         ["--head=[master|dev|stable]", "A branch/tag/commit to use as the head"],
         ["--fail-on-errors=<true|false>", "Should always fail the build process, defaults to false"],
         ["--dangerfile=<path/to/dangerfile>", "The location of your Dangerfile"],
-        ["--danger_id=<id>", "The identifier of this Danger instance"]
+        ["--danger_id=<id>", "The identifier of this Danger instance"],
+        ["--new-comment", "Makes the build process ignore other Danger instances, resulting in a new comment. Overrides '--danger_id'."]
```
JuanitoFatas: IMHO a little bit too implementation specific 😅? How about something like `Makes Danger post a new comment for each build`😄?

JuanitoFatas:
Would you show some screenshots for this danger new comment functionality? Thanks!

lib/danger/commands/runner.rb
bruno-rocha-movile: I didn't wanted to give the impression that removing the argument on a subsequent build would make Danger edit the new comment... I think someone could be confused by it, what do you think?

bruno-rocha-movile:
Here's what happens:
screen shot 2016-10-10 at 2 41 05 pm

It makes it easier to read your pull request, depending on your team's rules for pull requests. In my team's case, it's wonderful :)

orta:
Hah - that avatar

JuanitoFatas:
Would you like to join the org? @bruno-rocha-movile /cc @orta

lib/danger/commands/runner.rb
JuanitoFatas: 👍

orta:
👍 and @JuanitoFatas I've given you root org access to @danger now too - you've been doing an awesome job

bruno-rocha-movile:
@JuanitoFatas @orta of course! Thanks for the invite! I have some other interesting improvements that I was planning to do as well, Danger is amazing :)

orta:
Sent!

Issue #587, opened 2016-09-26
Warning on successful build for incorrect GitHub permissions


kylejm:
# Report

## What did you do?

Run Danger after a PR.

## What did you expect to happen?

A warning about lack of permission to mark the status of the PR. I've noticed that if Danger fails the PR, you _do_ get a warning stating that the token doesn't have `write` permissions. It would be cool if the same thing happened for the success case, particularly and/or less experienced users.

## What happened instead?

_All good_ was printed out, but the was no alert of lacking permissions to set the PR status on GitHub.

## Your Environment

* Using GitHub
* Using Bitrise for CI
* Danger version 3.2.2

PR #643, opened 2016-10-12
Hk status update improvements

hanneskaeufler:
So when danger failed a build, it would output a comment if it was unable to update the PR status. However, when the danger run was not containing any errors, it would fail silently to update the PR status. Reported in #587

### DIFF
--- a/spec/lib/danger/request_sources/github_spec.rb
+++ b/spec/lib/danger/request_sources/github_spec.rb
@@ -155,11 +155,11 @@ RSpec.describe Danger::RequestSources::GitHub, host: :github do
 
       it "aborts when access to setting the status was denied but there were errors" do
         stub_request(:post, "https://api.github.com/repos/artsy/eigen/statuses/pr_commit_ref")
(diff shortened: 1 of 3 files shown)


spec/lib/danger/request_sources/github_spec.rb
```diff
@@ -152,6 +152,34 @@
...
+        @g.pr_json = {
+          "head" => { "sha" => "pr_commit_ref" },
+          "base" => { "repo" => { "private" => true } }
+        }
+
+        expect do
+          @g.submit_pull_request_status!(warnings: violations(["error"]))
+        end.to output(/warning.*not have write access/im).to_stdout
+      end
     end
```
orta: nice, one line of app code ~30 lines of tests

hanneskaeufler: lol yeah, I even chickened out and did not add a test for the second branch of the `is_private` conditional. 😆

orta:
Nice one

Issue #641, opened 2016-10-11
--new-comment improvements


bruno-rocha-movile:
Creating this issue for documenting purposes, as I can't test my changes for now

Like I mentioned on [#632](https://github.com/danger/danger/pull/632) , one obvious improvement was to make --new-comment be able to know what was danger's last comment, so that removing the command would make Danger edit the correct comment instead of trying to edit the default "danger" instance (which could be any of the chain's posts, or none at all).

Another issue that showed up was that running --new-comment several times would produce several, immutable checks on the pull request. That was very ugly.

[Here](https://github.com/bruno-rocha-movile/danger/commit/9f84704cff6a798d7c2777703d3cb7f135cd4681), I changed the argument to stop changing the danger_id (so now we only get one PR check), and instead, the argument now affects the logic that decides whether or not Danger should update or create a new comment.

I also made Danger search for it's last comment instead of the first one, so that removing the argument on a subsequent run will make it edit the correct comment. For users that don't use this argument, this will hopefully make no difference.

I'll test this as soon as my CI server comes back online :)

bruno-rocha-movile:
Merged on #644

PR #644, opened 2016-10-13
--new-comment improvements

bruno-rocha-movile:
As mentioned on [#641](https://github.com/danger/danger/issues/641):

- Made --new-comment work with a single danger_id, so now we only get one, pretty Danger checkmark on your PR instead of multiple ones.
- Danger now always retrieves it's last comment (used to be the first one since it never had more than one showing up at the same time), so removing the argument on a subsequent build will now make it edit the right comment.


JuanitoFatas:
Rebase on master and we are good to go 👍

bruno-rocha-movile:
Merged!

orta:
Yep - cool, looks good to me 👍

PR #647, opened 2016-10-15
Support for Jenkins pipelines

fwal:
Added support for [Jenkins pipelines](https://jenkins.io/solutions/pipeline/).

Pipelines is a new way of configuring your jobs and is the preferred way in Jenkins 2.0+.
Unfortunately the old env vars isn't supported but on the other hand the vars should be uniform regardless of scm provider.

Currently works with:
- [x] GitHub
- [x] GitLab
- [x] BitBucket

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 * Add your own contribution below
+* Support for Jenkins pipelines - fwal
 
 ## 3.5.5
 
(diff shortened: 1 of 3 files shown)


orta:
Awesome :+1: - you may not need to actually do anything  got GitLab and Bitbucket

fwal:
Only need to fix line 83 in `lib/danger/ci_source/jenkins.rb`, I'm on it... 💪

orta:
Yeah - that makes sense

fwal:
Added support for GitLab and Bitbucket (in theory)

orta:
Then this looks good to me on green - thanks @fwal

marcelofabri:
Nice! I didn't even know about pipelines 💯

Issue #506, opened 2016-09-02
Migrate the dependency tree for gitlab into it's own gem


orta:
Similar to what we did with [danger-js-gem](https://github.com/danger/danger-js-gem) we should do the same thing for GitLab, and depending on where the bitbucket evolves to, the same for that

This requires documentation changes (it's not `danger` then, you would use `danger-gitlab`) as well as being a server bump.

This is not because I don't love GitLab, but to have the smallest footprint possible, the actual code would live in Danger still, but the `gitlab` gem dep would move out

orta:
https://github.com/danger/danger-gitlab-gem

PR #483, opened 2016-08-29
Sticky false by default

bogren:
Fixes #482. I agree with the issue, most of my dangerfile entries set sticky to false.

Issue #482, opened 2016-08-28
Proposal Switch messaging's sticky to default to false


orta:
When I looked at the Dangerfile for Danger, and Eigen etc, I see a lot of `sticky: false`. 

I think this is now the majority of the messages that I'm sending, and so I think it's time to move it back to the default.

marcelofabri:
Agreed 👍

k0nserv:
Yeah I agree seems like the better option

PR #654, opened 2016-10-26
Work for 4.0

orta:
Planning on only being https://github.com/danger/danger/issues/506 and https://github.com/danger/danger/pull/483 / https://github.com/danger/danger/issues/482

Will need subsequent docs updates too.


orta:
alright, that's all the pieces in place

PR #664, opened 2016-11-22
No blank lines in GitHub output

kivikakk:
Addresses https://github.com/danger/danger/issues/663 — specifically, GitHub's using a CommonMark-compliant parser now, so there are some stricter rules around HTML blocks in Markdown.

One thing I don't really like about this PR is the location of the new spec — so far everything in `comments_helper_spec` is about the helper itself and not about any particular `comment_generator`. Do you think maybe this would be better placed in a new spec file for comment generators? Feel free to push directly to this branch with adjustments!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Add your own contribution below
 
 * Fixed --new-comment not working with Bitbucket - Bruno Rocha
+* Adjust GitHub comment output for new Markdown parser - Yuki Izumi
 
 ## 4.0.0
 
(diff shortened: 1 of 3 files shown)


orta:
This looks 10/10 @kivikakk - I think this looks fine, yeah, it's a bit awkward but it's nothing too out of the ordinary ( and it fits the general gist )

Let's get this out into a release, we also invite anyone who has submitted a PR to join the org - you're welcome to join if you'd like?

kivikakk:
Awesome! And sure, I'd love to!

orta:
Rockin', sent

kivikakk:
Great, thank you! :smile_cat:

Issue #498, opened 2016-08-31
git.modified_files also includes "renaming indicators"


daniel-beard:
# Report
## What did you do?

Have a rule that checks for spaces in filenames:

```
# Warn on spaces in file names
####################################
files_with_spaces = git.modified_files.select { |file| (file =~ /\s/) != nil }
spaces_warning_text = ""
if files_with_spaces.length > 0 then
  spaces_warning_text += ""
  warn("Found filenames containing spaces, files should not contain spaces. See below for a list of files.", sticky: false)

  # Markdown below the list of failures
  markdown("##### Files with spaces")
  files_with_spaces.each { |file| markdown("- #{file}") }
end
```
## What did you expect to happen?

Should warn only when a file is added or modified with a space in the filename.
## What happened instead?
- Works well most of the time, I get correct warnings, but on a rename to fix the warning, modified files also includes the rename:

```
{Sub-projects/Source files => Modules/src}/UtilsTests/NSStringTemplateMakeTests.m
```
- Is there a way to only get the final filenames of modified files (without the renaming indication)?
## Your Environment

Jenkins, danger 3.1.0

orta:
Hrm, I think we should probably do some work in Danger to convert those to be usable file paths of the after in `git.modified_files`

orta:
https://github.com/danger/danger/pull/566 should fix this 👍

barakwei:
I hit this problem today. Danger 4.0.4.

daniel-beard:
thanks @barakwei, I recently ran into this issue again on version 4.2.2. @orta should there be a similar approach within Danger for `git.modified_files` ?

orta:
I think that makes sense, please do some tests to make sure that doesn't accidentally create a breaking change for most people also :D

barakwei:
Since the last comment i've upgraded it:
```ruby
modified_files = git.modified_files.map { |path| path.gsub(/{(.*) => (.*)}/, '\2') }
changed_files = (git.added_files + modified_files - git.deleted_files).uniq
```
Then I use `changd_files` with my plugins.

orta:
This got updated in 5.0.0

daniel-beard:
Thanks @orta I can confirm that this works great for me, thanks!

PR #681, opened 2016-12-17
Fixing handling of files moved to other directories in git.diff_for_file

nikolaykasyanov:
We've encountered a crash inside `git.diff_for_file` when PR/MR diff includes files moved to other folders.

It looks like [the code](https://github.com/danger/danger/blob/8e9a3ab12a78c518fac6dd917878fc3fc4f55b2a/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb#L115) is not enough to handle moves, here's how moved files are represented in `git.modified_files`:

```
Whatever/Path/{<old subpath> => <new subpath>}/File.m
```

But [@git.diff](https://github.com/danger/danger/blob/8e9a3ab12a78c518fac6dd917878fc3fc4f55b2a/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb#L115) doesn't have an entry for a such path, it contains an entry for the original path of the moved file.

The first idea was to detect such modified files, extract original file name and get the diff. The problem is, I couldn't find a reliable way to do that, because folder names can contain `=>` string too (see [failing test](https://github.com/nikolaykasyanov/danger/commit/2b66cf244da29d9449198b59620de38afc1abb0b) for an example).

I'd appreciate any guidance & advice.

- [x] Provide a failing test case
- [x] Provide a fix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ script:
   # Tests use real git commands
   - git config --global user.email "danger@example.com"
   - git config --global user.name "Danger McShane"
+  - git config --global diff.renames true
   - bundle exec rubocop lib
   - bundle exec rake spec
   - bundle exec danger --verbose
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ script:
   # Tests use real git commands
   - git config --global user.email "danger@example.com"
   - git config --global user.name "Danger McShane"
-  - git config --global diff.renames true
   - bundle exec rubocop lib
   - bundle exec rake spec
   - bundle exec danger --verbose
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## master
 
+* #681, #498 Fixes handling of files moved across directories by using diff to get list of modified files. Please note that the fix is not backward-compatible due to moved files names being the original file names.
 * Add your own contribution below
 
 ## 4.3.4
--- a/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
@@ -72,7 +72,7 @@ module Danger
     # @return [FileList<String>] an [Array] subclass
     #
     def modified_files
-      Danger::FileList.new(@git.diff.stats[:files].keys)
+      Danger::FileList.new(@git.diff.select { |diff| diff.type == "modified" }.map(&:path))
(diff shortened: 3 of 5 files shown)


nikolaykasyanov:
Another options is to use `@git.diff.select { |diff| puts diff.type == "modified" }` instead of `@git.diff.stats` to get the list of modified files, but I'm not sure about correctness and backward compatibility.

orta:
Oh yeah, this is a tricky one! 

I'm willing to break on backwards compatibility, I think I'm the only user or if so far.

nikolaykasyanov:
@orta I see. I'll try to tackle this one.

I noticed one more thing: test case depends on git in Travis runner. Apparently, older Git versions were reporting moved files as removed & added files.

orta:
Ah, now that one is tricky! Wonder if we need to check the git version and do different behavior if needed

nikolaykasyanov:
@orta it seems like c9b61eb67ee225fef0655f68ce9dc38601232de5 helps even with old Git (1.8.5.6 on Travis) versions.

Travis jobs still fail due to `The command "bundle exec danger --verbose" exited with 1.`, though, but nothing relevant in the logs.

barakwei:
Will this solve #498?

orta:
It could, yeah, I'm a bit wary on asking people to make git config changes to danger though

nikolaykasyanov:
@orta 

1. I believe it's a default behaviour since some Git version, so in most cases no changes are required.
2. Hopefully, we can pass this parameter while invoking Git, eliminating dependency on global configuration & Git version.

Am I missing something?

orta:
Ah - that makes sense, yeah 👍

nikolaykasyanov:
I believe this can be merged now. Should we elaborate more about backward-incompatibility in the changelog?

I made `diff.rename` setting as scoped as possible.

nikolaykasyanov:
Any ideas why does AppVeyor build fail?

orta:
I wonder if the appveyor version uses a different build of git? That failure is the same of some of the earlier commits

nikolaykasyanov:
@orta it looks like some of the git operations fail: https://ci.appveyor.com/project/orta/danger-47jiu/build/1.0.792#L817. Maybe Windows doesn't like paths with `=>`.

nikolaykasyanov:
Since we're not going to implement handling of paths like `Hello/ => World/file` in this PR anyway, I decided to simplify subfolder name to make it valid on Windows.

Issue #688, opened 2016-12-27
"undefined method `name' for nil:NilClass" when running "danger plugins"


barbosa:
danger.systems's [build](https://gitlab.com/danger-systems/danger.systems/builds/7637957) is breaking due to an error on `danger plugins` command. It executes `bundle exec rake generate`, which internally calls `bundle exec danger plugins json <plugin-name>` for each external plugin.

Investigating a little bit, I was able to find that [this return statement](https://github.com/danger/danger/blob/master/lib/danger/plugin_support/gems_resolver.rb#L19) is being called outside of the `Dir.chdir(dir)` block, which makes `paths` and `gems` methods run out of the temp dir scope. The solution could be either call `Dir.chdir(dir)` on both methods or pass the temporary dir as an argument.

# Report

## What did you do?

Run `bundle exec danger plugins json danger-prose`

## What did you expect to happen?

Output correct plugin metadata

## What happened instead?

```
bundler: failed to load command: danger (/Users/gustavo/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/bin/danger)
NoMethodError: undefined method `name' for nil:NilClass
  /Users/gustavo/Developer/danger/lib/danger/plugin_support/gems_resolver.rb:60:in `gem_metadata'
  /Users/gustavo/Developer/danger/lib/danger/plugin_support/gems_resolver.rb:47:in `block in gems'
  /Users/gustavo/Developer/danger/lib/danger/plugin_support/gems_resolver.rb:47:in `map'
  /Users/gustavo/Developer/danger/lib/danger/plugin_support/gems_resolver.rb:47:in `gems'
  /Users/gustavo/Developer/danger/lib/danger/plugin_support/gems_resolver.rb:19:in `call'
  /Users/gustavo/Developer/danger/lib/danger/plugin_support/plugin_file_resolver.rb:18:in `resolve'
  /Users/gustavo/Developer/danger/lib/danger/commands/plugins/plugin_json.rb:30:in `run'
  /Users/gustavo/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/claide-1.0.1/lib/claide/command.rb:334:in `run'
  /Users/gustavo/Developer/danger/bin/danger:5:in `<top (required)>'
  /Users/gustavo/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/bin/danger:23:in `load'
[...]

PR #689, opened 2016-12-29
Fixes the plugin JSON parser -

orta:
Fixes https://github.com/danger/danger/issues/688

And un-blocks Danger.Systems from deploying. This is obviously not the most elegant answer 👯 - but I'm happy enough, I did experiment with putting th return inside the chdir, but figured that was way more likely to break everything.

thanks @barbosa - who did the hard work on digging what was causing it

### DIFF
--- a/spec/lib/danger/plugin_support/gems_resolver_spec.rb
+++ b/spec/lib/danger/plugin_support/gems_resolver_spec.rb
@@ -50,7 +50,9 @@ RSpec.describe Danger::GemsResolver do
           fake_bundle_install_path_vendor_gems_in(spec_root)
         end
 
-        result = described_class.new(gem_names).call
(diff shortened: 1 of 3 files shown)


spec/lib/danger/plugin_support/gems_resolver_spec.rb
```diff
@@ -50,7 +50,9 @@ def fake_bundle_install_path_vendor_gems_in(spec_root)
           fake_bundle_install_path_vendor_gems_in(spec_root)
         end
 
-        result = described_class.new(gem_names).call
+        resolver = described_class.new(gem_names)
+        resolver.call
+        result = resolver.send(:all_gems_metadata)
```
orta: also a bit inelegant, but we can't call the original block here, I don't feel too bad calling the method after

PR #697, opened 2017-01-07
Fix danger pr command cannot specify API host

### LINKED ISSUE #691: `danger pr` doesn't respect env variables, unable to use with GH Enterprise
# Report

## What did you do?
`danger pr https://internalgithub.website.com/org/repo/pulls/111`

I expected `danger pr` to use `DANGER_GITHUB_API_HOST` to figure out what url to hit

## What happened instead?

It used `api.github.com` which does not work for GH enterprise instances.

```
/octokit-4.6.2/lib/octokit/response/raise_error.rb:16:in `on_complete': GET https://api.github.com/repos/org/repo/pulls/287: 401 - Bad credentials // See: https://developer.github.com/v3 (Octokit::Unauthorized)]
```

## Your Environment

Danger 4.0.4, ruby 2.3.0

JuanitoFatas:
Fixes #691.

After this PR, can specify API host by the following command:

```
$ DANGER_GITHUB_API_HOST=https://foo.bar.com https://foo.bar.com/org/repo/pulls/111
```

/cc @mtitolo

### DIFF
--- a/spec/lib/danger/ci_sources/support/pull_request_finder_spec.rb
+++ b/spec/lib/danger/ci_sources/support/pull_request_finder_spec.rb
@@ -103,5 +103,29 @@ RSpec.describe Danger::PullRequestFinder do
         expect(result.base).to eq "pr 518 base commit sha1"
       end
     end
+
+    context "specify api endpoint of octokit client" do
+      it "By DANGER_GITHUB_API_HOST" do
+        ENV["DANGER_GITHUB_API_HOST"] = "https://enterprise.artsy.net"
+
+        allow(Octokit::Client).to receive(:new).with(
+          access_token: ENV["DANGER_GITHUB_API_TOKEN"],
+          api_endpoint: "https://enterprise.artsy.net"
+        ) { spy("Octokit::Client") }
+
+        finder(pull_request_id: "42", remote: true, logs: "not important").call
+      end
+
+      it "fallbacks to DANGER_GITHUB_API_BASE_URL" do
+        ENV["DANGER_GITHUB_API_BASE_URL"] = "https://enterprise.artsy.net"
+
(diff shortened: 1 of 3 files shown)


spec/lib/danger/ci_sources/support/pull_request_finder_spec.rb
```diff
@@ -103,5 +103,29 @@ def open_pull_requests_info
         expect(result.base).to eq "pr 518 base commit sha1"
       end
     end
+
+    context "specify api endpoint of octokit client" do
+      it "By DANGER_GITHUB_API_HOST" do
+        ENV["DANGER_GITHUB_API_HOST"] = "https://enterprise.artsy.net"
+
+        allow(Octokit::Client).to receive(:new).with(
```
hanneskaeufler: Is there a specific reason you picked `allow` versus `expect` here (and below)?

orta: If I recall - Allow is used to say "mock theses upcoming functions" basically, and expect says "perform upcoming expectations on this object"

orta:
This looks 👍 to me

spec/lib/danger/ci_sources/support/pull_request_finder_spec.rb
hanneskaeufler: yeah I always regard `allow` as not an expectation, this would mean those two tests dont actually have an assertion in them (you could make it pass by removing all the production code from the method). However, I played around a bit with the tests and the way rspec seems to work this is not achievable so `allow` and `expect` is only a wording/intention difference here 👍

orta: nice one, thanks for the digging

Issue #684, opened 2016-12-20
Use GitHub Pull Request Reviews API to provide a Danger feedback


Antondomashnev:
Hi there, recently I had a thought, that It could be a great idea to integrate Danger more seamless into the GitHub workflow (I'm talking here specifically about GitHub) by using the review mechanism. So Danger could potentially __approve__, __comment__ or __request changes__ for the PR.

Before creating the proposal I was waiting for the API from GitHub team to do that, and finally, they released [it](https://developer.github.com/v3/pulls/reviews/) with a preview period.

Ideas, thoughts? 😄

orta:
Yeah, IMO, this can be an extension of the `github` object, perhaps something like:

```ruby
changelogName = false
unless changelogName 
  changelogFail = github.review.comment("changelog.md", 12, "Please include your name in the CHANGLOG")
  github.review.submit("reject", "test", [changelogFail])
end 
```

or something like

```ruby
github.review.start()

unless changelogName
  github.review.comment("changelog.md", 12, "Please include your name in the CHANGLOG")
end 

github.review.submit()
```

Whereby you declare the start and you keep adding comments until you fail. Then it posts the usual Danger comment as it's comment body.

orta:
Would also be nice to be able to specifically see if there are reviewers ( once that API is out ) for a PR

Antondomashnev:
@orta I personally find the second option nicer, would try to prepare PR soon.

orta:
👍

Antondomashnev:
I've started working on it and want to ask your opinion about architecture:

I've selected this approach your proposed
```ruby
github.review.start()

unless changelogName
  github.review.comment("changelog.md", 12, "Please include your name in the CHANGLOG")
end 

github.review.submit()
```

And I want to extend it with the following interface for `review` object
```ruby
# This adds a comment to the review body
def comment(message, type) where type in ["message", "warning", "failure"]

# This adds an inline comment to review
def inline_comment(file, line, message, type) where type in ["message", "warning", "failure"]

# This submits a review with one of the following event: ["approve", "request changes", "comment"]
# If there are failure comments or inline comments -> "request changes" otherwise -> "approve"
def submit
```

@JuanitoFatas @orta what do you think about that proposal?

orta:
Ideally WRT the param order  it should follow the same syntax as the rest of the messaging:

```
Print markdown to below the table
markdown (message: String, file=nil: String, line=nil: String)

Print out a generate message on the PR
[...]

PR #702, opened 2017-01-22
GitHub pull request review

Antondomashnev:
Hi, there this PR addresses an issue I opened a while ago #684. 
The status is kinda work in progress, but there is already functionality that can be potentially used.

The changes in this PR:
* new module `GitHubSource` to combine the only GitHub related features;
* new class `Review` under the `GitHubSource` module, which represents the pull request review object;
* extension to `octokit` which I think will be used until the API is not final and then I will create a PR to octokit itself with extension for pr reviews;
 
Here is an example of how it looks like in `Dangerfile`:
```ruby
github.review.start
github.review.fail("I can fail your PR in a moment")
github.review.warn("I'm warning yo")
github.review.message("100 message")
github.review.submit
```

You can check how it looks in a PR [here](https://github.com/Antondomashnev/MyAmazingDangerPlayground/pull/7)

The open points I want to discuss:
1. The overall design approach
2. Do we need the summary message in review? My intention to create it was to write something if the PR is well done and danger approves it without any comment.
3. Do you think it could be used without inline comments as a first step or you prefer the feature with inline comments?

P.S. the changelog is not modified yet and I haven't cleaned the commit history yet.

### DIFF
--- a/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
@@ -93,9 +93,22 @@ module Danger
     end
 
     # @!group PR Review
-    # Current PR Review object.
-    # @return [GitHubReview]
     #
+    # In Beta. Provides access to creating a GitHub Review instead of a typical GitHub comment.
+    #
+    # To use you announce the start of your review, and the end via the `start` and `submit` functions,
+    # for example:
+    #
+    # ```
+    #   github.review.start
+    #   github.review.fail("Please add a CHANGELOG entry") if has_no_changelog
+    #   github.review.warn("Highway to the Danger Zone") if pr_includes_word_danger
+    #   github.review.message("You might want to read #{url}") if may_require_docs
+    #   github.review.markdown("Please update your changelog entry according an #{example}") if changelog_format_not_valid
+    #   github.review.submit
+    # ```
+    #
+    # @return [ReviewDSL]
     def review
       @github.review
     end
--- a/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
@@ -92,6 +92,14 @@ module Danger
       "github"
     end
 
+    # @!group PR Review
+    # Current PR Review object.
+    # @return [GitHubReview]
+    #
+    def review
+      @github.review
+    end
+
     # @!group PR Metadata
     # The title of the Pull Request.
     # @return [String]
--- a/lib/danger/request_sources/github/github_review.rb
+++ b/lib/danger/request_sources/github/github_review.rb
@@ -42,7 +42,7 @@ module Danger
         end
 
         def status
-          return STATUS_PENDING unless self.review_json
+          return STATUS_PENDING if self.review_json.nil?
           return self.review_json["state"]
         end
 
@@ -55,10 +55,15 @@ module Danger
 
         def submit
           # We wanna add final markdown with the pull request review status such as Danger uses for pull request status
-          
-          submission_event = generate_event(generate_general_violations)
+          general_violations = generate_general_violations
+          @markdowns << Markdown.new(message, generate_description(warnings: general_violations[:warnings], errors: general_violations[:errors]))
+
+          submission_event = generate_event(general_violations)
           submission_body = generate_body
-          return unless ReviewResolver.new(self).should_submit?(submission_event, submission_body)
+
+          # If the review resolver says that there is nothing to submit we skip submission
+          return unless ReviewResolver.should_submit?(self, submission_event, submission_body)
+
           self.review_json = @client.create_pull_request_review(@ci_source.repo_slug, @ci_source.pull_request_id, submission_event, submission_body)
         end
 
@@ -88,14 +93,14 @@ module Danger
 
         private
 
+        # The only reason to request changes for the PR is to have errors from Danger
+        # otherwise let's just notify user and we're done
         def generate_event(violations)
           violations[:errors].empty? ? EVENT_APPROVE : EVENT_REQUEST_CHANGES
         end
 
         def generate_body(danger_id: "danger")
           previous_violations = parse_comment(body)
-          puts "Previous violations #{previous_violations} for body #{body}"
-
           general_violations = generate_general_violations
           new_body = generate_comment(warnings: general_violations[:warnings],
                                       errors: general_violations[:errors],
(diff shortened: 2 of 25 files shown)


lib/danger/request_sources/github/github_review_resolver.rb
```diff
@@ -0,0 +1,27 @@
...
+  module RequestSources
+    module GitHubSource
+      class ReviewResolver
+        def self.should_submit?(review, event, body)
+          return false if same_body?(body, review.body)
+          return event_changes_status?(event, review.status)
+        end
+
+        # @orta @JuanitoFatas to be discussed whether we should submit review with
+        # the same status as already submitted by danger before
```
orta: on an PR update? I think that's 👍

lib/danger/request_sources/github/octokit_pr_review.rb
```diff
@@ -0,0 +1,83 @@
...
+    def submit_pull_request_review(repo, pull_request_number, review_id, event, body = nil, options = {})
+      review = {
+        event: event,
+        accept: CUSTOM_ACCEPT_HEADER
+      }
+      review[:body] = body unless body.nil?
+      post "#{Repository.path repo}/pulls/#{pull_request_number}/reviews/#{review_id}/events", options.merge(review)
+    end
+  end
+end
```
orta: nice one

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
```diff
@@ -92,6 +92,14 @@ def self.instance_name
       "github"
     end
 
+    # @!group PR Review
+    # Current PR Review object.
+    # @return [GitHubReview]
+    #
```
orta: This will need considerable documentation before we ship ( as this is what an end user would see on the website ) - I'm happy to do this once you're ready though 👍

lib/danger/request_sources/github/github_review.rb
```diff
@@ -0,0 +1,132 @@
...
+require "danger/helpers/comments_helper"
+require "danger/helpers/comment"
+
+module Danger
+  module RequestSources
+    module GitHubSource
+      class Review
+        include Danger::Helpers::CommentsHelper
+
+        # @see https://developer.github.com/v3/pulls/reviews/ for all possbile events
```
orta: > possbile

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
Antondomashnev: Sure, do you think we should provide an example here or what are your thoughts? 😄

orta: Sure thing, something like this

```
# @!group PR Review
#
# In Beta. Provides access to creating a GitHub Review instead of a typical GitHub comment.
#
# To use you announce the start of your review, and the end via the `start` and `submit` functions,
# for example:
#
# ```
#   github.review.start
#   github.review.fail("Please add a CHANGELOG entry") if has_no_changelog
#   github.review.warn("Highway to the Danger Zone") if pr_includes_word_danger
#   github.review.message("You might want to read #{url}") if may_require_docs
#   github.review.submit
# ```
#
# @return [GitHubReview]
```

orta: People don't see the inline documentation for the `review` object, so it needs to be attached to this DSL attribute to end up on the reference - http://danger.systems/reference.html

lib/danger/request_sources/github/github_review_resolver.rb
Antondomashnev: Done in 5730d68

lib/danger/request_sources/github/github_review.rb
Antondomashnev: Fixed in 5730d68

orta:
> The overall design approach

This feels 👍 to me

> Do we need the summary message in review? My intention to create it was to write something if the PR is well done and danger approves it without any comment.

I think approval without comment is perfectly fine IMO, if this does't feel right, we can change that at a later date. Note you should check if that's possible before making the assumption it is possible.

> Do you think it could be used without inline comments as a first step or you prefer the feature with inline comments?

Fine by me

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
Antondomashnev: @orta looks cool 👍 
The small feedback: 
1. `@return [GitHubReview]` -> `@return [GitHubSource::Review]` or we can just leave `Review`.
2. Should we mention `github.review.markdown(...)` ?

Antondomashnev:
> I think approval without comment is perfectly fine IMO, if this doesn't feel right, we can change that at a later date. Note you should check if that's possible before making the assumption it is possible.

Currently, it looks like this for example:

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
orta: > @return [GitHubReview] -> @return [GitHubSource::Review] or we can just leave Review.

IMO `@return Review` is fine, maybe even `@return ReviewDSL` as these docs are for end-users who don't care about our internal semantics

> Should we mention github.review.markdown(...) ?

Yep :D - please make it a real-ish example like the ones above

orta:
Yeah, if a review can be posted with 0 comments, I think it's worth adding

orta:
Yeah, like this one ^

JuanitoFatas:
Rubocop is not happy 👮  Other than offenses, 👍  Great work!

Antondomashnev:
@JuanitoFatas yep, I've noticed it, but I'm a bit confused now why it's green on my local machine with the same command and as far as I understand according to `.rubocop.yaml` it should exclude spec folders, but Travis finds the offenses actually there.
Do you have any idea why it could happen?

orta:
> Do you have any idea why it could happen?

I've always wondered if Rubocop actively ignores the fact that you're using bundler with it

Antondomashnev:
Aha, that could be the case https://github.com/bbatsov/rubocop/issues/3968

Antondomashnev:
Yes the latest rubocop release of 0.47.1 was the issue.

orta:
Alright - this looks good to me, lets get this shipped.

orta:
Awesome - shipped as 4.0.1

Antondomashnev:
@orta thanks, could you explain why the `continuous-integration/appveyor/pr` is red? Never worked with it 😄

orta:
It runs our CI on windows, it's been having issues on all my OSS - might be capacity issues

PR #707, opened 2017-01-26
Minor Docs fix

orta:
Nothing too important


orta:
True, they were not. I'm alright with this

PR #711, opened 2017-01-30
Use colored2 instead of colored

orta:
Handles the same transition seen in https://github.com/CocoaPods/Xcodeproj/pull/463


orta:
Yeah, let's do it.

Issue #701, opened 2017-01-15
Add a Danger comment preview summary


orta:
# Report

Noted in https://github.com/danger/danger-js/issues/98

## What did you do?

Let Danger write a comment, see a slack notification like:

![screen shot 2017-01-15 at 11 50 14 am](https://cloud.githubusercontent.com/assets/49038/21964383/d4423cb2-db18-11e6-9991-1fccdeb7b825.png)

This ain't optimal 

## What did you expect to happen?

A proposal I had was. Make the first 5 lines a hidden summary in HTML, take the first ~140 chars if 1 of the types and summarize if more.

```
<!-- Danger report:
1 Fail: No Changelog added. You can find it at...
3 Warnings.
2 Markdown notices.
-->
<table>
  <thead>
```

## Your Environment

* Danger 4.0.5

PR #714, opened 2017-02-02
Hk comment preview summary

hanneskaeufler:
I totally like the idea of this html comment summary to show up in for example slack. As proposed in #701 and available in danger/danger-js#101.  This PR will add a 

```html
<!--
  1 Error: This is an error
  1 Warning: Violations that are very very ...
  1 Message: This is a message
  1 Markdown
--!>
```

comment to the top of dangers comment. What do you guys think?

### DIFF
--- a/lib/danger/core_ext/string.rb
+++ b/lib/danger/core_ext/string.rb
@@ -12,4 +12,9 @@ class String
       tr("-".freeze, "_".freeze).
       downcase
   end
+
+  # @return [String] truncates string with ellipsis when exceeding the limit
+  def danger_truncate(limit)
+    length > limit ? "#{self[0...limit]}..." : self
+  end
 end
--- a/spec/lib/danger/helpers/comments_helper_spec.rb
+++ b/spec/lib/danger/helpers/comments_helper_spec.rb
@@ -379,7 +379,8 @@ COMMENT
         warnings: [],
         errors: violations(["an error"]),
         messages: [],
-        previous_violations: previous_violations)
+        previous_violations: previous_violations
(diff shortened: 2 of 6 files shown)


spec/lib/danger/helpers/comments_helper_spec.rb
```diff
@@ -275,29 +297,29 @@ class Dummy
   describe "#generate_comment" do
     it "no warnings, no errors, no messages" do
       result = dummy.generate_comment(warnings: [], errors: [], messages: [])
-      expect(result.gsub(/\s+/, "")).to eq(
+      expect(result.gsub(/\s+/, "")).to end_with(
```
hanneskaeufler: Not excited about this, but it was a pain to update those verifications after each change I made during implementing the summary. Most of those verifications should not really care about the summary? Matching on a regexp was also not something I wanted to do so I copped out with `end_with`.

orta: Yeah, this is  👍 to me

lib/danger/core_ext/string.rb
```diff
@@ -12,4 +12,9 @@ def danger_underscore
       tr("-".freeze, "_".freeze).
       downcase
   end
+
+  # @return [String] truncates string with ellipsis when exceeding the limit
+  def danger_truncate(limit)
+    length > limit ? "#{self[0...limit]}..." : self
+  end
```
orta: 👍

orta:
I got some time, I can make a release, let's give it a go

PR #728, opened 2017-02-09
[wip] Add test for danger local

hanneskaeufler:
As those `pr` and `local` command are some seriously untested territory, I wanted to see how bad it was going to be to add an integration test. So far pretty bad :) Turns out theres loads of requests, I didnt expect that.

- [ ] Does this even make sense?
- [ ] Why does webmock not stub the request to rubygems?
- [ ] Wow the entire danger Dangerfile is ran (duh) which means contributors are fetched from the github api
- [x] Aha, on circle ci there is probably some env var set to make the host be `https://enterprise.artsy.net`?


orta:
I'm still not entirely convinced that these tests make sense, at least every time I've tried doing them, they've turned out brittle and we've removed them at a later date.

IMO, best move would be to refactor as much of the `danger local` code into another class, and test that class - with the local command only really setting up the params. That's what we did with `danger run` (the default) and using the `EnvironmentManager`

hanneskaeufler:
Moving to different approach in #729

PR #729, opened 2017-02-10
Hk danger local extract all the things

hanneskaeufler:
As discovered in #728 and pointed out by orta, adding highly integrated tests for `danger local` might not be the best approach. This pr goes the other route and extracts some helper classes which can be unit tested instead. `DangerLocal#run` is now _mostly_ a shell around those classes performing the meat of the work.

- [x] Naming
- [x] `DangerfileEvaluator` has major feature envy on the `Dangerfile`
- [x] ~~Bit of a weird mix of fakes and rspec doubles in the new tests~~ Maybe revisit this later

*this branch* 
2218 / 2799 LOC (79.24%)
*master*
2192 / 2794 LOC (78.45%) covered.


orta:
❤️

lib/danger/commands/local_helpers/local_setup.rb
```diff
@@ -0,0 +1,41 @@
+module Danger
+  class LocalSetup
```
orta: I think this naming makes sense :+1:

spec/lib/danger/commands/local_helpers/local_setup_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  class FakeCiSource
+    attr_reader :repo_slug, :pull_request_id
+
+    def initialize(repo_slug, pull_request_id)
+      @repo_slug = repo_slug
+      @pull_request_id = pull_request_id
+    end
+  end
+end
```
orta: Ah yeah, this is  much better :D

orta:
This is good to go when you're happy

hanneskaeufler:
I am okay with it now for the moment :octocat:

PR #735, opened 2017-02-15
Fix inline comment url

leonhartX:
The inline comment seems to use a fixed `https://github.com/` url regardless the `DANGER_GITHUB_HOST` setting, when using github enterprise, the inline comment fails and the link in comment is broken.

Try to use `host` in the url instead of `https://github.com/`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 -* Add your own contribution below
+* Fix inline comment url when using github enterprise
 
 ## 4.2.2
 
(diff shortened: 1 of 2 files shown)


orta:
Oh yeah this is great - I think all these fails are legitimate, I wonder why - [master seems fine](https://circleci.com/gh/danger/danger/1955)

anyone got any ideas?

hanneskaeufler:
I checked this out locally and it was all green. then I bundle updated and it was broken just like on circle.

hanneskaeufler:
Shortterm fix is in #737, one should probably investigate whats going on there.

orta:
Perfect, I've rebased this with all the changes - if this greens it's good to go

orta:
also,  @leonhartX I've invited you to the org, as a part of [Moya Community Continuity](https://github.com/Moya/contributors/issues/6) - you're welcome to participate in PRs and help out at a level that works for you.

PR #737, opened 2017-02-15
add changelog entry for pinning gitlab

hanneskaeufler:
Tests in #735 are failing because of a change in NARKOZ/gitlab that breaks this gem for us. This PR will exhibit the breaking build, then pin the version to the last known good commit.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ if RUBY_VERSION == "2.3.1"
 end
 
 # This should get removed when > 3.7.0 comes out
-gem "gitlab", git: "https://github.com/NARKOZ/gitlab.git", branch: "master"
+gem "gitlab", git: "https://github.com/NARKOZ/gitlab.git", branch: "master", :ref => "0cc49d008c6448d652fa27b3f9b3f60202dffab3"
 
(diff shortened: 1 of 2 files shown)


hanneskaeufler:
Yeah so this is currently broken, I only added the changelog entry to see it fail first. Now if I make the commit to pin the gitlab gem to `0cc49d008c6448d652fa27b3f9b3f60202dffab3` it should be green.

Gemfile
```diff
@@ -9,7 +9,7 @@ if RUBY_VERSION == "2.3.1"
 end
 
 # This should get removed when > 3.7.0 comes out
-gem "gitlab", git: "https://github.com/NARKOZ/gitlab.git", branch: "master"
+gem "gitlab", git: "https://github.com/NARKOZ/gitlab.git", branch: "master", :ref => "0cc49d008c6448d652fa27b3f9b3f60202dffab3"
```
hanneskaeufler: So I guess depending on master is biting us here :) There has not been a release after 3.7.0 which we would want as the comment points out.

hanneskaeufler:
Ah note that I rather think #738 is the way to go here, I think I wann close this.

PR #738, opened 2017-02-15
Gitlab project names dont need to be urlencoded

hanneskaeufler:
https://github.com/NARKOZ/gitlab/pull/253 changed encoding behaviour, and we depend on master branch of the gem. So we need to run after it 💃 This is whats failing in #735. And alternative approach to fixing this was taken in #737, though I think I prefer this one here 🍡


orta:
Perfect - awesome, thanks!

PR #740, opened 2017-02-16
only find diff position in current file

leonhartX:
what did i do:
- move `file_header_regexp` check to the beginning of `each` block. 
- make `file_header_regexp` more strict to avoid unexpected match.

For the following code:
https://github.com/danger/danger/blob/master/lib/danger/request_sources/github/github.rb#L360-L364

` line.match file_header_regexp` will never be matched since the `line`  must match to `range_header_regexp `. to reach this line of code.
And then `find_position_in_diff` will fall down to other files to find the diff line. when the other files happened to has a range just match the message's line number, a wrong inline comment will be created and finally caught an error from github.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 -* Add your own contribution below
+* Fix find wrong diff position for inline comment
 
 ## 4.2.2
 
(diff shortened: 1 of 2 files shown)


orta:
Cool, this looks good to me 👍

PR #746, opened 2017-02-20
fix inline comment's fall back bugs

leonhartX:
Fix two bugs with inline comment:

- when violations is all inline comment, those message which is out of range will not shown  in main comment due to the wrong check of `main_violations`
- when there is only makrdowns inline comment, it won't be post as inline.


orta:
👍

Issue #742, opened 2017-02-17
Support for only show message in diff


leonhartX:
Hi  there, currently the `inline comment` will try to set message directly to the line in diff, when it's out of the diff's range, goes back to a comment.
But sometimes I just want to let them show in diff, when out of range, it means we don't care them in this pull request.(something like `swiftlint`).

I'm trying to work on this, add an option like `strict_inline` to the message plugin. 
Any ideas about this?

orta:
I think the API specifically does not allow this, which is why we have to do such a weird dance around putting it in the main message

leonhartX:
yes, so can we have an option to ignore those out-of-diff messages when they are out-of-my-concern too, like i don't want the whole file's lint violations result but only in my changes?

orta:
oh, I see, interesting. 

I can see a case when you still have a lot of ignored warnings that could be outside of that range. Sure, I think that makes sense as an option on messaging, perhaps `messaging.dismiss_out_of_range_violations = true` ? yeah, it's verbose, but it makes it very obvious

manicmaniac:
Fixed in https://github.com/danger/danger/pull/747

PR #747, opened 2017-02-20
Only show inline in diff

leonhartX:
a simple approach for #742.
since the inline message is currently only available in github, add an property for `github request_source` and an interface in `github plugin` to set it.

### DIFF
--- a/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
@@ -224,10 +224,10 @@ module Danger
       paths.first(paths.count - 1).join(", ") + " & " + paths.last
     end
 
-    # @!group Github Misc
-    # Tell github to ingore any inline message whihc is not in diff's range, not post as main comment.
+    # @!group GitHub Misc
+    # Use to ignore inline messages which lay outside a diff's range, thereby not posting them in the main comment.
     # @param    [Bool] dismiss
-    #           Ingore out of range inline messages, defaults to `true`
+    #           Ignore out of range inline messages, defaults to `true`
     #
     # @return   [void]
     def dismiss_out_of_range_messages(dismiss: true)
--- a/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
@@ -224,6 +224,16 @@ module Danger
       paths.first(paths.count - 1).join(", ") + " & " + paths.last
     end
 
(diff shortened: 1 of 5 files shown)


lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
```diff
@@ -224,6 +224,16 @@ def html_link(paths, full_path: true)
       paths.first(paths.count - 1).join(", ") + " & " + paths.last
     end
 
+    # @!group Github Misc
+    # Tell github to ingore any inline message whihc is not in diff's range, not post as main comment.
```
orta: Capital "H" in GitHub please, for "Github Misc"

Nice work on the docs! What about a small evolution: "Use to ignore inline messages which lay outside a diff's range, thereby not posting them in the main comment." instead?

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
```diff
@@ -224,6 +224,16 @@ def html_link(paths, full_path: true)
...
+    # @param    [Bool] dismiss
+    #           Ingore out of range inline messages, defaults to `true`
```
orta: `Ingore`

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
```diff
@@ -224,6 +224,16 @@ def html_link(paths, full_path: true)
...
+    #
+    # @return   [void]
+    def dismiss_out_of_range_messages(dismiss: true)
```
orta: Please make this default to false, I would expect that to be the most common experience. I would always want to know all the warnings/errors to force people to actually fix them ;)

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
leonhartX: thanks for the advice! i'm not so good at english so it's really helpful.

lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
leonhartX: since this method name is **dismiss**_out_of_range_messages, i think when someone calls it without any arguments, he will expect this to dismiss messages as the name's mean.
how do you think about this? maybe change to another name is better?

orta: ah, no that makes sense! Yeah, nah - this is cool to stay

orta:
👍

PR #752, opened 2017-02-28
fix wrong inline comment position at last line

leonhartX:
in `find_position_in_diff`, the diff hunk's `range_end` should not contain the last line.
when a inline comment happened to be found at the `range_end` line, it get an invalid position which is out of the diff.

like this:
diff

the range is 9 - 15, not 9 - 16(which is `match[:start] + match[:end]`).


orta:
👍

PR #755, opened 2017-03-07
Add inline comments to changed lines only

leonhartX:
With the `dismiss_out_of_range_messages` options, only comment within the diff's range will shown as inline comment.
But in diff's range does not mean it's a change since diff contains unchanged lines. so also ignore those comments at unchanged line too when `dismiss_out_of_range_messages` is enabled.


orta:
Cool, looks 👍  to me

PR #757, opened 2017-03-10
Fix a wrong doc in dangerfile_danger_plugin

duck8823:
In import_dangerfile, opts `file:` does not work, but `path:` works.


orta:
Great - thanks~

orta:
bah, on a mobile so can't merge, will do so when I get the chance

PR #761, opened 2017-03-19
Add buddybuild documentation.

sunshinejr:
Hey guys! 🎉

So this is my first PR here, and first in ruby, so I'm sorry for everything wrong here :D

Basically, Danger can be used with buddybuild and it is pretty easy! But on the website there was no info about it, so I've decided to make a new `ci_source` just with documentation. Dunno if the ruby code is correct here, so I'd love a review for that. I've also added a caveat that you can change the bot that is used by Danger (simply by adding a new token to the ENV vars), but I had the chance of testing it only using GitHub project - but you can also create projects using GitLab and Bitbucket. If anyone has any info about these two, it would be awesome as well.

Cheers! 

**Edit:** Also I'm not sure, should I write test spec for this class?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Add your own contribution below
 * Fix find inline comment position at the last line of diff - leonhartX
 * Add inline comments to changed lines only when `dismiss_out_of_range_message` is enabled - leonhartX
+* Add documentation for [buddybuild](https://www.buddybuild.com) - sunshinejr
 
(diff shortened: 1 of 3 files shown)


orta:
Nah no need for tests, it's existence is enough 👍

sunshinejr:
Huh, that was clean! 😅 Thanks @orta! 💪

orta:
Shipped in https://github.com/danger/danger/pull/762

PR #773, opened 2017-04-04
[Fix] Try make sure we use bundled rubocop #trivial

orta:
Force us to use the bundle'd rubocop


JuanitoFatas:
CI fixed 🙏

orta:
Ah, that worked? Awesome. Let's ship it.

PR #781, opened 2017-04-16
Prepare for v5.0.0

orta:
///////////////////⚡///////////////////

fortune-global-500-2016-logo

///////////////////⚡///////////////////

I know, major version bump, but it's a breaking change and we are doing real semver.


nikolaykasyanov:
👍

PR #783, opened 2017-04-17
Fix typo in DangerfileGitPlugin example

a2:


orta:
Hah - Danger is not happy with you, I'm fine with those error - thanks!

PR #787, opened 2017-04-17
Bbs network call fix

RishabhTayal:
///////////////////⚡///////////////////

# 🚫 AWESOME A PR!

Fixes a bug where `use_ssl` was called for url without `https://`.

///////////////////⚡///////////////////

PR #788, opened 2017-04-17
Fix diff_for_file for added files

allewun:
Looks like a previous change caused `git.diff_for_file(file)` to return `nil` if the file was an _added file_.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Add your own contributions to the next release on the line below this, please include your name too.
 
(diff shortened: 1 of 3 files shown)


orta:
This looks good - can you please add a CHANGELOG, and I'll get this and #787 into a 5.0.2 - thanks!

allewun:
Thanks for the amazingly quick turnaround @orta!

orta:
Thank. you for the succinct and tested PR 👯‍♂️ - makes it easy to work with

orta:
Shipped in 5.0.2

PR #792, opened 2017-04-18
Gitlab gem update + Make chandler run on `rake release`

orta:
Somehow the travis deploy script wasn't working, so now I've improved the `rake release` task to also run chandler if you have either one of those ENV vars in your shell.

### DIFF
--- a/spec/support/gitlab_helper.rb
+++ b/spec/support/gitlab_helper.rb
@@ -27,19 +27,19 @@ module Danger
 
       def stub_merge_request(fixture, slug, merge_request_id)
         raw_file = File.new("spec/fixtures/gitlab_api/#{fixture}.json")
-        url = "https://gitlab.com/api/v3/projects/#{slug}/merge_request/#{merge_request_id}"
+        url = "https://gitlab.com/api/v3/projects/#{slug}/merge_requests/#{merge_request_id}"
         WebMock.stub_request(:get, url).with(headers: expected_headers).to_return(raw_file)
       end
 
       def stub_merge_request_changes(fixture, slug, merge_request_id)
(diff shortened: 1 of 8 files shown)


orta:
@k0nserv - does this look fine to you?

orta:
And @connorshea - these api route changes feel right to you?

```diff
-        url = "https://gitlab.com/api/v3/projects/#{slug}/merge_request/#{merge_request_id}"
 +        url = "https://gitlab.com/api/v3/projects/#{slug}/merge_requests/#{merge_request_id}"
```

spec/support/gitlab_helper.rb
```diff
@@ -27,19 +27,19 @@ def stub_request_source
 
       def stub_merge_request(fixture, slug, merge_request_id)
         raw_file = File.new("spec/fixtures/gitlab_api/#{fixture}.json")
-        url = "https://gitlab.com/api/v3/projects/#{slug}/merge_request/#{merge_request_id}"
```
k0nserv: I'm surprised that this worked at all, seems like I made a mistake here

orta: I think the gitlab gem was using that route, and as of 4.0.0 it now does the `s`

k0nserv: So in the API both worked?

orta: That's what I was hoping @connorshea would know :D

orta: I expect people are using both 3.x and 4.x in production ( as we migrated Danger last week) - I found the issue, looks correct https://github.com/NARKOZ/gitlab/issues/229

connorshea: yep, I think we should probably move to v4 soon but for now this is fine :)

PR #797, opened 2017-04-21
[Drone] Support old and new version

k0nserv:
This reinstates the support for version pre 0.6 while still
maintaining support for version 0.6.


fmartingr:
@k0nserv Good approach, thank you for this.

PR #799, opened 2017-04-26
[Gitlab] Fetch all MR comments

k0nserv:
Previously we only fetched the first page which means
that when the danger comment is not on the first page it
adds new comments instead of updating the existing one.


k0nserv:
💩 need to fix the tests too

orta:
👍

PR #808, opened 2017-05-05
Fix danger_id and new_comment flag not working

4oby:
Fix when running `danger --new-comment --danger_id="id" pr <link>`
the flags `--new-comment` and `--danger_id` where set to `nil`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 * Add your own contributions to the next release on the line below this, please include your name too. Please don't set a new version if you are the first to make the section for `master`.
+* Fix danger_id and new_comment flag not being set for pr checks [@4oby](https://github.com/4oby)
 * Add support for DotCi - Daniel Beard
 
 ## 5.1.1
--- a/lib/danger/commands/pr.rb
+++ b/lib/danger/commands/pr.rb
@@ -24,15 +24,18 @@ module Danger
       @pr_url = argv.shift_argument
       @clear_http_cache = argv.flag?("clear-http-cache", false)
       dangerfile = argv.option("dangerfile", "Dangerfile")
-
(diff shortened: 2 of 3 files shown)


orta:
Danger `pr` doesn't actually post comments - it's for running your local Dangerfile against a PR and shows the output in your terminal. You want `danger` or `danger run` for commenting on a PR.

orta:
Or does it just not work without those?

4oby:
It works, but does not post, `danger pr` is easier, IMO, to hook into our current Jenkins setup.

orta:
You should be using `danger` or `danger run` for that. 

Adding the ability to run danger from CI via `danger pr` will make it more confusing to understand what commands to use. Can you explain why it's easier to use Danger PR?

4oby:
Jenkins is setup to run on a detached head, so any attempts to run `danger` ends in "not a pr" message.

nikolaykasyanov:
@4oby this message could mean that something important is missing in the environment (PR identifier, for example).

I'm using Danger on Jenkins with detached head builds, haven't had such message with properly configured environment.

orta:
Yeah, this does sound more like your Jenkins setup needs the env vars I think. The check looks [like this](https://github.com/danger/danger/blob/master/lib/danger/ci_source/jenkins.rb#L74).

mihai8804858:
We finally configured danger to run on our Jenkins. The problem was that `"ghprbPullId"` environment variable was not exported. I saw that this variable is exported automatically by `GitHub pull request builder plugin`, but we're not using this plugin because it's not compatible with our job's configuration, so we manually export this environment before running danger. And it works!
Thank you @orta and @nikolaykasyanov for your help.
@4oby I think we can close this PR.

orta:
👍

PR #811, opened 2017-05-08
Add ability to get list of renamed files

sleekybadger:
## Summary

Hi. This add ability to access list of renamed files via `git.renamed_files`.

## Example

`git.renamed_files` will return Array with Hashes that have keys `:before` and `:after`.
```
[
  { before: "README", after: "README.MD" },
  { before: "some/path/file", after: "some/other/path/file" }
]
```

## Notes

Unfortunately [git gem](https://github.com/schacon/ruby-git) that we are using, has tough limited api which don't provide ability to pass some options for diff output customization. 

So I decided to use `git diff  --diff-filter=R`, which returns something like this:

```
[
  "diff --git a/README b/README.MD\n", 
  "similarity index 100%\n", 
  "rename from README\n", 
  "rename to README.MD\n", 

  "diff --git a/hello1 b/hello1_renamed\n", 
  "similarity index 100%\n", 
  "rename from hello1\n", 
  "rename to hello1_renamed\n", 

  "diff --git a/hello3 b/hello3_renamed\n", 
  "similarity index 100%\n", 
  "rename from hello3\n", 
  "rename to hello3_renamed\n", 

  "diff --git a/hello5 b/hello5/some_folder/hello5_renamed\n", 
  "similarity index 100%\n", 
  "rename from hello5\n", 
  "rename to hello5/some_folder/hello5_renamed"
]
```

Maybe it makes sense to move from [git gem](https://github.com/schacon/ruby-git) to [rugged gem](https://github.com/libgit2/rugged), cause it better maintained and looks much more flexible?

### DIFF
--- a/lib/danger/scm_source/git_repo.rb
+++ b/lib/danger/scm_source/git_repo.rb
@@ -19,6 +19,32 @@ module Danger
       self.log = repo.log.between(from, to)
     end
 
+    def renamed_files
+      # Get raw diff with --find-renames --diff-filter
+      # We need to pass --find-renames cause
+      # older versions of git don't use this flag as default
+      diff = exec(
+        "diff #{self.diff.from} #{self.diff.to} --find-renames --diff-filter=R"
+      ).lines.map { |line| line.tr("\n", "") }
+
+      before_name_regexp = /^rename from (.*)$/
+      after_name_regexp = /^rename to (.*)$/
+
+      # Extract old and new paths via regexp
+      diff.each_with_index.map do |line, index|
+        before_match = line.match(before_name_regexp)
+        next unless before_match
+
+        after_match = diff.fetch(index + 1, "").match(after_name_regexp)
+        next unless after_match
+
+        {
+          before: before_match.captures.first,
+          after: after_match.captures.first
+        }
+      end.compact
+    end
+
     def exec(string)
       require "open3"
       Dir.chdir(self.folder || ".") do
--- a/spec/lib/danger/scm_source/git_repo_spec.rb
+++ b/spec/lib/danger/scm_source/git_repo_spec.rb
@@ -195,4 +195,50 @@ RSpec.describe Danger::GitRepo, host: :github do
       end
     end
   end
+
+  describe "#renamed_files" do
+    it "returns array of hashes with names before and after" do
+      Dir.mktmpdir do |dir|
+        Dir.chdir dir do
+          `git init`
+          `git remote add origin git@github.com:danger/danger.git`
+
+          Dir.mkdir(File.join(dir, "first"))
+          Dir.mkdir(File.join(dir, "second"))
+
+          File.open(File.join(dir, "first", "a"), "w") { |f| f.write("hi") }
+          File.open(File.join(dir, "second", "b"), "w") { |f| f.write("bye") }
+          File.open(File.join(dir, "c"), "w") { |f| f.write("Hello") }
+
+          `git add .`
+          `git commit -m "Add files"`
+          `git checkout -b rename_files --quiet`
+
+          File.delete(File.join(dir, "first", "a"))
+          File.delete(File.join(dir, "second", "b"))
+          File.delete(File.join(dir, "c"))
+
+          File.open(File.join(dir, "a"), "w") { |f| f.write("hi") }
+          File.open(File.join(dir, "first", "b"), "w") { |f| f.write("bye") }
+          File.open(File.join(dir, "second", "c"), "w") { |f| f.write("Hello") }
+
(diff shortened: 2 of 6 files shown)


sleekybadger:
Hm, everything green locally 😞

orta:
This is where it can be tricky - different versions of git may give different results :-/

daniel-beard:
@sleekybadger @orta here's an issue on the ruby-git repo about this: https://github.com/schacon/ruby-git/issues/328

daniel-beard:
We can get the full diff info from ruby-git currently, maybe it's worth parsing that to get the filenames from there?

sleekybadger:
@orta hm, that can be reason. I'll rewrite this to raw diff without `--diff-filter=R`. 

@daniel-beard yep, unfortunately ruby-git not really well maintained, so I don't feel that issue will solved soon.

lib/danger/scm_source/git_repo.rb
```diff
@@ -19,6 +19,26 @@ def diff_for_folder(folder, from: "master", to: "HEAD")
       self.log = repo.log.between(from, to)
     end
 
+    def renamed_files
+      before_name_regexp = /rename from (.*)rename/
+      after_name_regexp = /rename to (.*)$/
```
daniel-beard: These regexes aren't correct or restrictive enough (you'll end up matching file contents as well).
Should use a single match with multiple groups like this: `/^diff --git a\/(.*?) b\/(.*?)/`

lib/danger/scm_source/git_repo.rb
```diff
@@ -19,6 +19,26 @@ def diff_for_folder(folder, from: "master", to: "HEAD")
...
+
+        before_match = patch.match(before_name_regexp)
+        next unless before_match
+
+        after_match = patch.match(after_name_regexp)
+        next unless after_match
+
+        {
+          before: before_match.captures.first,
+          after: after_match.captures.first
```
daniel-beard: Doesn't this need to be done per line, rather than just one capture?

spec/lib/danger/scm_source/git_repo_spec.rb
```diff
@@ -194,5 +195,32 @@
         end
       end
     end
+
+    describe "#renamed_files" do
+      it "returns array of hashes with names before and after" do
```
daniel-beard: Should also add a test with multiple renamed files.

lib/danger/scm_source/git_repo.rb
sleekybadger: No, with `--diff-filter=R`, diff will not contain file contents diff and will formated like this

```
diff --git a/hello1 b/hello1_renamed
similarity index 100%
rename from hello1
rename to hello1_renamed

diff --git a/hello3 b/hello3_renamed
similarity index 100%
rename from hello3
rename to hello3_renamed

diff --git a/hello5 b/hello5/some_folder/hello5
similarity index 100%
rename from hello5
rename to hello5/some_folder/hello5
```

spec/lib/danger/scm_source/git_repo_spec.rb
sleekybadger: @daniel-beard thanks, I will

orta:
🍏

orta:
don't have time to review tonight, but can look tomorrow ( someone else is welcome to take a stab instead of me though )

lib/danger/scm_source/git_repo.rb
```diff
@@ -19,6 +19,34 @@ def diff_for_folder(folder, from: "master", to: "HEAD")
       self.log = repo.log.between(from, to)
     end
 
+    def renamed_files
+      # Get raw diff with --find-renames --diff-filter
+      diff = exec(
+        "diff #{self.diff.from} #{self.diff.to} --find-renames --diff-filter=R"
```
daniel-beard: This adds another dependency for this library (one that may not be on all machines).
I'd *really* like to see this implemented using `ruby-git`s `diff.patch` functionality instead, as that ensures that there aren't any other dependencies required.

sleekybadger: No, it's [default](https://git-scm.com/docs/git-diff) git flags. 

Unfortunately we can't use it cause of `--find-renames` flag. 
We need to pass it to diff query cause older versions of git don't have this option as default (That caused errors on travis ci which has git version 1.8.0).

daniel-beard: OK, makes sense. That `exec` made me think this was shelling out :)

nikolaykasyanov:
I wonder if moved files will still be present in `modified_files` or not.

Issue #703, opened 2017-01-22
Danger should keep violations order


marcelofabri:
# Report

## What did you do?

Run `danger`.

## What did you expect to happen?

The comment should report violations in the same order as they were reported. This could be related to previous runs of Danger.

## What happened instead?

It didn't respect the order. See [the report on CI](https://circleci.com/gh/realm/SwiftLint/276) vs the [comment on the PR](https://github.com/realm/SwiftLint/pull/1226#issuecomment-274313662).

## Your Environment

* Which CI are you running on? Circle
* Are you running the latest version of Danger? Yep, 4.0.5
* What is your Dangerfile? https://github.com/marcelofabri/SwiftLint/blob/35209910b74c94731cfbb20376876ef6b8d4002d/Dangerfile

orta:
Maybe we need a timestamp on the creation of a violation?

marcelofabri:
I think this is a simpler issue than that - see that in the Circle log the violations are shown in the expected order. Something between that and reporting have changed it (maybe https://github.com/danger/danger/blob/master/lib/danger/request_sources/github.rb#L136-L139?)

marcelofabri:
I've opened https://github.com/realm/SwiftLint/pull/1230 which did a fresh run and it showed the same problem, so it's not an issue edit problem.

sleekybadger:
@marcelofabri @orta is this still actual?

I created Dangerfile that simply prints your messages:

```
message('Linting Alamofire with this PR took 3.33s vs 3.32s on master (0% slower)')
message('Linting swift with this PR took 11.55s vs 11.55s on master (0% slower)')
message('Linting Aerial with this PR took 0.43s vs 0.43s on master (0% slower)')
message('Linting SourceKitten with this PR took 1.55s vs 1.54s on master (0% slower)')
message('Linting Sourcery with this PR took 2.27s vs 2.31s on master (1% faster)')
message('Linting ios-oss with this PR took 17.73s vs 17.47s on master (1% slower)')
message('Linting Moya with this PR took 0.47s vs 0.44s on master (6% slower)')
message('Linting firefox-ios with this PR took 16.59s vs 16.57s on master (0% slower)')
message('Linting Nimble with this PR took 1.77s vs 1.8s on master (1% faster)')
message('Linting Quick with this PR took 0.53s vs 0.54s on master (1% faster)')
message('Linting realm-cocoa with this PR took 3.66s vs 3.65s on master (0% slower)')
[...]

PR #813, opened 2017-05-10
Fix broken violations order for github

sleekybadger:
## Summary

This should fix https://github.com/danger/danger/issues/703. I rewrite `update_pull_request!`, `submit_inline_comments!`, `submit_inline_comments_for_kind!` to not mutate violations and to apply sort only for inline comments.

### About sort proc:
Root of evil was in sort proc:
```
cmp = proc do |a, b|
  next -1 unless a.file
  next 1 unless b.file

  next a.line <=> b.line if a.file == b.file
  next a.file <=> b.file
end
```
Ruby uses [Quicksort](https://en.wikipedia.org/wiki/Quicksort) algorithm, which is pretty fast but can be unstable with not correct usage as ruby doc [says](https://ruby-doc.org/core-2.2.0/Array.html#method-i-sort). For example:
```
:001 > [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].sort { -1 }
 => [10, 2, 3, 4, 5, 1, 7, 8, 9, 6]
```
So when sort proc receives array where all comments not inline, order blew to hell. 

### About inline comments detection:
I think we can simply remove:
```
next -1 unless a.file
next 1 unless b.file
```
from sort proc, cause in `submit_inline_comments_for_kind!` we are skipping such files `next false unless m.file && m.line`. Also [here](https://github.com/sleekybadger/danger/blob/2d0ff0905a59650a17c88d202d413236d0400fc6/lib/danger/danger_core/messages/violation.rb#L23-L25) we have quite opposite logic, so I wonder can we change this condition to `file && line`?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -71,7 +71,7 @@ Style/AndOr:
 
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 320
+  Max: 350
 
 Metrics/CyclomaticComplexity:
   Max: 17
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Add your own contributions to the next release on the line below this, please include your name too. Please don't set a new version if you are the first to make the section for `master`.
 
+* Fix broken violations order for github - [@sleekybadger](https://github.com/sleekybadger)
+
 ## 5.2.1
 
 * Add ability to get list of renamed files - [@sleekybadger](https://github.com/sleekybadger)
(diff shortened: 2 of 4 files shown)


orta:
Awesome! Looks good (the fail is just a rubocop issue!)

sleekybadger:
Oh required named params missing in ruby 2.0.0. I'll fix this. Please, can you check second note?

JuanitoFatas:
```
Offenses:
lib/danger/request_sources/github/github.rb:13:5: C: Class has too many lines. [348/320]
    class GitHub < RequestSource ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
81 files inspected, 1 offense detected
```

we can increase [the rule](https://github.com/danger/danger/blob/66acdaf3a1f55eaa668848f1be0b820889521cdc/.rubocop.yml#L73) to 348 for now 😝

orta:
Amazing 👍

Issue #831, opened 2017-05-30
Add ability to set dismiss_out_of_range_messages per each message kinds


giginet:
# Feature request

If we set `dismiss_out_of_range_messages`, all kind of inline comments are dismissed.

However, I want to dismiss specific kind message only.
(For example, warnings out of ranges should be dismissed. errors should be commented.)

I'd like to set `dismiss_out_of_range_messages` per each kinds.

```ruby
dismiss_out_of_range_messages = {
  warning: true,
  error: false,
  message: true,
  markdown: true,
}
```

Could you tell me your opinion?
I think `dismiss_out_of_range_messages` should accept boolean or `[Symbol: boolean]`.

If it is OK, I'll send PR.

sleekybadger:
@giginet hi, can we close this?

orta:
Yes, this got shipped - unsure if it's in production yet though

PR #832, opened 2017-05-31
Added feature to set dismiss_out_of_range_messages per kind

giginet:
issue : [Add ability to set dismiss\_out\_of\_range\_messages per each message kinds · Issue \#831 · danger/danger](https://github.com/danger/danger/issues/831)

Added the feature to set dismiss_out_of_range_messages per kind.

```ruby
dismiss_out_of_range_messages({
  error: false,
  warning: true,
  message: true,
  markdown: true
})
```

If it is set as above, all errors are commented and warnings out of ranges are dismissed.

Also, you can set boolean. it works same with prior.

```ruby
dismiss_out_of_range_messages(true)
```


orta:
Cool - yeah, this looks good 👍

giginet:
I found the issue about this feature.
I'll fix and add test in another PR.
Please do not release until that. 🙇

PR #833, opened 2017-06-01
#trivial Fix the issue dismiss_out_of_range_messages

giginet:
Fix the problem of #832

#832 doesn't work fine. So I fixed in this PR. 🙇 

You can set like following on `Dangerfile`.
I tried this feature, it works fine for me.

```ruby
dismiss_out_of_range_messages({
  error: false,
  warning: true,
  message: true,
  markdown: true
})
```

I added tests for `Dangerfile`.

### DIFF
--- a/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
@@ -231,11 +231,11 @@ module Danger
     #           Ignore out of range inline messages, defaults to `true`
     #
     # @return   [void]
-    def dismiss_out_of_range_messages(dismiss: true)
+    def dismiss_out_of_range_messages(dismiss = true)
       if dismiss.kind_of?(Hash)
         @github.dismiss_out_of_range_messages = dismiss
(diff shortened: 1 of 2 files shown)


lib/danger/danger_core/plugins/dangerfile_github_plugin.rb
```diff
@@ -231,11 +231,13 @@ def html_link(paths, full_path: true)
     #           Ignore out of range inline messages, defaults to `true`
     #
     # @return   [void]
-    def dismiss_out_of_range_messages(dismiss: true)
+    def dismiss_out_of_range_messages(dismiss = true)
       if dismiss.kind_of?(Hash)
         @github.dismiss_out_of_range_messages = dismiss
       elsif dismiss.kind_of?(TrueClass)
         @github.dismiss_out_of_range_messages = true
+      elsif dismiss.kind_of?(FalseClass)
```
giginet: You can disable `dismiss_out_of_range_messages`.

```ruby
dismiss_out_of_range_messages(false)
```

orta:
Sorry - was a bit AFK - this looks 👍

Issue #835, opened 2017-06-09
Danger's inline comment duplicated in github


litmon:
# Report

https://github.com/litmon/danger-sample/pull/1

seems like above sample PR.
When the danger's inline comment have a reply comment, push some commit(move head_ref in this PR), and run `bundle exec danger`, danger commented duplicated line and comment as new conversation.
After this cause, every time I run `bundle exec danger`, danger's commented as new conversation.

<img width="541" alt="2017-06-10 2 50 06" src="https://user-images.githubusercontent.com/6662768/26987769-c56d6aae-4d87-11e7-8000-399aba5f1956.png">

<img width="482" alt="2017-06-10 2 53 14" src="https://user-images.githubusercontent.com/6662768/26987832-fca4e22c-4d87-11e7-9c82-dd3bfc801862.png">

This is dirty, so I want to fix this bug.

## What did you do?

- run the `bundle exec danger`, and you replied some danger's comments
- push some commit, and run `bundle exec danger`

## What did you expect to happen?

Do not duplicated comment.

## What happened instead?

Do duplicated comment as new conversation.

## Your Environment

* Which CI are you running on?

jenkins

* Are you running the latest version of Danger?

latest

* What is your Dangerfile?

```
warn("It is typo", file: "sample.txt", line: 1)
```

orta:
Is this duplicated in the case of it being shown twice on 1 danger run, or it being duplicated per danger-run?

litmon:
Yes, run danger after some commits pushed, duplicated comments happen every time.
it is happend when head_ref changed.

litmon:
I tried a patch like below, the bug fixed.


```
--- a/lib/danger/request_sources/github/github.rb
+++ b/lib/danger/request_sources/github/github.rb
@@ -312,13 +312,14 @@ module Danger
           end
 
           matching_comments = danger_comments.select do |comment_data|
-            if comment_data["path"] == m.file && comment_data["commit_id"] == head_ref && comment_data["position"] == position
+            if comment_data["path"] == m.file && comment_data["position"] == position
               # Parse it to avoid problems with strikethrough
               violation = violations_from_table(comment_data["body"]).first
               if violation
                 messages_are_equivalent(violation, m)
               else
-                comment_data["body"] == body
+                blob_regexp = %r{blob/[0-9a-z]+/}
+                comment_data["body"].sub(blob_regexp, "") == body.sub(blob_regexp, "")
               end
             else
               false
```

orta:
[...]

PR #837, opened 2017-06-09
Look at fixing accidental dupes - re: #835

orta:
This isn't _super well_ tested code: the detection of existing comments for deletion or updating. So I wanted to get a sense of CI etc.

If it's all green I'm OK with giving it a shot. It reads to me to be 👍


orta:
Fixes #837

PR #844, opened 2017-06-23
update plugin list to gitlab

leonhartX:
Since the danger.systems is moved to gitlab, the `danger plugins list` should search list from gitlab instead of github.


orta:
Ah yeah - great thinking, must have missed that

PR #845, opened 2017-06-23
add bitbucket_cloud to xcode_server supported_request_sources

anreitersimon:
Noticed that Bitbucket-Cloud is not available on XcodeServer.

I saw no reason why this wouldnt work now.
Please let me know if i missed something.


orta:
Thanks!

PR #853, opened 2017-07-03
Use reviews API from octokit;

Antondomashnev:
Danger's pull request reviews support has been done while the API was in beta testing. Now it's released and [octokit](https://github.com/octokit/octokit.rb) is updated with that API. No need for custom methods anymore 😄


orta:
I think we'll need to bump the gem spec's version of Octokit too 👍 

Otherwise looks great

Antondomashnev:
> Please include a CHANGELOG entry. 

Hm, is that a bug or I'm missing something?

orta:
Interesting, it could be - yeah

orta:
Probably just CI on an older commit, it's gone now

Antondomashnev:
> Probably just CI on an older commit, it's gone now

Yeah seems like not a Danger related 👍

orta:
Think you can remove this rubocop error: `lib/danger/commands/init.rb:185:21: C: Prefer using YAML.safe_load over YAML.load.` ?

Antondomashnev:
@orta sure thing

Antondomashnev:
@orta hah, it's not that easy, because the method is not available for ruby 2.0 😄

orta:
Ah sorry, yeah, I meant to just add an ignore for it :D

I just merged a PR that did that - want to revert back to 0cb6141 and then merge master, then you are good to go and I'll make a release

Antondomashnev:
@orta done 😉

orta:
Cool - shipped!

Issue #774, opened 2017-04-05
Utilizing --verbose flag from bundle


Tristan-zimmerman:
# Report

## Quick Note:
This less of a bug report and more of a feature request for more output if the `--verbose` flag is passed to bundler. Danger was working on some of our PRs, but not others, so I assume it might be something in our own configuration. I was hoping to be able to see what the out put to Github would have been so I could maybe identify the `UnprocessableEntity` it was complaining about.

Either way, thanks! Danger has been a very helpful tool. 👍 

## What did you do?
Ran `bundle exec danger` / `bundle exec danger --verbose` from Travis.

## What did you expect to happen?

ℹ Please replace this with what you expected to happen.
Expected it to run danger and report back to the Github PR.

## What happened instead?

ℹ Please replace this with of what happened instead.
Received an `UnprocessableEntity` error from Octokit:

```shell
2.56s$ bundle exec danger --verbose
/home/travis/.rvm/gems/ruby-2.3.1/gems/octokit-4.6.2/lib/octokit/response/raise_error.rb:16:in `on_complete': POST https://api.github.com/repos/localmotors/forth/pulls/340/comments: 422 - Validation Failed (Octokit::UnprocessableEntity)
Error summary:
  resource: PullRequestReviewComment
  code: invalid
  field: position // See: https://developer.github.com/v3/pulls/comments/#create-a-comment
	from /home/travis/.rvm/gems/ruby-2.3.1/gems/faraday-0.12.0.1/lib/faraday/response.rb:9:in `block in call'
	from /home/travis/.rvm/gems/ruby-2.3.1/gems/faraday-0.12.0.1/lib/faraday/response.rb:61:in `on_complete'
	from /home/travis/.rvm/gems/ruby-2.3.1/gems/faraday-0.12.0.1/lib/faraday/response.rb:8:in `call'
	from /home/travis/.rvm/gems/ruby-2.3.1/gems/octokit-4.6.2/lib/octokit/middleware/follow_redirects.rb:73:in `perform_with_redirection'
	from /home/travis/.rvm/gems/ruby-2.3.1/gems/octokit-4.6.2/lib/octokit/middleware/follow_redirects.rb:61:in `call'
	from /home/travis/.rvm/gems/ruby-2.3.1/gems/faraday-0.12.0.1/lib/faraday/rack_builder.rb:139:in `build_response'
[...]

PR #854, opened 2017-07-04
Show more detail for UnprocessableEntity

litmon:
Danger crashed and stopped CI Job because of Octokit::UnprocessableEntity when use github inline-comments.
This is maybe Danger's bug but the causes could not found because its error has no information.
So, it makes to show more detail for UnprocessableEntity, and not fail CI job.

See detail: https://github.com/danger/danger/issues/774

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -148,3 +148,6 @@ PercentLiteralDelimiters:
     "%w": ()
     "%W": ()
     "%x": ()
+
+Security/YAMLLoad:
+  Enabled: false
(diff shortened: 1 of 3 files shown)


litmon:
@orta `bundle exec rubocop lib` has failed due to below. Should I fix this?

```
Offenses:

lib/danger/commands/init.rb:185:21: C: Prefer using YAML.safe_load over YAML.load.
      config = YAML.load(File.read(".travis.yml"))
                    ^^^^
lib/danger/commands/init.rb:198:21: C: Prefer using YAML.safe_load over YAML.load.
      config = YAML.load(File.read("circle.yml"))
                    ^^^^
```

orta:
Can you disable those checks please? (`YAML.load` is on Ruby 2.0, which we need to support)

orta:
Awesome - thanks! Will see if I can get #853 in and ship these two

orta:
Thanks - shipped!

PR #862, opened 2017-07-19
#trivial Update buddybuild doc to include link to tutorial

Palleas:
#trivial Update buddybuild doc to include link to tutorial


orta:
Interesting, fails are definitely unrelated.

Palleas:
Thanks for the quick merge!

PR #864, opened 2017-07-21
#trivial Formatting fix for Bitrise setup

viktorbenei:
Fix for broken code sample at http://danger.systems/guides/getting_started.html for Bitrise

Should I also indent the block with spaces? https://github.com/danger/danger/blob/master/lib/danger/ci_source/travis.rb#L13 is indented, but probably that's not required as e.g. https://github.com/danger/danger/blob/master/lib/danger/ci_source/teamcity.rb#L27 is not indented inside the tripple-backtick block.

What's the preferred format?

### DIFF
--- a/lib/danger/ci_source/bitrise.rb
+++ b/lib/danger/ci_source/bitrise.rb
@@ -6,13 +6,15 @@ module Danger
   # ### CI Setup
   #
   # Add a script step to your workflow:
-  # ``` yml
+  #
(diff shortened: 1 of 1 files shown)


viktorbenei:
@DangerCI would love to, thanks ;)

viktorbenei:
Tests were not updated because no code was changed

orta:
👍

PR #867, opened 2017-07-26
[WIP] Don't explicitly look for 'github' in URL of GitHub repository environment variable (DANGER_GITHUB_API_BASE_URL)

showpony:
NOT TESTED

### DIFF
--- a/lib/danger/danger_core/environment_manager.rb
+++ b/lib/danger/danger_core/environment_manager.rb
@@ -83,12 +83,14 @@ module Danger
     private
 
     def get_repo_source()
-      if defined? DANGER_GITHUB_API_BASE_URL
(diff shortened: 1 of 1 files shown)


lib/danger/danger_core/environment_manager.rb
```diff
@@ -82,18 +82,18 @@ def raise_error_for_no_request_source(env, ui)
 
     private
 
-    def get_repo_source(repo_url)
-      if repo_url =~ /github/i
+    def get_repo_source()
+      if defined? DANGER_GITHUB_API_BASE_URL
```
orta: This is optional I'm afraid, maybe these could be a second check inside the `else` at the end?

PR #883, opened 2017-08-28
Fix implementation of bsearch in GitlabCI#determine_merge_request_id

dvlprliu:
This is a very simple PR, just change the `==` operator to `>=` according to [ruby doc](https://ruby-doc.org/core-2.2.0/Array.html#method-i-bsearch)

### DIFF
--- a/lib/danger/ci_source/gitlab_ci.rb
+++ b/lib/danger/ci_source/gitlab_ci.rb
@@ -39,7 +39,7 @@ module Danger
 
       merge_requests = client.merge_requests(project_id, state: :opened)
       merge_request = merge_requests.auto_paginate.bsearch do |mr|
-        mr.sha == base_commit
+        mr.sha >= base_commit
       end
 
       merge_request.nil? ? 0 : merge_request.iid


dvlprliu:
@DangerCI Of course, would love to. 😃

orta:
👍 Alright - I've sent the invite, and this looks good to me

nikolaykasyanov:
I'm not sure this PR is the reason but Danger posts comments to the wrong Merge Request.

nikolaykasyanov:
I'm not even sure whether it's safe to use bsearch on this array. It's definitely sorted but not by `mr.sha`. Am I missing something?

sleekybadger:
@nikolaykasyanov @orta I'm pretty sure that this is a reason, cause `sha` and `base_commit` are string and in ruby:

```
'dcb' > 'abc'
 => true
'abc' >= 'dcb'
 => false
```
We are not sorting array here, this block returns first merge request, which has same sha as base commit

orta:
good debugging 👍

nikolaykasyanov:
What about replacing bsearch with stupid linear search? I can make a PR.

orta:
Seems fine to me, can't imagine it's a massive list to go through 👍

sleekybadger:
Not sure that this is necessary, cause commit sha is unique, so we can simply revert this pr

nikolaykasyanov:
@sleekybadger I'm still confused and probably missing something. Doesn't using `bsearch` on an array assumes that it's sorted by the key we use in bsearch's block?

sleekybadger:
@nikolaykasyanov yeah, you're right. We can use `Array.index` which is leaner.

```
merge_request = merge_requests.auto_paginate.index do |mr|
  mr.sha == base_commit
end
  
merge_request.nil? ? 0 : merge_request.iid
```

nikolaykasyanov:
@sleekybadger @orta thanks for discussion & making it quick!

PR #887, opened 2017-09-05
Stricter generated_by_danger? check

marcelofabri:
This was causing issues when one danger_id is a substring of another one.


orta:
Fails are legit

marcelofabri:
Will take a look later

PR #889, opened 2017-09-07
Add vsts plugin

petester42:
Adds all the helpful Dangerfile properties you know and love from other sources to VSTS. You can now specify `vsts` in your Danger file like so `warn "PR is classed as Work in Progress" if vsts.pr_title.include? "[WIP]"` 

VSTS does not handle adding html to comments so for now you can only use the `markdown_link` property to add links to files.


orta:
Great - yeah, this looks 👍 to me - good to go on 🍏

PR #880, opened 2017-08-23
Add `verify_ssl` argument to skip ssl verification

nikhilsh:
- Add tests for verify ssl
- Add no_proxy_fix gem to fix issue in Ruby 2.4.0/2.4.1

As discussed in https://github.com/danger/danger/issues/830

Done with the help of @winston

PR #890, opened 2017-09-07
Add support for toggling SSL Verification for Octokit

nikhilsh:
This is similar to #880, in that it allows toggling of Octokit's SSL Verification. In #880, only the `pr` command had an option to toggle SSL verification (e.g. `bundle exec danger pr --verify-ssl`). 

While that PR added the verification toggling for PR, danger (`bundle exec danger`) itself didn't have the option to toggle SSL Verification. After looking through `runner.rb` and how it passes arguments down to Octokit, I surmised that a clean way to allow toggling would be to simply read an environment variable (`DANGER_OCTOKIT_VERIFY_SSL`) that would be used by `Octokit` when starting up the Octokit client. 

This would prevent introducing a new argument in command line and having to pass that down several layers to get Octokit to use it from within `github.rb`. 

As should be expected, not having `DANGER_OCTOKIT_VERIFY_SSL` would leave SSL verification, as is the default behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 * Add your own contributions to the next release on the line below this, please include your name too. Please don't set a new version if you are the first to make the section for `master`.
 
+## 5.4.1
+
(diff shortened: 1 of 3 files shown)


orta:
Yeah, this looks 👍

PR #892, opened 2017-09-12
Add support for Codeship CI

ghiculescu:
Added [codeship](https://codeship.com/) as a new CI source. I mostly copied the style of the Semaphore CI source, and used the same logic that [danger.js uses](https://github.com/danger/danger-js/blob/master/source/ci_source/providers/Codeship.ts#L40) in getting a PR from a branch name.

### DIFF
--- a/lib/danger/ci_source/codeship.rb
+++ b/lib/danger/ci_source/codeship.rb
@@ -19,12 +19,16 @@ module Danger
     def self.validates_as_pr?(env)
       return false unless env["CI_BRANCH"] && !env["CI_BRANCH"].empty?
 
-      # this is fairly hacky, see https://github.com/danger/danger/pull/892#issuecomment-329030616 for why
-      !Danger::RequestSources::GitHub.new(nil, env).get_pr_from_branch(env["CI_REPO_NAME"], env["CI_BRANCH"], self.class.owner_for_github(env))
+      !pr_from_env(env, nil).nil?
     end
 
     def self.owner_for_github(env)
       env["CI_REPO_NAME"].split("/").first
+    end
+
+    # this is fairly hacky, see https://github.com/danger/danger/pull/892#issuecomment-329030616 for why
+    def self.pr_from_env(env, ci_source)
+      Danger::RequestSources::GitHub.new(nil, env).get_pr_from_branch(env["CI_REPO_NAME"], env["CI_BRANCH"], owner_for_github(env))
     end
 
     def supported_request_sources
@@ -33,7 +37,7 @@ module Danger
 
     def initialize(env)
       self.repo_slug = env["CI_REPO_NAME"]
-      self.pull_request_id = Danger::RequestSources::GitHub.new(self, env).get_pr_from_branch(env["CI_REPO_NAME"], env["CI_BRANCH"], self.class.owner_for_github(env))
+      self.pull_request_id = self.class.pr_from_env(env, self)
       self.repo_url = GitRepo.new.origins
     end
   end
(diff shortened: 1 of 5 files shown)


lib/danger/ci_source/semaphore.rb
```diff
@@ -4,7 +4,7 @@
 module Danger
   # ### CI Setup
   #
-  # For Semaphor you will want to go to the settings page of the project. Inside "Build Settings"
+  # For Semaphore you will want to go to the settings page of the project. Inside "Build Settings"
```
JuanitoFatas: 👍

ghiculescu:
Oops, should be good now @JuanitoFatas

ghiculescu:
Hey @JuanitoFatas what do you think about https://github.com/danger/danger/pull/892/commits/056f8cc2af26bf0ae4c87ca5394e4be9fd3a2598 ?

The problem here is that codeship doesn't expose a PR number in ENV the way most other CIs seem to. The workaround is to look up Github and check for a PR whose `head` matches the branch we are merging into ([code](https://github.com/ghiculescu/danger/blob/feature/codeship/lib/danger/request_sources/github/github.rb#L36)). However, it is unlikely that there would ever be a PR where master is the head (unless you're using a weird git structure).

Are you happy with these sorts of assumptions in Danger? The other idea I had was to do something like this: `!Danger::RequestSources::GitHub.new(nil, env).get_pr_from_branch(..., ..., ...).nil?` but it seems fairly grimey, and means there will be an extra github API called made for every build.

deecewan:
for reference @ghiculescu, the extra API call [is what we do on `danger-js`](https://github.com/danger/danger-js/blob/master/source/ci_source/providers/Codeship.ts)

I don't necessarily think you can assume that master is : a) the 'master' branch, or b) never PRed

ghiculescu:
Updated: https://github.com/ghiculescu/danger/compare/41e8b500673ca86dce6c3df9302459854cfe272c...8dd90faa7b47a9d1bd5524cf6d6b3ffc36a4af4a

lib/danger/ci_source/codeship.rb
```diff
@@ -0,0 +1,44 @@
...
+      !pr_from_env(env, nil).nil?
+    end
+
+    def self.owner_for_github(env)
+      env["CI_REPO_NAME"].split("/").first
+    end
+
+    # this is fairly hacky, see https://github.com/danger/danger/pull/892#issuecomment-329030616 for why
+    def self.pr_from_env(env, ci_source)
+      Danger::RequestSources::GitHub.new(nil, env).get_pr_from_branch(env["CI_REPO_NAME"], env["CI_BRANCH"], owner_for_github(env))
```
JuanitoFatas: do we need to pass down the `ci_source` to `Danger::RequestSources::GitHub`? 🤔 
do we need `ci_source`? 😂

ghiculescu: `ci_source` gets used by a bunch of stuff eg [here](https://github.com/danger/danger/blob/master/lib/danger/request_sources/github/github.rb#L74) and [here](https://github.com/danger/danger/blob/master/lib/danger/request_sources/github/github.rb#L80) - i wasn't keen to take on a full refactor.

ghiculescu: it's just not required in my particular case - i just need a github client. if you prefer, i can try and find a neater way of writing that.

ghiculescu: wait, i misread your comment... fixed in https://github.com/danger/danger/pull/892/commits/a49fc306196afe79f69b77fd03eb4f3052cf040d

JuanitoFatas:
Yay! 🎉  Thank you for the contributions! It's so awesome!!! <3 <3 <3

> @ghiculescu is not a contributor yet, would you like to join the Danger org?

Would you like to join? 🙏 🙇

ghiculescu:
Yeah okay sign me up!

Can you please let me know when the next gem release goes out so I can point our gemfile to it? :D

orta:
Think it already went out 👍

orta:
And the invite has been sent <3 thanks

PR #894, opened 2017-09-12
Fix Violation#to_s

edvakf:
It was showing file and line when they are empty and not showing when they are set.

Originally `spec/lib/danger/danger_core/plugins/messages_spec.rb` which I've added was failing like this.

1____prog_danger___zsh_


edvakf:
Thank you.

edvakf:
Is this going to be released as a new version any time soon? I need it for a PR to danger-rubocop.

orta:
https://github.com/danger/danger/pull/892 was looking close, so I was planning on waiting - but I shipped a version anyway

PR #897, opened 2017-09-15
Fetch both PR head and base branch

daveallie:
# Problem

If you have a PR where the head is `my-branch` and the base is `master`, but master has been committed to since you opened your PR, then your git tree will look something like this:

```
--*-------*  <- master
  \
    ---*  <- my-branch
```

Annoyingly, Codeship runs the following command to checkout the codebase:

```
git clone --branch "my-branch" --depth 50 git@github.com:danger/danger.git ~/src/github.com/danger/danger
```

Which means, all that's stored locally is

```
--*
  \
    ---*  <- my-branch
```

This causes issues as the current `origin/master` ref can't be found, and a `git fetch --depth 1000000` will only fetch from currently known refs, which doesn't reveal our `origin/master` commit.

# Solution

When looking commit hashes, or the merge base of multiple commits, first attempt to resolve them. If that fails, progressively fetch deeper and deeper into the specific branches before resorting to the original `git fetch --depth 1000000`. Fetching specific branches means that the in the example above the `master` ref gets resolved and Danger no longer crashed on Codeship.

I have applied the same changes to all git systems supported by Danger, but based my changes off the public API, and would appreciate some validation to ensure that my changes won't break danger for users on these VCS platforms.

This logic is based heavily on the Gitlab CE code: https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/gitlab/ee_compat_check.rb#L184


daveallie:
Hmmm, those rubocop violations seem to all be from the `0.50` upgrade? Did you guys recently do that update?

orta:
Oh dear, sorry about the super late review - for some reason the rubocop version isn't being locked by bundler when it's being called. I'll see if I can make a PR as it blocks all of the open ones

orta:
Alright, that's rubocop handled. 

So yeah, this looks good to me, this must have been quite a bit of thinking - so thanks for that. It's surprising how intricate the ability to just get a reasonable diff is 👍 

As this was non-trivial, I've invited you to the org, as a part of [Moya Community Continuity](https://github.com/Moya/contributors/issues/6) - you're welcome to participate in PRs or not, and help out at a level that works for you if you choose.

PR #902, opened 2017-09-26
Look at locking the rubocop version

orta:
I'm not really sure _why_ the `bundle exec` is  being ignored, so I want to first verify that it's using 0.50.x


orta:
Found it, the default ruby `.gitignore` ignores the Lockfile. So I've applied semver to rubocop.

```sh
# for a library or gem, you might want to ignore these files since the code is
# intended to run in multiple environments; otherwise, check them in:
# Gemfile.lock

```

PR #908, opened 2017-10-03
Fix empty commits list error on GitLab

makadaw:
#trivial GitLab API may return an empty list of commits and in this situation, danger fails with runtime error without explanation. This changes can process this kind of situations and raise danger error with a better explanation.


orta:
Yeah, this is great @makadaw 👍

orta:
Will release when #909 gets wrapped up too

PR #909, opened 2017-10-03
Fix VSTS comments

cyupa:
This fixes the VSTS commenting logic.

At this point, the VSTS module looks for the `new_comment` flag to decide if it should post a new comment or not - which is set to false.
It then tries to update the previous comment which doesn't exist for a new pull request.

The fix looks for any comment posted by danger to determine if a previous comment was posted and tries to update it.
If no comment added by danger is found, it adds a new comment.

### DIFF
--- a/lib/danger/request_sources/vsts.rb
+++ b/lib/danger/request_sources/vsts.rb
@@ -101,15 +101,13 @@ module Danger
           comment_id = comment[:id]
           comment_content = comment[:content].nil? ? "" : comment[:content]
           # Skip the comment if it wasn't posted by danger
-          next if !comment_content.include?("generated_by_#{danger_id}")
+          next unless comment_content.include?("generated_by_#{danger_id}")
           # Updated the danger posted comment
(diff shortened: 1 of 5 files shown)


cyupa:
@orta sorry for the messy way of doing this. I have limited experience with Ruby and it was mostly a trial and error process 😄

orta:
Nice one - I think you're gonna want a test for this ( go check the github/gitlab/bitbucket test files, they'll basically be something you can copy & edit slightly) as you don't want that breaking in the future.

Looks good to me though, I also don't mind about the commit faffing, I can just use the GitHub button to squash all commits

cyupa:
Cool, I'll look at how I can update the tests to cover this. I've tested the implementation against a VSTS project, hence the commit madness 💩

cyupa:
@orta This should do it. Thanks!

orta:
Congrats @cyupa - you did a great job there

cyupa:
Thanks again for this awesome tool @orta 👍

orta:
Shipped as 5.5.3

PR #910, opened 2017-10-04
Use linear search to get GitLab MR ID in GitLab CI

nikolaykasyanov:
`bsearch` can't be used on an array not sorted by `mr.sha`. See discussion: https://github.com/danger/danger/pull/883#issuecomment-334143581.


nikolaykasyanov:
I tested this on my CI and it does indeed work as expected.

nikolaykasyanov:
🎉

PR #914, opened 2017-10-23
#trivial Fix GitLab Setup — GitHub env variable to GitLab env variable

k15a:
I suppose this is the right environment variable. GitHub token doesn't make sense here for me.


orta:
Thanks! <3 doc updates

Issue #915, opened 2017-10-26
nil:NilClass for Reviewers on BitBucket Server


caffeineflo:
# Report

## What did you do?

`bundle exec danger --fail-on-errors=true`  on BuddyBuild that's connected to a BitBucket Server v5.4.0

## What did you expect to happen?

Running  Dangerfile without it crashing.

## What happened instead?

```
    Danger::DSLError: 
    [!] Invalid `Dangerfile` file: undefined method `length' for nil:NilClass
     #  from Dangerfile:22
     #  -------------------------------------------
     #  # Please assign someone!
     >  warn "This PR does not have any assignees yet." unless bitbucket_server.pr_json["reviewers"].length == 0
     #  
     #  -------------------------------------------
      Dangerfile:22:in `block in parse'
```

## Your Environment

* Which CI are you running on?
BuddyBuild
* Are you running the latest version of Danger?
Yes, started using Danger yesterday
* What is your Dangerfile?

  ```ruby
  # Please assign someone!
  warn "This PR does not have any assignees yet." unless bitbucket_server.pr_json["reviewers"].length == 0


  # Warn if Carthage got an update
  warn "#{bitbucket_server.html_link("Cartfile.resolved")} was edited. Make sure to update carthage 
  when reviewing this PR" if git.modified_files.include? "Cartfile.resolved"

  # Warn when there is a big PR
  warn("Big PR") if git.lines_of_code > 500

  swiftlint.lint_files
  ```

orta:
This is just a normal error in your code, it looks like `bitbucket_server.pr_json["reviewers"]` is null. Perhaps it's not set up correctly, or BB returns different data? Danger is doing the right thing here.

caffeineflo:
I'm able to parse and use the `pr_author` property from `bitbucket_server`, shouldn't .pr_json have some value then?

I simply took this part from your examples ...

orta:
You might want to try adding something like `message(bitbucket_server.pr_json)` to see what's in there then - maybe we need to update the docs and their API has change, I have any BB repos to keep up to date with

caffeineflo:
@orta I put a debug message in, trying to output the pr_json object. Unfortunately nothing came back, so I assume the whole object is missing. 

Do you happen to know if this is a BitBucket Server setting oder how I could further debug the absence of the JSON object?

orta:
[...]

PR #916, opened 2017-10-27
#trivial: Fixed "Ensure that all PRs have an assignee" example logical error

caffeineflo:
This closes the issue #915 that I opened yesterday.

Basically, the example shown in the documentation was off. Keys have to be clarified via colon and the logic on the warning was wrong:
You want to warn if there's no assignee to the PR, not if there is one assigned. 

Tested it internally fwiw.


orta:
<3 - thanks!

caffeineflo:
Thank you! Great project

PR #922, opened 2017-11-09
Handle Github repos with dots in them

orj:
The PR corrects the regex that was used to extract the Github repo slug to handle repositories with periods (dots) in them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 ## master
-
+ 
 > _Add your own contributions to the next release on a new line above this; please include your name too._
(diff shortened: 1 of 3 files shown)


JuanitoFatas:
Amazing ✨ ! What do you think about we also apply this to [other ci sources](https://github.com/danger/danger/tree/master/lib/danger/ci_source), maybe extract to a class to reuse it :D

orj:
@JuanitoFatas I looked at the other CI sources and it wasn't clear that it was relevant.  Other CI sources get the repo name/slug differently.

JuanitoFatas:
Thank you for contributions and this is released in v5.5.5!

Issue #931, opened 2017-12-18
Feature request: dry-run against all open PR's


rattrayalex-stripe:
I see `danger local` and `danger pr` in [the docs](http://danger.systems/guides/troubleshooting.html), which look great. 

We're evaluating danger and would like to be able to test new rules against historical PR's before committing them. It is okay if this is slow. 

We're considering writing a wrapper script, like `get-open-prs -n 100 | xargs bundle exec danger pr` but would be great if this was built in. 

It's also not clear from the docs whether `danger pr` will post a comment to the PR, or print out what it _would_ do to the console (I desire the latter behavior). 

There isn't documentation for the CLI on the website that I can find, but `danger --help` locally does not list a `--dry` option.

rattrayalex-stripe:
This seems related to https://github.com/danger/danger/issues/382 but I haven't read that thread thoroughly.

orta:
You could use `danger pr` and it wouldn't post a comment - that should work fine 👍

I'll ship a PR making the CLI more explicit about this

I'm not *actively* working on Danger Ruby (but I'm maintaining it), so the `local` deprecation isn't really going anywhere.

orta:
WRT a feature for trying your Dangerfile against all runs, I could see it working.

Maybe `danger trial https://github.com/artsy/emission --count 10` and it would run through the same as `bundle exec danger pr https://github.com/artsy/emission/pull/1`, `bundle exec danger pr https://github.com/artsy/emission/pull/2` etc?

rattrayalex-stripe:
Awesome, yeah `danger trial <repo_url> --count` sounds ✨ ✨  !

PR #932, opened 2017-12-18
Improve the CLI docs for local/pr

orta:
From the feedback in #931 

I think the CLI could give a better example of what is up with local, and that pr won't post to the PR.


rattrayalex-stripe:
❤️ 

This is terrific!

Issue #926, opened 2017-11-28
Getting Could not set up API... on GitLab CI Runner


mbogh:
# Report

## What did you do?

`bundle exec danger --verbose`

## What did you expect to happen?

Running my Dangerfile and failing the MR.

## What happened instead?

```
$ bundle exec danger --verbose

Could not set up API to Code Review site for Danger

For Danger to run on this project, you need to expose a set of following the ENV vars:
 - GitHub: DANGER_GITHUB_API_TOKEN
 - GitLab: DANGER_GITLAB_API_TOKEN
 - BitbucketServer: DANGER_BITBUCKETSERVER_USERNAME, DANGER_BITBUCKETSERVER_PASSWORD, DANGER_BITBUCKETSERVER_HOST
 - BitbucketCloud: DANGER_BITBUCKETCLOUD_USERNAME, DANGER_BITBUCKETCLOUD_PASSWORD
 - VSTS: DANGER_VSTS_API_TOKEN, DANGER_VSTS_HOST


[...]

PR #934, opened 2017-12-19
Raise error if port number is included in DANGER_GITLAB_HOST

mbogh:
Based on the experience and feedback in #926 

An error will be raised if you accidentally include the port number in `DANGER_GITLAB_HOST`.


orta:
Awesome, looks good 👍

PR #941, opened 2018-01-05
Use distinct on validations prior to publish

saantiaguilera:
Currently the `validations_reports` hash is a freezed map with messages/errors/etc. If we somehow have duplicated messages we end up having the exact same messages under a category.

In my case this occurs because I have a ton of test reports and the `danger-junit` gem posts the `'A test failed'` message for each. Since I cant operate over the `validation_reports` hash because its freezed, I ended up having N times the `'A test failed'` message

I will make later a PR to @orta in the `danger-junit` gem, but I thought this would be a desired logic for danger itself (whats the point of having twice the exact same message under a category?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 * Add Bitbucket Server support for Bitrise CI. [@copini](https://github.com/copini)
+* Use unique entries for the validation reports
 
 > _Add your own contributions to the next release on a new line above this; please include your name too._
 > _Please don't set a new version if you are the first to make the section for `master`._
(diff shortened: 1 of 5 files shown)


orta:
Agree 👍 

Any chance you can get a test for this?

CI is going to fail, master is red ATM - I know this

saantiaguilera:
Done :) I had to implement the `hash` and `eql?` method since `Array#uniq` uses them for comparison (its not bad having the DTO with those so I guess its still useful for hashing and performance equality). I can implement the unicity on myself if not.

As for the hash function, I used the standard one based on prime numbers and delegations..

I have no idea how to become a contributor. How should I?

orta:
It'll [happen automatically now](https://github.com/danger/peril-settings/blob/master/org/aeryn.ts#L17), I need to update Danger's Dangerfile 

The hash function looks 👍  to me

orta:
travis keeps failing, not our fault.

PR #942, opened 2018-01-16
Added support for Bitbucket Server to TeamCity CI

davevdveen:
Added support for Bitbucket Server to TeamCity CI. Because TeamCity doesn't setup all the required variables those need be added to the ENVIRONMENT variables.
Tested it on our TeamCity and Bitbucket Server setup and it is working if all the required variables are set. Also added support for older versions of Bitbucket Server API which provide latestChangeset instead of latestCommit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## master
 
 * Add Bitbucket Server support for Bitrise CI. [@copini](https://github.com/copini)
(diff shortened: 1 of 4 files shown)


neilkimmett:
Looks reasonable to me 👍 . If I were to nitpick I'd suggest maybe squashing those little one line commits into the relevant earlier commit?

orta:
I'm chill with some small commits  - I'll get this shipped

PR #949, opened 2018-01-26
Add --remove-previous-comments functionality

### LINKED ISSUE #948: Ability to make danger less spammy for big PRs
Currently, Danger has two options to report back to GitHub, either update the comment or create a new one. 
That would be really helpful if she has also a possibility to replace a comment but put the comment in the end of the PR rather than updating it somewhere on the page 'x'.

Motivation: 
- `--new-comment` is very nice but can become a bit spammy for big projects and PRs, as you may want to post several danger outputs for every PR
- updating the comment can be easily lost, as it can be somewhere in the very beginning of the PR

JoeSSS:
This PR adds `--remove-previous-comments` CLI flag. The flag can be used in order to get the new danger comments in the tail of the PR and remove previous comments, so the PR looks bot so spammy.

Closes #948

### DIFF
--- a/lib/danger/request_sources/github/github.rb
+++ b/lib/danger/request_sources/github/github.rb
@@ -179,7 +179,7 @@ module Danger
         main_violations_sum = main_violations.values.inject(:+)
 
         if (previous_violations.empty? && main_violations_sum.empty?) || remove_previous_comments
-          # Just remove the comment, if there's nothing to say of --remove-previous-comments CLI was set.
+          # Just remove the comment, if there's nothing to say or --remove-previous-comments CLI was set.
           delete_old_comments!(danger_id: danger_id)
         end
 
--- a/lib/danger/request_sources/github/github.rb
+++ b/lib/danger/request_sources/github/github.rb
@@ -140,12 +140,12 @@ module Danger
       end
 
       # Sending data to GitHub
-      def update_pull_request!(warnings: [], errors: [], messages: [], markdowns: [], danger_id: "danger", new_comment: false, comment_at_tail: false)
+      def update_pull_request!(warnings: [], errors: [], messages: [], markdowns: [], danger_id: "danger", new_comment: false, remove_previous_comments: false)
         comment_result = {}
         editable_comments = issue_comments.select { |comment| comment.generated_by_danger?(danger_id) }
         last_comment = editable_comments.last
         should_create_new_comment = new_comment || last_comment.nil?
-        should_remove_comments = comment_at_tail || last_comment.nil?
+        should_remove_comments = remove_previous_comments || last_comment.nil?
 
         previous_violations =
           if should_create_new_comment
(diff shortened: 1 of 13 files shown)


orta:
TBH, I wouldn't recommend doing this - this is gonna be real annoying in your emails. Danger editing the same comment means you don't get an email per CI run. If you think this is definitely what you want to do, then I'm fine with this in.

I'd recommend looking into editing the PR's body though

JoeSSS:
What I’m trying to solve is the problem with the spam in PRs, as we have several jobs responding to the PR via Danger after every commit. We tried to use standard flags but it didn’t work as the PRs can be quite long or maybe people are just too lazy to look through whole PR searching Danger output. So, everything works fine using ‘—new-comment’ but it is spammy. 
True, emails are annoying, but we deal with the environment where developers most likely disabled all the emails from git, as we are ~30 devs developing on a daily basis. Would be really nice to reduce the noise at least on the GitHub level.

orta:
cool cool, yeah, get it covered so that someone doesn't break it in the future and I'm 👍

lib/danger/request_sources/github/github.rb
```diff
@@ -178,8 +178,8 @@ def update_pull_request!(warnings: [], errors: [], messages: [], markdowns: [],
 
         main_violations_sum = main_violations.values.inject(:+)
 
-        if previous_violations.empty? && main_violations_sum.empty?
-          # Just remove the comment, if there's nothing to say.
+        if (previous_violations.empty? && main_violations_sum.empty?) || remove_previous_comments
+          # Just remove the comment, if there's nothing to say of --remove-previous-comments CLI was set.
```
orta: minor typo

lib/danger/request_sources/gitlab.rb
```diff
@@ -128,8 +128,8 @@ def update_pull_request!(warnings: [], errors: [], messages: [], markdowns: [],
           previous_violations = parse_comment(comment)
         end
 
-        if previous_violations.empty? && (warnings + errors + messages + markdowns).empty?
-          # Just remove the comment, if there"s nothing to say.
+        if (previous_violations.empty? && (warnings + errors + messages + markdowns).empty?) || remove_previous_comments
+          # Just remove the comment, if there's nothing to say of --remove-previous-comments CLI was set.
```
orta: same here

JoeSSS:
Thanks for fast reviews 👍

orta:
OK, that's shipped

znacloud:
--remove-previous-comments flag work incorrect on gitlab-ci, all comments will be removed and no comment displayed.
--new-comment flag still update previous comment instead of creating new one.
danger version is 5.6.7

orta:
@znacloud  you're welcome to take a look and fix this 👍

JoeSSS:
the problem must be here: https://github.com/JoeSSS/danger/blob/3bda96eb104a50f0f255ba6b1a66b07d788162f6/lib/danger/request_sources/gitlab.rb#L131

the logic should be decoupled from comment posting.

PR #957, opened 2018-02-09
Security Updates for rubocop, take 2

JuanitoFatas:
I still see warnings while visit danger/danger:

screen shot 2018-02-09 at 16 35 45

that's because we specify two rubocop in gemspec and Gemfile. Let's specify all dev dependencies in Gemfile.

This PR:

- move dev dependencies to Gemfile, sort alphabatically
- Let's try turn the plugins lint back on, disabled in #942
- Rubocop updates for 0.49.1

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -148,3 +148,6 @@ Security/YAMLLoad:
 
 Style/RaiseArgs:
   Enabled: false
+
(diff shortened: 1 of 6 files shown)


orta:
take two 🍶

orta:
I don't see any now on the home

PR #958, opened 2018-02-09
Enable BitBucket Cloud for Circle CI

andrewlord1990:
Circle CI supports BitBucket Cloud.

For Danger, it should be as simple as adding it as a supported request source on Circle CI, as it is already checking the environment variables.


orta:
Ah - nice 👍

orta:
Shipped as 5.5.10

Issue #966, opened 2018-03-09
Crash by `sh: 2: [commit hash]: not found`


tbrand:
# Report

## What did you do?

I'm hosting a danger as a CI platform.

## What did you expect to happen?

Every repositories will be executed successfully.

## What happened instead?

Some of repositories be crashed.

```
From https://[our enterpise github]/[org]/[repo]
 * [new branch]      master     -> origin/master
 * [new tag]         ... (omitted)
 * [new tag]         ...
 * [new tag]         ...
 * [new tag]         ...
sh: 2: 84bea298cb68c14109c2310dc01ffd9d31a28954: not found
sh: 3: 39f5e1aec479d55978061219fbc5c4dc6df9dd23^{commit}: not found
Danger::DSLError: [31m
[!] Invalid `Dangerfile.hosted` file: git '--git-dir=/tmp/548d6d34617e8ac269c8773a0edfe6a2/.git' '--work-tree=/tmp/548d6d34617e8ac269c8773a0edfe6a2' diff '--numstat' '5032e7c10fc1378e92954bf65d242519c4ed4f67
84bea298cb68c14109c2310dc01ffd9d31a28954
39f5e1aec479d55978061219fbc5c4dc6df9dd23' 'danger_head'  2>&1:fatal: ambiguous argument '5032e7c10fc1378e92954bf65d242519c4ed4f67
84bea298cb68c14109c2310dc01ffd9d31a28954
39f5e1aec479d55978061219fbc5c4dc6df9dd23': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'[0m
```

The error message says `sh: 2: 84bea298cb68c14109c2310dc01ffd9d31a28954: not found`.
So the **commit hash is recognized as a command**.

I did debugging by printing `obj1` and `obj2` at here. https://github.com/ruby-git/ruby-git/blob/master/lib/git/lib.rb#L354

The result is
```
---------- obj1 ----------
5032e7c10fc1378e92954bf65d242519c4ed4f67
84bea298cb68c14109c2310dc01ffd9d31a28954
39f5e1aec479d55978061219fbc5c4dc6df9dd23
--------------------------

---------- obj2 ----------
danger_head
--------------------------
```

So the problem is that obj1 is passed as a String of **multiple lines**.

So my question are
- Why is this happend?
- How to fix this situation? (Especially, what is the correct way to fetch the repositories.)

## Your Environment

[...]

Issue #965, opened 2018-03-06
Error with multiple merge bases


raymondhoagland:
# Report

## What did you do?
Run Danger on a PR from a Jenkins build.

## What did you expect to happen?
Danger runs without crashing.

## What happened instead?
In _some_ scenarios, Danger crashes with an error remarkably similar to the error in #904.

## Your Environment
* Running in Jenkins
* Running on v5.5.5
  ```ruby
  # please paste here, with 2-space identation, thank you!
  xcov.report(
     scheme: ENV["FL_XCOV_SCHEME"],
     workspace: ENV["FL_XCOV_WORKSPACE"],
     skip_slack: ENV["FL_XCOV_SKIP_SLACK"],
     derived_data_path: ENV["FL_XCOV_DERIVED_DATA"],
     only_project_targets: true
  )
  xcode_summary.ignored_files="Pods/**"
  xcode_summary.report(ENV["FL_XCS_REPORT"])
  ``` 
Not exactly the right Dangerfile but I don't think it's important.  If it is, I'll sanitize it and update.

## Food For Thought
My best guess for how this ended up happening is related to [this line](https://github.com/danger/danger/blob/0d70bc4560976f30ca30dd3e7d2582d944bc4a83/lib/danger/scm_source/git_repo.rb#L168), specifically the fact that the call is for `merge-base --all` which has the chance of providing multiple merge bases.  

In most cases there should be a single merge base returned because the incremental fetch starts out very small.  However if the incremental fetch size increases because a merge base was not found, it may end up becoming large enough that multiple merge bases are returned at once.  The result (multiple lines of text) isn't sanitized so the error ends up getting displayed.

This normally ends up sorting itself out as a merge base eventually appears closer to the tip of the branch and only one result is found again, making it harder to debug.  I think this happens to me more often because I am using shallow cloning - the merge base isn't found right away and a larger fetch is often required.

Is there a reason to include the `--all` flag?  If I'm right removing that would easily solve this and provide a much better explanation for #904 than not using a puts statement.

tbrand:
Hi @raymondhoagland

I've created the patch at https://github.com/danger/danger/pull/967
BTW, I've fixed it by chnaging a way to fetch the repositories like this
```
git init
git remote add origin <some origin>
git fetch origin <base_branch> --depth 1
git fetch origin +refs/pull/<pr_number>/merge
git reset --hard FETCH_HEAD
```

[...]

Issue #904, opened 2017-09-27
git revision fails


josete89:
# Report

## What did you do?

danger

## What did you expect to happen?

 Running your Dangerfile without it crashing.

## What happened instead?

+ danger
sh: 2: 83182330e5aaf0f046323f8dfe6377ea63d75159^{commit}: not found
You used `puts` in your Dangerfile. To print out text to GitHub use `message` instead
/usr/local/bundle/gems/git-1.3.0/lib/git/lib.rb:937:in `command': [31m (Danger::DSLError)
[!] Invalid `Dangerfile` file: git '--git-dir=/opt/jenkins/workspace/tration_feature_persistency-QZMBLPRGAKDHVA7NPFO6YBZLTO2TYAOSQBFS4F5NEVKXT7VA6S5A/.git' '--work-tree=/opt/jenkins/workspace/tration_feature_persistency-QZMBLPRGAKDHVA7NPFO6YBZLTO2TYAOSQBFS4F5NEVKXT7VA6S5A' diff '-p' 'bf63ba14b8c56055a5c0ee27bd6858dd9d80ab38
83182330e5aaf0f046323f8dfe6377ea63d75159' 'danger_head'  2>&1:fatal: ambiguous argument 'bf63ba14b8c56055a5c0ee27bd6858dd9d80ab38
83182330e5aaf0f046323f8dfe6377ea63d75159': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'[0m
 #  from Dangerfile:123
 #  -------------------------------------------
 #  
 >  if git.modified_files.include?("swagger.yml")
 #    warn("Swagger modified!")
 #  -------------------------------------------
	from /usr/local/bundle/gems/git-1.3.0/lib/git/lib.rb:351:in `diff_full'
	from /usr/local/bundle/gems/git-1.3.0/lib/git/diff.rb:103:in `cache_full'
	from /usr/local/bundle/gems/git-1.3.0/lib/git/diff.rb:108:in `process_full'
	from /usr/local/bundle/gems/git-1.3.0/lib/git/diff.rb:68:in `each'
	from /usr/local/bundle/gems/danger-5.4.3/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb:75:in `select'
	from /usr/local/bundle/gems/danger-5.4.3/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb:75:in `modified_files'
[...]

PR #967, opened 2018-03-09
Removing `--all` option for `git merge-base`

tbrand:
A patch for fixing
- https://github.com/danger/danger/issues/965
- https://github.com/danger/danger/issues/904
- https://github.com/danger/danger/issues/966

I've tried this patch in my environment and it solves all of the crashes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## master
 
 * Fix GitLab `html_link` url's for external repos. [@sogame](https://github.com/sogame)
+* Removing `--all` option for `git merge-base` [@tbrand](https://github.com/tbrand)
 
 ## 5.5.10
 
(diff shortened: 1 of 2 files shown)


orta:
I'm really wary of making any changes around this (the code churn is really high on this bit of code because there are so many different CI/git setups) can you make this a CLI option instead?

raymondhoagland:
@orta not sure I understand why this would make sense as a CLI option.  Is there an expected benefit to using the `--all` flag?  

The current code only handles if there is one response, but asks for multiple.  That seems like a bug that should be addressed for everyone not made optional unless there is a use case where we'd prefer to get multiple responses and fail for some reason.  If that's the case then maybe documentation/comments are needed to explain why that's desired.

orta:
It's more that I don't entirely have all of danger ruby in my head anymore, and so I value codebase stability. I re-read the history around these changes, and this is probably fine to go 👍

orta:
Shipped as 5.5.11

PR #772, opened 2017-04-04
GitHub file_url API for import_dangerfile

hotbott:
`danger.import_dangerfile(github: 'foo/bar')` was not respecting the `DANGER_GITHUB_HOST` setting.
The `file_url` implementation now calls the API to get the actual `download_url` of a *GitHub* file.
`import_dangerfile` now works with private repos and GitHub Enterprise :D

Fixes #770

@orta Please review

PR #971, opened 2018-03-23
Fix a multistage `import_dangerfile` error

tbrand:
Forks.

The problem is occurred when I use multistage `import_dangerfile` like this.

My Dangerfile
```ruby
danger.import_dangerfile(github: "tbrand/danger1")
```

Dangerfile at `tbrand/danger1`
```ruby
danger.import_dangerfile(github: "tbrand/danger2")
```

Dangerfile at `tbrand/danger2`
```ruby
message "hello"
```

When I execute `danger pr` for it, it falls into infinite loop.

After debugging, I've found that this line disturbs updating `@download_url`.
https://github.com/danger/danger/blob/dbfd46e6350ae5dd30ca3ebcc0872d6b68f1a6cf/lib/danger/request_sources/github/github.rb#L468

So `github.rb#file_url` always returns the same url even if the actual `@download_url` is updated.

The history of the line is here.
https://github.com/danger/danger/pull/772

I may know you've lost a motivation for fixing ruby's danger... :cry:
But it's quite helpful for me if you take a look at this. :smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
+* Fix a Fix a multistage `import_dangerfile` error [@tbrand](https://github.com/tbrand)
+
 ## 5.5.11
 
 * Fix GitLab `html_link` url's for external repos. [@sogame](https://github.com/sogame)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ## master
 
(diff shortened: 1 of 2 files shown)


orta:
Heh, yeah, I leave others to make the fixes now as active dev happens on danger-js. It does mean that it can be hard to grok all the changes that could do occur. Though this codebase is well tested.

PR seems fine to me.

PR #974, opened 2018-03-27
Fix array sublcass typo

ivantse:
Previously, the helper was named `ArraySublcass`. This renames it to `ArraySubclass`.


orta:
Nice - Thanks!

Issue #924, opened 2017-11-19
comparison of Danger::Violation with Danger::Violation failed (ArgumentError)


Nef10:
# Report

## What did you do?

Ran `danger(fail_on_errors: true)` in fastlane on Travis

## What did you expect to happen?

Running my Dangerfile without crashing.

## What happened instead?

```
[23:05:04]: ▸ Coverage report successfully created at path: /Users/travis/build/Nef10/MyProjectName/fastlane/xcov_report/tmp/report.json20171118-3166-9l7ts0
/Users/travis/.rvm/gems/ruby-2.4.2/gems/danger-5.5.5/lib/danger/request_sources/github/github.rb:503:in `sort': comparison of Danger::Violation with Danger::Violation failed (ArgumentError)
	from /Users/travis/.rvm/gems/ruby-2.4.2/gems/danger-5.5.5/lib/danger/request_sources/github/github.rb:503:in `inline_violations_group'
	from /Users/travis/.rvm/gems/ruby-2.4.2/gems/danger-5.5.5/lib/danger/request_sources/github/github.rb:163:in `update_pull_request!'
	from /Users/travis/.rvm/gems/ruby-2.4.2/gems/danger-5.5.5/lib/danger/danger_core/dangerfile.rb:248:in `post_results'
	from /Users/travis/.rvm/gems/ruby-2.4.2/gems/danger-5.5.5/lib/danger/danger_core/dangerfile.rb:278:in `run'
	from /Users/travis/.rvm/gems/ruby-2.4.2/gems/danger-5.5.5/lib/danger/danger_core/executor.rb:27:in `run'
	from /Users/travis/.rvm/gems/ruby-2.4.2/gems/danger-5.5.5/lib/danger/commands/runner.rb:66:in `run'
	from /Users/travis/.rvm/gems/ruby-2.4.2/gems/claide-1.0.2/lib/claide/command.rb:334:in `run'
	from /Users/travis/.rvm/gems/ruby-2.4.2/gems/danger-5.5.5/bin/danger:5:in `<top (required)>'
	from /Users/travis/.rvm/gems/ruby-2.4.2/bin/danger:23:in `load'
	from /Users/travis/.rvm/gems/ruby-2.4.2/bin/danger:23:in `<main>'
	from /Users/travis/.rvm/gems/ruby-2.4.2/bin/ruby_executable_hooks:15:in `eval'
[...]

PR #978, opened 2018-04-16
Fix crash on nil lines

fwal:
Initially reported in #924

Having multiple messages with the same file but at least one without line would cause sorting to crash as `nil <=> 1` is `nil`.

```ruby
# This was fine
warn("A warning", file: "A.file", line: nil) 

# This crashed
warn("A warning", file: "A.file", line: 1)
warn("A warning", file: "A.file", line: nil)
```


orta:
Nice spot - thanks!

Issue #952, opened 2018-02-05
Not a Pull Request - skipping `danger` run


Azuka:
# Report

## What did you do?

Run `bundle exec danger --verbose` and `bundle exec pr --verbose`

## What did you expect to happen?
Expected the first command to write a comment on my pull request. The second should show output without commenting.

Suggestion: verbose should show what CI it detected as well as the steps it took to decide it wasn't a PR. This should help when troubleshooting. 

## What happened instead?
The first showed "Not a Pull Request - skipping `danger` run". Locally, I did an export of all the environment variables from CircleCI and was able to run danger with no problems, including writing a comment to the PR. The second command works as-expected.

This just started happening on Circle 2.0 (on an enterprise instance with enterprise GitHub).

## Your Environment

* Which CI are you running on? `CircleCI`
* Are you running the latest version of Danger? `Yes`
* What is your Dangerfile?

  ```ruby
  # please paste here, with 2-space identation, thank you!
  # Sometimes it's a README fix, or something like that - which isn't relevant for
  # including in a project's CHANGELOG for example
  declared_trivial = github.pr_title.include? "#trivial"

  # Make it more obvious that a PR is a work in progress and shouldn't be merged yet
  warn("PR is classed as Work in Progress") if github.pr_title.include? "[WIP]"

  # Warn when there is a big PR
  warn("Big PR") if git.lines_of_code > 500

  # Warn for Todos
  todoist.warn_for_todos

  # JUnit report
  junit.parse "#{ENV['CIRCLE_TEST_REPORTS']}/report.xml"
  junit.report

  # Coverage Report
  username = ENV['CIRCLE_PROJECT_USERNAME']
  project_name = ENV['CIRCLE_PROJECT_REPONAME']
  build_number = ENV['CIRCLE_BUILD_NUM']
  circle_token = ENV['DANGER_CIRCLE_CI_API_TOKEN']
  circle_artifacts = ENV['CIRCLE_ARTIFACTS']
  if username && project_name && build_number
    # submit message giving the coverage report that was generated by coverage.py
    message("[html coverage report](https://circleci.mheducation.com/gh/MHEducation/#{project_name}/#{build_number}/artifacts/0/#{circle_artifacts}/htmlcov/index.html?circle-token=#{circle_token})")
  end

  # Warn about config.go changes
  diff = git.diff_for_file("app/config/config.go")
  if diff && !diff.patch.empty?
[...]

PR #982, opened 2018-04-29
Notes which CI is used when exiting with an invalid PR

dbgrandi:
Hiya 👋 

When danger runs and finds a valid CI environment, but does not find a valid PR, you end up with a failure message like `Not a Pull Request - skipping 'danger' run`

This PR injects the CI name into that error message. e.g. `Not a Travis Pull Request - skipping 'danger' run`, to add a little more context.

Handy for debugging on non-supported CIs. Yep, there are still a few of those left. :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## master
 
 * Enable Bitbucket Cloud for Bitrise CI [@masahide318](https://github.com/masahide318)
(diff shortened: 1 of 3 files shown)


dbgrandi:
Oh, looks like this may partially help https://github.com/danger/danger/issues/952

orta:
Yeah, let's get this shipped

PR #986, opened 2018-05-15
Fix crash when messages are nil

allewun:
If `nil` is unexpectedly passed to `message|warn|fail`, Danger will crash. While the Dangerfile should resolve this so that it doesn't happen in the first place, it's probably safer to do nothing IMO.


orta:
Agree 👍

PR #992, opened 2018-05-30
Update screwdriver ci initializer

fandyfyf:
SCM_URL usually include a branch name

Ex. `git@github.com:danger/danger.git#branch`

repo_slug -> `danger/danger`
repo_url -> `git@github.com:danger/danger.git`

with the current logic, Danger will pick up a wrong repo_slug which causes failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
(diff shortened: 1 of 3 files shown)


orta:
These fails are pretty legit, just some minor formatting - everything else looks good

fandyfyf:
Thanks! I am still new to ruby. :)

orta:
Rockin' - looks good

PR #1002, opened 2018-07-18
Allow two slashes in URL slug

dstranz:
Fix for issue described in #968 - correct slug detection for repository with following url:
`https://git.myhost.com/ORGANISATION/TEAM/REPOSITORY_NAME` allowed by GitLab.

I cannot simply add new regexp group `(/[^/]+){1,2}` because then it will also match part of domain name, so I added `://[^/]+/` at the beginning to catch only url path.

PR #1003, opened 2018-07-23
Move docs on the CI

orta:
Moves the doc checker from inside Danger evaluation, to on CI like normal. Might get better error messaging - which could disambiguate #1002


orta:
OK, wanna try rebase @dstranz

PR #1007, opened 2018-08-07
Move rubocop to Rake task and autocorrect errors #trivial

stephengroat:
moves rubocop to a Rake task and autocorrects associated errors


orta:
I just moved them out of there so it was more obvious what was failing on CI for people -but I'm cool with this

Issue #1006, opened 2018-08-06
Raise doesn't trigger an error. 


damacus:
# Report

## What did you do?

Replace `fail` with `raise` in the Dangerfile.

I'm sure there's a good reason for the reasoning behind it. I just want to know if I should stop everyone trying to `rubocop -a` this file rom raise to fail ;)

## What did you expect to happen?

Danger fails a PR

## What happened instead?

Nothing

## Your Environment

* Which CI are you running on? CircleCI
* Are you running the latest version of Danger? Yes always just installed as part of CI using gem install
* What is your Dangerfile?

  ```ruby
def code_changes?
  code = %w(libraries attributes recipes resources)
  code.each do |location|
    return true unless git.modified_files.grep(/#{location}/).empty?
  end
  false
end

def test_changes?
  tests = %w(spec test .kitchen.yml .kitchen.dokken.yml)
  tests.each do |location|
    return true unless git.modified_files.grep(/#{location}/).empty?
  end
  false
end

raise 'Please provide a summary of your Pull Request.' if github.pr_body.length < 10

raise 'Please add labels to this Pull Request' if github.pr_labels.empty?

warn 'This is a big Pull Request.' if git.lines_of_code > 400

# Require a CHANGELOG entry for non-test changes.
if !git.modified_files.include?('CHANGELOG.md') && code_changes?
  raise 'Please include a [CHANGELOG](https://github.com/sous-chefs/line-cookbook/blob/master/CHANGELOG.md) entry.'
end

# A sanity check for tests.
if git.lines_of_code > 5 && code_changes? && !test_changes?
  warn 'This Pull Request is probably missing tests.'
end

  ```

orta:
I guess it could? That rubocop rule came up after building danger, I didn't even know that `fail` was an alias to `raise` - it just felt like the right word at the time.

So long as the fail/raise override only happens during Dangerfile eval, I'm fine to add it as an alias to the existing `fail` method.

damacus:
It's slightly different semantically, fail isn't going to be a caught exception whereas raise is expect to be caught. In Danger's context I think *either* is fine to be honest.

damacus:
Any chance we can get this monkey patched in please? :)

orta:
Sure, you’re welcome to take a look at it

PR #1014, opened 2018-09-10
Adds failure as the recommended way to express a blocking error in a Dangerfile

orta:
I ended up adding a dangerfile to one of our ruby apps, and obviously this was the right fit https://github.com/artsy/exchange/pull/168 - but the default rubocop settings don't want you writing `fail`. 

I originally aliased `raise` - like mentioned in https://github.com/danger/danger/issues/1006 but it didn't feel right, so I chatted with @yuki24 added `failure` as an option, and made it the default in all the docs.

### DIFF
--- a/lib/danger/danger_core/plugins/dangerfile_messaging_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_messaging_plugin.rb
@@ -17,12 +17,12 @@ module Danger
   #
   # message 'Hello', 'World', file: "Dangerfile", line: 1
   # warn ['This', 'is', 'warning'], file: "Dangerfile", line: 1
-  # fail 'Ooops', 'bad bad error', sticky: false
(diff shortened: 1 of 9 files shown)


orta:
( This doesn't break anything, so I'm shipping it as a patch 💯 )

lib/danger/danger_core/plugins/dangerfile_messaging_plugin.rb
```diff
@@ -178,6 +178,8 @@ def fail(*failures, **options)
       end
     end
 
+    alias_method :failure, :fail
```
yuki24: Are you going to deprecate it eventually or `fail` will be available as-is but it's just not recommended?

orta: Just not recommended, no plans to remove it - when I pitch Danger Ruby is mature and not breaking anymore, I wanna live up to that.

PR #1018, opened 2018-09-14
fix `--no-ansi` behavior for commands

otaznik-net:
colored2 seems to be used as drop in replacement for colored. It works
great unless you want to disable ansi codes in output as `ansi` flag is
handled by CLAide and not by Danger. Added
`Danger::Runner#adjust_colored2_output` is called from initializer
as that is only place common for commands. Otherwise some wrapper of
Runner#run will be needed (which does not seems worth it).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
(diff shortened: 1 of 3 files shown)


orta:
Cool. yeah, this makes sense - thanks!

PR #1019, opened 2018-09-20
Add dry_run command and associated ci_source and request_source

otaznik-net:
motivation is ability to run danger against local only repo
in condition when there is no actual PR/MR created (or is not available
to danger).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 -->
 
 ## master
+* Adds `dry_run` command to allow running danger on localhost without actual PR/MR [@otaznik-net](https://github.com/otaznik-net)
 
 ## 5.6.7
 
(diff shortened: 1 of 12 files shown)


orta:
Ah nice yeah, this is basically what `danger local` is in Danger JS 👍 

The build failed because the internal plugin you made needs documenting as it ends up in the reference page of danger.systems

> Errors
>  - Examples - DangerfileLocalOnlyPlugin (class):
>    - You should include some examples of common use-cases for your plugin.
>    - @see - https://github.com/dbgrandi/danger-prose/blob/v2.0.0/lib/danger_plugin.rb#L8#-L27
>   - /home/travis/build/danger/danger/lib/danger/danger_core/plugins/dangerfile_local_only_plugin.rb:10

otaznik-net:
Oh, sorry for letting the PR broken (I was little too tired to wait for tests =/ - ashamed), will fix that today or tomorrow ...
and thanks for instruction how to fix ;o)

=)K

orta:
👍 looking good, once #1020 is ready I'll ship a release

PR #1020, opened 2018-09-28
Decouple remove_previos_comments logic

JoeSSS:
Decouple `remove_previos_comments` logic from the logic of having nothing to say

Fixing an issue from https://github.com/danger/danger/pull/949#issuecomment-425128457

cc @znacloud : I do not use gitlab. So it would be very nice if you can verify this patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ## master
 
+## 5.6.8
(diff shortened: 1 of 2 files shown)


JoeSSS:
@znacloud can you please create a PR with your changes? :)

Issue #877, opened 2017-08-10
Invalid `Dangerfile` file: unknown keywords: organisation, repository, branch, path


adam-moss:
# Report

## What did you do?
```sh
danger --verbose
```

## What did you expect to happen?

Danger runs local Dangerfile, importing and running shared Dangerfile

## What happened instead?
```sh
/home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb:157:in `import_dangerfile_from_github': (Danger::DSLError)
[!] Invalid `Dangerfile` file: unknown keywords: organisation, repository, branch, path
 #  from Dangerfile:3
 #  -------------------------------------------
 #  
 >  danger.import_dangerfile(gitlab: "org/repo", branch: "branch/name")
 #  
 #  -------------------------------------------
	from /home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb:80:in `import_dangerfile'
	from Dangerfile:3:in `block in parse'
	from /home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/danger_core/dangerfile.rb:199:in `eval'
	from /home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/danger_core/dangerfile.rb:199:in `block in parse'
	from /home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/danger_core/dangerfile.rb:195:in `instance_eval'
	from /home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/danger_core/dangerfile.rb:195:in `parse'
	from /home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/danger_core/dangerfile.rb:272:in `run'
	from /home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/danger_core/executor.rb:27:in `run'
	from /home/user/.gem/ruby/gems/danger-5.3.4/lib/danger/commands/runner.rb:66:in `run'
	from /usr/local/share/gems/gems/claide-1.0.2/lib/claide/command.rb:334:in `run'
	from /home/user/.gem/ruby/gems/danger-5.3.4/bin/danger:5:in `<top (required)>'
	from /home/user/bin/danger:23:in `load'
	from /home/user/bin/danger:23:in `<main>'
ERROR: Job failed: exit status 1
```

## Your Environment

* Which CI are you running on?
GitLab CI
[...]

PR #1022, opened 2018-10-03
Fix gitlab source for #import_dangerfile

koffeinfrei:
Implements the method `Danger::RequestSources::GitLab#file_url` such that importing a danger file from a GitLab repository works. Example: 

```ruby
danger.import_dangerfile(gitlab: "ruby-grape/danger")
```

This is documented at https://danger.systems/reference.html but currently doesn't work due to the missing implementation.

Fixes https://github.com/danger/danger/issues/877

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 -->
 
 ## master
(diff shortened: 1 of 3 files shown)


orta:
Great - perfect 👍

PR #1023, opened 2018-10-09
Replacing spaces with %20 in VSTS host urls

MadsBogeskov:
In VSTS you can have spaces in the URL to the repository, however URI doesn't support this. The solution I have implemented is to simply replace these spaces with %20 which is what VSTS itself does.


MadsBogeskov:
Ill implement a test later this evening for this.

PR #1024, opened 2018-10-09
Use GitLab API to determine base_commit

### LINKED ISSUE #1009: base_commit for gitlab wrong in certain workflows
# Report

## What did you do?

Run danger on a branch when its merge request has been updated with a merge from the target branch (rather than rebasing) 

Similar to bitbucket or github, ideally the base_commit should have been updated to the tip of the target branch at the time of merge request being updated.

## What happened instead?

The base_commit was unchanged, and now the modified files list in danger includes changes that have gone into master.

## Your Environment

* Which CI are you running on?

GitLab CI

* Are you running the latest version of Danger?

yes

* What is your Dangerfile?

https://gitlab.com/charts/gitlab/blob/master/scripts/support/changelog/Dangerfile

leipert:
The `base_commit` is falsely calculated if the target branch is merged into the MRs branch. GitLab API v4 exposes the correct hashes under a field called [`diff_refs`](https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr).

Fixes #1009 

- - -

It seems like danger's GitLab integration was build for the old API v3,
so it still needed to calculate the `base_commit` itself. With the API
v4 (which has been around since July 2016) the Merge Request actually
exposes the `base_commit`.

With this commit we use `#diff_refs#base_sha` to determine the
`base_commit` and `#diff_refs#head_sha` to determine the `head_commit`.
No need to retrieve commits anymore and calculate the SHAs from there.


leipert:
@orta I finalized this MR. If you have any questions, feel free to ask me. I am not a ruby pro, but we at @gitlab really need it.

orta:
This looks great to me 👍 

Let's get this in and shipped

leipert:
Thanks for the fast turn-around and the org invite!

I see that `danger.js` is missing GitLab support completely. Maybe my next weekend thing 😆

PR #1036, opened 2018-11-14
Protect against 'path' reassignment when evaling dangerfile

daveallie:
Move eval into seperate method to prevent local variable reassignment.
If the eval rescue is hit and path has been reassigned Danger will crash.
This doesn't address any concerns around a malicious Dangerfile rewriting methods.

#trivial


orta:
Hah, yea, that's an interesting edge case 👍

PR #1038, opened 2018-11-19
BitbucketServerAPI Error Message

cnoon:
This PR updates the error message thrown by Danger when posting a comment to Bitbucket Server and receiving a 4XX or 5XX status code. 

---

### Problem Statement

We've been spinning up new projects left and right at Nike with Danger (thanks for the fantastic project BTW! 🍻 ). When spinning up new projects, we tend to have plugins like SwiftLint misconfigured a bit which can end up triggering large numbers of warnings in Danger. This generally results in the following error thrown by Danger:

```
Error posting comment to Bitbucket Server: 400 ()
```

As you can see, this doesn't give us much to go on. We have no idea why the 400 is being thrown. Room for improvement!

---

### Solution

By appending the response body to the error message, we are able to see why Bitbucket Server is throwing the 400 status code:

```
Error posting comment to Bitbucket Server: 400 () - {"errors":[{"context":"text","message":"Please enter a non-empty value less than 32768 characters","exceptionName":null}]}
```

With this change, we'll be able to see that the payload we're sending to Bitbucket is too large. As far as I can tell, the CHANGELOG entry should be sufficient for this PR. If you'd like me to update anything else, please don't hesitate to ask.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
+* Update `BitbucketServerAPI` error message to include response body [@cnoon](https://github.com/cnoon)
+
 ## 5.8.0
 
 * Add AppVeyor support [@tumugin](https://github.com/tumugin)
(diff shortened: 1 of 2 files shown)


orta:
Yep, this is perfect - merge on green!

PR #1039, opened 2018-11-20
Fix diff range detection logic

r7kamura:
## Problem

I added the following line into my Dangerfile, and created a pull request including this change.

```rb
# Dangerfile
warn("foo", file: "Dangerfile", line: 1)
```

### Expected

An inline-comment should be posted on the diff for Dangerfile#L1.

### Actual

A normal comment was posted like this:

image

## Solution

In many versions of GNU diff, each range can omit the comma and trailing value, in which case the value is 1. (See https://en.wikipedia.org/wiki/Diff for more details about this.)

Here is an example of the omission. Look at `@@ -1 +1,8 @@`.

image

I changed the regex pattern in Danger::RequestSources::GitHub#find_position_in_diff so that it supports the omitted format on the range for the original file too.

```rb
range_header_regexp = /@@ -([0-9]+),([0-9]+) \+(?<start>[0-9]+)(,(?<end>[0-9]+))? @@.*/
                                   ^^^^^^^^^
```

I think this case occurs when we prepended some lines into the head of an existing file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
+* Fix diff range detection logic [@r7kamura](https://github.com/r7kamura)
+
 ## 5.8.1
 
 * Update `BitbucketServerAPI` error message to include response body [@cnoon](https://github.com/cnoon)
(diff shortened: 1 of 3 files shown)


orta:
Great work @r7kamura  - this looks 💯

PR #1045, opened 2018-11-28
Add git.tags

hotbott:
Add `git.tags` to get a list of existing Git tags.

I want to check on a PR level if the version was already released.

@orta Please review


orta:
Yeah, I think this makes sense 👍

orta:
Ace, thanks!

PR #1046, opened 2018-12-09
Amend diff position caluculation

colorbox:
## Problem
`find_position_in_diff` method returns wrong position when there are some addition codes after `No newline...` annotation in diff.

Because `find_position_in_diff` method doesn't consider about annotation.

## Solution
Add extra position increments when the line is `No newline...` annotation.


colorbox:
I'll fix CI.

orta:
👍

PR #1048, opened 2018-12-10
Fix wrong test

colorbox:
My PR https://github.com/danger/danger/pull/1046 included wrong test related with backslash.
So the implementation is not correct.
This PR fix it.


orta:
We've all been there, thanks! Merge on green.

PR #1066, opened 2019-01-08
Fix compatibility with gitlab 11.6+ Merge Requests API

### LINKED ISSUE #1065: Gitlab API after 11.6.1 use IID for merge_requests
Gitlab now uses IID instead of ID on API call for merge_requests endpoint

https://docs.gitlab.com/ee/api/README.html#id-vs-iid

https://github.com/danger/danger/blob/ca052e180f8973e0b28b2845a489f27d0996ab1f/lib/danger/ci_source/gitlab_ci.rb#L35

leslau:
Fix #1065

### DIFF
--- a/lib/danger/ci_source/gitlab_ci.rb
+++ b/lib/danger/ci_source/gitlab_ci.rb
@@ -32,6 +32,7 @@ module Danger
     end
 
     def self.determine_merge_request_id(env)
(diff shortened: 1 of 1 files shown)


orta:
I *think* this is deprecated by #1071  -so I'm going to close this PR, thanks for giving it a shot @leslau

PR #1067, opened 2019-01-08
Support more bundlers

orta:
CI is failing, lets see what happens when we allow 2.x


orta:
Fixed by https://github.com/danger/danger/pull/1073

PR #1068, opened 2019-01-13
Run "danger pr" for Bitbucket Cloud

RicardoBelchior:
Adding changes to `danger pr` command in order to allow running on bitbucket cloud repositories.

There's a few things I was unsure, but since the tests still passed, I kept going. Also my ruby skills are not great, so feel free to criticise..

For example, file `lib/danger/danger_core/plugins/dangerfile_bitbucket_cloud_plugin.rb` seemed very buggy, mostly being a copy paste from the bitbucket_server equivalent.

The method `#find_scm_provider` is not great but I followed a similar technique to what was previously being done to detect bitbucket server urls. I suppose changing that would involve a bigger refactor on the `pr` command.

And that was it.. I was able to run this locally, so I suppose it works. Let me know if you need any other changes.

### DIFF
--- a/lib/danger/ci_source/support/pull_request_finder.rb
+++ b/lib/danger/ci_source/support/pull_request_finder.rb
@@ -62,18 +62,29 @@ module Danger
     end
 
     def generate_remote_pull_request
-      if remote_url =~ %r{/pull-requests/}
+      scm_provider = find_scm_provider(remote_url)
+
+      case scm_provider
+      when :bitbucket_cloud
+        RemotePullRequest.new(
+          remote_pull_request[:id].to_s,
+          remote_pull_request[:source][:commit][:hash],
+          remote_pull_request[:destination][:commit][:hash]
+        )
+      when :bitbucket_server
         RemotePullRequest.new(
           remote_pull_request[:id].to_s,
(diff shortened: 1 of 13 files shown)


lib/danger/commands/pr.rb
```diff
@@ -8,7 +8,7 @@
 
 module Danger
   class PR < Runner
-    self.summary = "Run the Dangerfile locally against Pull Requests (works with forks, too). Does not post to the PR.".freeze
+    self.summary = "Run the Dangerfile locally against Pull Requests (works with forks, too). Does not post to the PR. Usage: danger pr <URL>".freeze
```
orta: nice work

lib/danger/ci_source/support/pull_request_finder.rb
```diff
@@ -142,5 +165,15 @@ def api_url
...
     end
+
+    def find_scm_provider(remote_url)
+      if remote_url =~ %r{/bitbucket.org/}
+        :bitbucket_cloud
+      elsif remote_url =~ %r{/pull-requests/}
+        :bitbucket_server
+      else
+        :github
+      end
```
orta: 👍

RicardoBelchior:
That's awesome! Thank you.

Issue #1080, opened 2019-02-11
Fail if no PR?


ghiculescu:
Not sure if this is the right place to ask, but would you accept a PR that adds a command line flag for Danger to fail if there's no PR found?

Currently it will `exit(0)`: https://github.com/danger/danger/blob/master/lib/danger/danger_core/executor.rb#L61

I would like to have that cause a red build, so that we can ensure that danger has 100% run before we merge. We use [CircleCI](https://danger.systems/guides/troubleshooting.html#circle-ci-doesnt-run-my-build-consistently) for context.

orta:
I'm pretty sure that flag already exists?

ghiculescu:
Not sure if ^ is the right label, but I'm happy to build this if it will get accepted.

ghiculescu:
@orta 	I'm not seeing it in https://github.com/danger/danger/blob/master/lib/danger/danger_core/executor.rb#L7

There's `fail_on_errors` but that only fails if Danger runs - which won't happen if no PR is found. (Which tbh I would consider an error, but I'm wary of repurposing that flag.)

orta:
Oh, wait, you want it when it's not a PR - yeah, there wouldn't be something for that. 

Yeah, I don't see a reason why not. Most CIs do both PR and the merge build by default, so it seems fine as an edge case.

orta:
Sure, give it a shot

PR #1081, opened 2019-02-11
Add "Fail if no PR" flag

ghiculescu:
Flag is disabled by default, but can be opted in which might be handy for Circle CI users, or any other CI system where you can't guarantee Danger will run for each PR. Fixes https://github.com/danger/danger/issues/1080

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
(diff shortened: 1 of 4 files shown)


orta:
Yep, looks good. Thanks.

I'll ship you a release

ghiculescu:
Thank you!

PR #1082, opened 2019-02-14
Added support for GitLab inline comments

pbendersky:
( when running on GitLab >= 10.8.0)

Some things I'm not sure and would like feedback on:

- Are there enough tests for this to be merged?
- Are we ok with how the two cases are handled? My thinking here was that the non-inline comments will be short-lived (at some point, we might decide to only support GitLab starting with a newer version), so it wasn't worth doing something like a strategy for this case.


orta:
I don't know much about the GitLab API for this, but in GitHub we have to make sure the inline comments are inside the diff (or they don't get posted) - this PR doesn't do that, do you know if you need to?

pbendersky:
I saw it, and it doesn't appear to be needed... Just including the new path of the file and the line seems to be enough (and the commit hashes, that we always have).

orta:
Cool, then yeah - this feels about right then 👍 

I think it's worth adding a test or two about editing existing messages, and deleting them - as historically that's a place that's broken in both ruby and JS

pbendersky:
Will do. What's the desired funcionality? When is a comment edited / deleted?

orta:
If one danger run adds inline comments, then a subsequent run with different warn/fail/messages

- deletes inline comments that don't up the 2nd time
- edits inline comments that are on the same `file:line:` but with different test

A very naive implementation (which would send a lot of notifications) would be to wipe all inline comments at the start of a run, then create all the new ones.

pbendersky:
I think my latest commit covers the missing cases with specs. In any case, feel free to review and comment :)

orta:
I've merged master, which should hopefully fix CI - if so, this is good to go IMO

orta:
Looks like there's a bunch of failing specs, any idea what's up there

pbendersky:
I'll check, they were passing for me...

orta:
Hrm, looks like an old ruby (which I still want to support alas) issue

pbendersky:
What version should it support? I'll fix the specs in that version.

orta:
2 and above really, I base that on what Xcode supports :D

orta:
Oh interesting, IMO we can make those tests somewhat consistent with the version of the gitlab gem you have

pbendersky:
How would that work? Older versions of the `gitlab` gem use the (now deprecated) `v3` API...

orta:
Detect that and skip the tests/

pbendersky:
@orta not sure why Circle failed the last time, but I think the tests are ok now.

orta:
Cool, yeah, looked like a circle bug. This feels 👍  to me

PR #1100, opened 2019-03-06
Support Kramdown 2.0+ (requires Ruby 2.3)

davidstosik:
Kramdown released [2.0.0](https://kramdown.gettalong.org/news.html#kramdown-200-released) on 2019/01/20, followed by [2.1.0](https://kramdown.gettalong.org/news.html#kramdown-210-released) on 2019/01/31.
For projects staying on the edge, it is important that Danger's dependency on Kramdown `~> 1.5` does not become a blocker.

As it is not possible to write a gemspec with conditions on the Ruby version, this is my final proposition:

## Suggest drastic update of Danger that will require Ruby 2.3

This is a drastic change triggered by Kramdown 2.x requiring Ruby 2.3 and `kramdown-parser-gfm`. Danger 6.0+ will require:
- Ruby 2.3+
- kramdown gem 2.0+
- kramdown-parser-gfm gem

To keep using Ruby 2.0\~2.3 and kramdown gem 1.5\~2.0, you'll have to stick to Danger 5.x.

I'd like to suggest creating a `v5.x` branch on this repository, right before this PR gets merged. If there is a desire to keep maintaining Danger 5.x for a while, patches would need to be applied to both `master` and `v5.x`.

### DIFF
--- a/danger.gemspec
+++ b/danger.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.required_ruby_version = ">= 2.0.0"
+  spec.required_ruby_version = ">= 2.3.0"
 
   spec.add_runtime_dependency "claide", "~> 1.0"
   spec.add_runtime_dependency "claide-plugins", ">= 0.9.2"
@@ -26,12 +26,8 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "colored2", "~> 3.1"
   spec.add_runtime_dependency "faraday", "~> 0.9"
   spec.add_runtime_dependency "faraday-http-cache", "~> 1.0"
-  if RUBY_VERSION >= "2.3.0"
-    spec.add_runtime_dependency "kramdown", "~> 2.0"
-    spec.add_runtime_dependency "kramdown-parser-gfm", "~> 1.0"
-  else
-    spec.add_runtime_dependency "kramdown", "~> 1.5"
-  end
+  spec.add_runtime_dependency "kramdown", "~> 2.0"
+  spec.add_runtime_dependency "kramdown-parser-gfm", "~> 1.0"
   spec.add_runtime_dependency "octokit", "~> 4.7"
   spec.add_runtime_dependency "terminal-table", "~> 1"
   spec.add_runtime_dependency "cork", "~> 0.1"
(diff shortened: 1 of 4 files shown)


orta:
Cool, yeah - we'll need it back in the gemspec - but I'm open to what the dependency looks like 👍

davidstosik:
@orta Thanks for the early review! 🙇 
Still trying to figure things out with Gemfile, gemspec and dependency, but I'll get there. 😅

danger.gemspec
```diff
@@ -26,8 +26,13 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "colored2", "~> 3.1"
   spec.add_runtime_dependency "faraday", "~> 0.9"
   spec.add_runtime_dependency "faraday-http-cache", "~> 1.0"
+  if RUBY_VERSION >= "2.3.0"
```
davidstosik: I thought dependencies where sorted alphabetically, so I put this conditional here, but then noticed `cork` is after `terminal-table`. I'm open to moving this somewhere else if it looks better.

orta: I'm not too precious on the order etc, this is totally fine

Looks like unfortunately this technique doesn't work though: https://stackoverflow.com/questions/39346334/manage-gem-dependencies-based-on-ruby-version

There might not even be a way to do this, so maybe we'll need to go to use 2.0 only? Does that mean forcing 2.3.0 only?

davidstosik: Yes, kramdown >2 requires Ruby 2.3. kramdown-parser-gfm has dependencies that will fail to install with Ruby < 2.3...

Maybe Danger needs a major version update?
- keep requiring `"kramdown", "~> 1.5"` in 5.x
- switch to `"kramdown", "~> 2.0"`  with the addition of `"kramdown-parser-gfm, "~> 1.0"` in 6.x, with the added requirement of Ruby 2.3+

5.x can keep getting bugfixes for a few months to allow a smooth transition.

davidstosik:
♻️ This is a drastic suggestion, but I'm interested in alternatives if there are any.

Maybe could use a different gem to parse Github-flavored Markdown (such as [CommonMarker](https://github.com/gjtorikian/commonmarker))?

orta:
OK, so 2.3.0 has been default on OSX for 2 versions - which makes me feel pretty reasonable about it (yeah, I know it's somewhat arbitrary, but I want to keep Danger somewhat inline with CocoaPods' support systems because then it's less for me to think about)

So, I'm fine with having a major semver bump to 6 and taking this, I want to let it sit till the weekend incase I think of something else, but this should be fine 👍

orta:
OK, let's do it - merge on green

PR #1119, opened 2019-04-10
Update bundler gem

RDIL:
We should stay up to date on this

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
         - RACK_ENV=test
     steps:
       - checkout
-      - run: gem update --system 2.4.8
+      - run: gem update --system 2.5.0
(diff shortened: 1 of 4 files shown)


orta:
This won't update the gem, you'd also need to update the lockfile also

RDIL:
I don't see a lockfile anywhere...

orta:
Ah, yeah, looks like we manually install bundler on CIs with a specific version: https://github.com/danger/danger/blob/master/.travis.yml#L16

So if this works that's fine during dev then that might be ok, but we had bundler 2.0 problems for a while

RDIL:
Circle passed, now just travis and appveyor

RDIL:
Travis is green

orta:
👍 - nice one, thanks!

PR #1122, opened 2019-04-15
For GitLab >= 10.7, use a faster API to find the MR for a commit

craigmiskell-gitlab:
Original code queries all open PRs for the project, iterating over them looking for the one that has the commit SHA.  This can take quite some time if there are a lot of MRs open.  For example https://gitlab.com/gitlab-org/gitlab-ce/merge_requests has ~800 open MRs, which causes (@ 20 per page) 40 API requests to fetch, each doing an HTTPS connection setup, plus run-time, which really adds up.

Since GitLab 10.7 there is an API which will directly find/return any MRs for a given commit; this takes one API call, and is speedy.  This PR uses this new API, if GitLab is new enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
+* For GitLab >= 10.7, use a newer faster API to find the MR for the commit
+
 ## 6.0.3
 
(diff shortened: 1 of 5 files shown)


orta:
Very cool, looks good! - merge on green.

RDIL:
Merging now

PR #1123, opened 2019-04-19
Fixed missing Unified diff headers for `gitlab.mr_diff`

matteomatassoni:
Fixed missing Unified diff headers for `gitlab.mr_diff` due to a Gitlab API bug (See https://gitlab.com/gitlab-org/gitlab-ce/issues/53229).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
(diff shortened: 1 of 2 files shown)


orta:
A test for it would be nice but not a blocker for now 👍

PR #1127, opened 2019-05-10
Update faraday-http-cache dependency from 1.0 to 2.0

ignazioc:
The version `2.0` of `faraday-http-cache` was released 3 years ago. Pinning Danger to the version `~> 1.0` of that gem makes it incompatible with other recent gems.


orta:
👍

PR #1128, opened 2019-05-15
Update way to import Dangerfile from Gitlab

manu-d:
Gitlab added restrictions for security reasons, and you can no longer download a raw file without going through their API (Cf: https://gitlab.com/gitlab-org/gitlab-ce/issues/54572#note_120858359)
Therefore, using `danger.import_dangerfile(gitlab: "repo", path: 'Dangerfile')` would fail.
To fix this, in this PR, I updated the `file_url` for Gitlab request source, and updated the tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## master
 
 * Update `faraday-http-cache` dependency from 1.0 to 2.0
+* Fixed importing Dangerfile from Gitlab (uses Gitlab API now)
 
 ## 6.0.6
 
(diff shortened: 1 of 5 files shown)


orta:
I wonder what's up with the code coverage thing?

manu-d:
> I wonder what's up with the code coverage thing?

I noticed that when doing `context "Gitlab", host: :gitlab do` the tests actually don't run, but I can't understand why

manu-d:
@orta I updated the PR, tests are now passing

orta:
Thanks @manu-d - alright, let's get this shipped

PR #1133, opened 2019-05-23
Fix branch_for_head api of Bitbucket cloud

susan335:
`branch_for_head` and `branch_for_base` return a same value on Bitbucket cloud.

It is simple.
get the value from the same key in Json.


orta:
Yeah, this looks good :+1:

PR #1134, opened 2019-06-04
feat(codefresh): Codefresh Provider

br4in3x:
This PR introduces support of [Codefresh CI/CD](http://codefresh.io) system. 

The code was heavily inspired (I would even say ported) with TypeScript version of Danger - https://github.com/danger/danger-js/blob/master/source/ci_source/providers/Codefresh.ts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ## master
 
+## 6.0.10
+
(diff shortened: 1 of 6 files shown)


orta:
Looks great 👍

br4in3x:
@orta Thanks a lot! I just refactored a little, please see: https://github.com/danger/danger/compare/master...br4in3x:feat/codefresh. Do you think I should create another PR since this one is already merged? Or it is good enough? :)

Issue #1135, opened 2019-06-18
Danger crashes when not running on PR in CircleCI


mvandervelden:
# Report

## What did you do?

* Update Danger from 5.5.13 to 6.0.9

## What did you expect to happen?

* Running Dangerfile on master branch on CircleCI should report `Not a CircleCI Pull Request - skipping ``danger`` run`, without crashing/failing the build.

## What happened instead?

* Crash:
```
bundler: failed to load command: danger (/Users/distiller/project/vendor/bundle/ruby/2.4.0/bin/danger)
NoMethodError: undefined method `first' for nil:NilClass
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/ci_source/circle_api.rb:38:in `fetch_pull_request_url'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/ci_source/circle_api.rb:24:in `pull_request_url'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/ci_source/circle_api.rb:8:in `pull_request?'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/ci_source/circle.rb:59:in `validates_as_pr?'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/danger_core/environment_manager.rb:15:in `pr?'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/danger_core/executor.rb:63:in `validate_pr!'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/danger_core/executor.rb:47:in `validate!'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/danger_core/executor.rb:22:in `run'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/lib/danger/commands/runner.rb:72:in `run'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/claide-1.0.2/lib/claide/command.rb:334:in `run'
  /Users/distiller/project/vendor/bundle/ruby/2.4.0/gems/danger-6.0.9/bin/danger:5:in `<top (required)>'
[...]

PR #1140, opened 2019-07-09
Add extra check in fetching Circle CI pull request url

mvandervelden:
It appears that the `pull_requests` field can be nil in the json that the Circle CI fetch build endpoint returns, in case the build is not in the context of a PR.

Previously, it would crash in this case, but now it will report that current build has no PR, and Danger will not run, as intended.

Related issue: #1135

My Ruby skills are pretty poor, so I hope this PR conforms to coding guidelines. It's my first contribution to Danger, I'd be happy to incorporate any suggestions you have for this PR.


orta:
Nah, no changes needed - this looks good! Thanks, will ship it when another PR comes in

Issue #1141, opened 2019-07-09
Get danger_id inside Dangerfile


rpassis:
Is there a way to retrieve the danger_id from inside the Dangerfile? 

I have a single generic Dangerfile that loads other Dangerfiles. What I’d like to do is set my danger_id to `pre` or `post` and use that as the prefix for the filename to be loaded ie `Dangerfile.pre` or `Dangerfile.post`.

Is this an ok approach to this?

orta:
Not in the public API, but its ruby so you might be able to find it somewhere - I'd follow from the CLI down and see if any of the plugins get access to it, then see if you can get it from them

rpassis:
Thanks! Does it make sense to have this exposed publicly and if so, is there in interest in a PR?

orta:
Seems reasonable to me, yeah

PR #1142, opened 2019-07-10
Exposes danger_id via EnvironmentManager

rpassis:
This PR addresses issue #1141 by exposing the value of danger_id via `EnvironmentManager`.

To access the value from inside a Dangerfile you can use `danger.env.danger_id`.

Unit tests have also been added to ensure the value passed is being set, and when none is provided that the default danger is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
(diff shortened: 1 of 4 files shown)


orta:
Cool, yeah 👍🏾 

It's been a few years since I've used danger on ruby, are you sure `danger.env` is exposed to dangerfiles?

rpassis:
Yup this is it here https://github.com/danger/danger/blob/master/lib/danger/danger_core/dangerfile.rb#L20

orta:
Rockin' thanks!

PR #1149, opened 2019-08-05
Adds complex gitlab.api example

H3npi:
This commit adds another example to [https://danger.systems/reference.html](https://danger.systems/reference.html).

Currently there is no example for the gitlab.api method. 
This is solved with this example.


orta:
Thank you!

PR #1165, opened 2019-10-04
Updated Semaphore to support 2.0

gabrielrinaldi:
Updated the Semaphore CI strategy to version 2.0.

This is a breaking change, so I am not sure how to proceed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 * Add an example for gitlab.api on [reference](https://danger.systems/reference.html)
 * Fix `html_links` of `dangerfile_gitlab_plugin` for non `gitlab`/`jenkins` ci [@mfiebig](https://github.com/mfiebig) [#1157](https://github.com/danger/danger/pull/1157)
-* Updated Semaphore to work with Semaphore 2.0. **this is a breaking change** [@gabrielrinaldi](https://github.com/gabrielrinaldi) [#1165](https://github.com/danger/danger/pull/1165)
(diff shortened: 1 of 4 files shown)


orta:
I don't know much about semaphore - is v2 opt in?

orta:
either way - this doesn't need to be a breaking change if it still supports the old environment variables as well. Maybe just do that anyway to be safe and then updating requires no thinking?

gabrielrinaldi:
@orta what do you think? Should be easy to remove original Semaphore once they retire it.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 * Add an example for gitlab.api on [reference](https://danger.systems/reference.html)
 * Fix `html_links` of `dangerfile_gitlab_plugin` for non `gitlab`/`jenkins` ci [@mfiebig](https://github.com/mfiebig) [#1157](https://github.com/danger/danger/pull/1157)
+* Updated Semaphore to work with Semaphore 2.0. **this is a breaking change** [@gabrielrinaldi](https://github.com/gabrielrinaldi) [#1165](https://github.com/danger/danger/pull/1165)
```
orta:
```suggestion
* Updated Semaphore to also work with Semaphore 2.0. [@gabrielrinaldi](https://github.com/gabrielrinaldi) [#1165](https://github.com/danger/danger/pull/1165)
```

orta: Get this in and we're good to go - thanks, so try it took a while for me to get back to this

gabrielrinaldi:
@orta thanks, accepted the changes

PR #1166, opened 2019-10-07
Remove legacy gh actions meta & hopefully fix docker

RDIL:


orta:
Looks like there's been some pretty drastic changes to the docker file with https://github.com/danger/danger/pull/1168

We ok to have that instead?

Issue #1148, opened 2019-08-04
Simple Dangerfile fails when run through GitHub Actions


Killectro:
# Report

👋

I have a simple Dangerfile that prints a message, and a `main.workflow` that points to the `danger/danger@master` Dockerfile, and when the action is run the workflow fails with the below error message.

Running Danger locally works completely as expected, but running through Actions fails.

----
## What did you do?

Attempt to set up Danger in GitHub Actions to print a message.

## What did you expect to happen?

I see a `message` printed.

## What happened instead?

I received this error message:

```### STARTED Danger 14:19:29Z

Pulling image: gcr.io/github-actions-images/action-runner:latest
latest: Pulling from github-actions-images/action-runner
169185f82c45: Pulling fs layer
0ccde4b6b241: Pulling fs layer
d0372f57daa2: Pulling fs layer
165911d108d6: Pulling fs layer
54996bce1de5: Pulling fs layer
165911d108d6: Waiting
54996bce1de5: Waiting
d0372f57daa2: Verifying Checksum
d0372f57daa2: Download complete
0ccde4b6b241: Verifying Checksum
0ccde4b6b241: Download complete
169185f82c45: Verifying Checksum
169185f82c45: Download complete
165911d108d6: Verifying Checksum
165911d108d6: Download complete
54996bce1de5: Verifying Checksum
54996bce1de5: Download complete
169185f82c45: Pull complete
0ccde4b6b241: Pull complete
d0372f57daa2: Pull complete
165911d108d6: Pull complete
54996bce1de5: Pull complete
Digest: sha256:c9bb432ec5ec08ee08b040a9fccacebbbf8a91444dac4721600cf5dca9dae57e
Status: Downloaded newer image for gcr.io/github-actions-images/action-runner:latest
Step 1/11 : FROM ruby:2.5
2.5: Pulling from library/ruby
5ae19949497e: Already exists
ed3d96a2798e: Already exists
f12136850781: Already exists
1a9ad5d5550b: Already exists
6f18049a0455: Already exists
d1e86a347d84: Already exists
60032ada6609: Pulling fs layer
6152433ccd3b: Pulling fs layer
6152433ccd3b: Verifying Checksum
6152433ccd3b: Download complete
60032ada6609: Verifying Checksum
60032ada6609: Download complete
60032ada6609: Pull complete
6152433ccd3b: Pull complete
[...]

PR #1168, opened 2019-10-11
Use COPY instead of ADD, ENTRYPOINT instead of RUN, and use relative path in Github actions docker

sergioisidoro:
This attempts to fix the issue:

https://github.com/danger/danger/issues/1148

Of the code:
```
ADD ./Gemfile $APP_HOME/Gemfile
ADD Gemfile.lock $APP_HOME/Gemfile.lock
```

The first line goes smooth, but the second fails, which seems to confirm this comment: https://github.com/danger/danger/issues/1148#issuecomment-518532561 

This adds a relative path to the ADD command in the dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN mkdir $APP_HOME
 WORKDIR $APP_HOME
 
 ADD ./Gemfile $APP_HOME/Gemfile
-ADD Gemfile.lock $APP_HOME/Gemfile.lock
+ADD ./Gemfile.lock $APP_HOME/Gemfile.lock
 
 RUN bundle install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,8 @@ ENV APP_HOME /myapp
 RUN mkdir $APP_HOME
 WORKDIR $APP_HOME
 
-ADD ./Gemfile $APP_HOME/Gemfile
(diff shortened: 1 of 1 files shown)


sergioisidoro:
Ok, doesn't seem to fix the problem..
```
2019-10-11T12:45:47.2454722Z Step 8/11 : ADD ./Gemfile.lock $APP_HOME/Gemfile.lock
2019-10-11T12:45:47.2517295Z ADD failed: stat /var/lib/docker/tmp/docker-builder466665871/Gemfile.lock: no such file or directory
```

sergioisidoro:
Ok, I think I got it... I was looking at Danger-js and everything works. I started to change things to match the Dockerfile in Danger-js Dockerfile, and arrived to this: 

`COPY . /app`

Apparently Github might be doing some magic to start building containers and before checkout unless the dockerfile contains `COPY . ` ??? 

I now have another problem, which is
`Bundler::LockfileError: You must use Bundler 2 or greater with this lockfile.`

But that's another thing, probably coming form the base image.

sergioisidoro:
Installs bundler becaue of https://github.com/docker-library/ruby/pull/255#issuecomment-451007457

orta:
Cool, yeah, looks good to me 👍

Issue #985, opened 2018-05-14
Bamboo integration


michal-skalski:
Do you plan to add support for Bamboo? If not, could you point me direction so I could achieve that?

orta:
No plans, you're welcome to add it though - a PR adding a new CI platform looks like this: https://github.com/danger/danger/pull/983

Qbeorama:
Hi there, I made a PR with Bamboo support for Danger: #1172 .

PR #1172, opened 2019-11-17
#985 Add Bamboo support

Qbeorama:
Hi there,

This PR adds support for Bamboo CI to Danger. 
Currently, it is only integrated with Bitbucket Server, as this is what I could test it against.


orta:
Oops, I completely missed this! 

Looks like what I'd expect, let's get this in

Issue #1177, opened 2020-01-12
Run danger from subdirectory


scranen:
I'd basically like to re-open #219, for the following reason. We are trying to set up a build in a mono repo, where each project might have its own Dangerfile, or even no Dangerfile at all. We'd like each Dangerfile to be able to assume that it is in the root of the project, so in the following diagram, we want to run `danger` from `./project1`, rather than from `.`.
```
.
├── .git
├── project1
│     ├── Dangerfile
│     ┆
│     └── ...
└── project2
      ├── Dangerfile
      ┆
      └── ...
```

**Current behaviour**
Running `danger` in a subdirectory of a git repository results in an error like the following:
```
/path/to/repo/project1/.git: path does not exist (ArgumentError)
```

**Desired behaviour**
Running `danger` in a subdirectory of a git repository just looks up the root of the git repository itself.

**Suggested fix**
It seems that the fix suggested in #219 actually does work: changing
```diff
     def diff_for_folder(folder, from: "master", to: "HEAD")
       self.folder = folder
-      repo = Git.open self.folder
+      repo = Git.open exec("rev-parse --show-toplevel")

       ensure_commitish_exists!(from)
       ensure_commitish_exists!(to)

       merge_base = find_merge_base(repo, from, to)
       commits_in_branch_count = commits_in_branch_count(from, to)

-      self.diff = repo.diff(merge_base, to)
+      self.diff = repo.diff(merge_base, to).path(folder)
       self.log = repo.log(commits_in_branch_count).between(from, to)
     end
```
in `/lib/danger/scm_source/git_repo.rb` seems to do the trick.

orta:
I feel slightly wary that this could break someone's setup if they had already introduced danger using a pattern like the original fix of `--dangerfile` - so if it can be done behind a flag somehow, I'm game 👍

scranen:
I've limited the scope of the change to the only place that I think is currently triggering the issue by making the lookup behaviour explicit in the interface of `GitRepo`. I added tests for `GitRepo`, but would appreciate if someone could help me to get a test in place for actually calling danger itself from a subdirectory. (Current test shows that `GitRepo` can deal with being called from a subfolder, but not that calling `danger` from a subfolder would work as expected)

scranen:
I also changed my mind on the additional `.path(folder)` in the diff I posted previously. On second thought I think this filtering should be left up to the user if they want it.

scranen:
[...]

PR #1178, opened 2020-01-15
Allow Danger to be run from a subfolder in a git repository

scranen:
Addresses #1177. The idea is to enable Danger to be run from a subdirectory within a git repository without it complaining that the current path is not a git repository. Instead, it should do an educated guess about what we intended to do, and look up the top level git directory (by calling git).

To ensure backward compatibility, we only do this lookup in the special case that is currently causing Danger to fail (when doing the initial diff on `.`), but the functionality is exposed in `GitRepo` so is in principle reusable.

The two added tests serve to illustrate the new behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## master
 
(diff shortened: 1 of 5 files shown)


orta:
Hah, I missed this - doh, sorry.

Yeah, this looks perfect, let's get this shipped.

PR #1184, opened 2020-02-07
Fix runtime errors in BitbucketCloud#delete_old_comment

telyn:
I had renamed BitbucketCloudAPI#fetch_last_comments, but neglected to refactor its use in BitbucketCloud. And I also broke the while loop I wrote, despite having had it working at some point. A touch embarrassing - sorry for the extra work! I should've waited until today to have a quick re-test and then submit my PR, rather than pushing ahead at 9PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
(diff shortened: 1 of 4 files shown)


orta:
👍 will ship a release

Issue #1183, opened 2020-02-07
Message aggregation


telyn:
This is a problem I've noticed with the bitbucket-cloud request source, at least. I believe it might already be implemented in the Github request source, but it's actually a general feature that should be refactored out.

The problem is: Multiple messages for the same line of a file go in multiple comments, rather than in one comment that's a simple list.

![Screenshot from 2020-02-07 11-05-05](https://user-images.githubusercontent.com/175827/74024405-b690fa80-4999-11ea-9a2a-9825f9be5ab2.png)

I propose we add a `type` field to `Violation` (with `:error`, `:warning`, or `:message`), and a `type` method to `Markdown`, which always returns `:markdown`.

Then add a `MessageAggregator` which would take the warnings, errors, messages and markdowns lists, and produce just one list of `MessageGroup`s:

**MessageGroup**
| Field | Description |
| ----- | ----- |
| file | Possibly-nil filename |
| line | Possibly-nil line number |
| messages | array of `Markdown` & `Violations` |

The list of messages should be sorted, errors at the top, followed by warnings, then messages, then markdowns.
Ultimately perhaps `line` should be a range, and `MessageGroup` should have a `merge` method - to combine the messages arrays & `line` ranges as long as the `file` is equal, but I'm keen to not over-design this at this point. That'd allow a potentially-customisable minimum-distance-between-

`Danger::Dangerfile#post_results` would have two code paths: one to use `MessageAggregator` and a new method on `request_source` (say, `update_pr_message_groups!`), and one to use the old violation_report and `update_pull_request!`, all based on whether the request support has a method `message_groups?` that returns positive.

Longer-term, I see `MessageGroup` and `Violation` taking over a lot of the logic in the `GitHub` class relating to figuring out whether violations can be ticked-off, and request sources having a new method for fetching the comments as `MessageGroup`s. All the markdowns can be chucked away, as there's no way to give them a white tick anyway.

If that doesn't sound like a total load of arse then I'll write some code sometime Feb/March :)

orta:
This sounds good - yep!

Kaspik:
@telyn Hey! Are you still planning to add it for Github too? We have currently 7 Danger files on PR, and having 7 PR comments by the same "bot" user is really annoying especially with combination of posting PR comments to Slack channel.

Thanks!

telyn:
[...]

PR #1186, opened 2020-02-14
Foundation code for Message Groups

telyn:
Hey again :)

I got started working on #1183 and I thought I'd get this PR made sooner rather than later so you can take a sneak peek if you want - I've got a decent amount of the foundational work done towards it without (I believe) causing any breakage.

**Mostly** though I want to ask the question: How would you like to organise getting my changes PR'd into danger? All in one massive PR when I'm all done, or piece-by-piece?

The summary of the changes I've made are that:
* The `MessageAggregator` class has been added without any methods.
* `MessageAggregator.aggregate` (which does not exist) will be called from `Dangerfile#post_results`, and it will be created if a request_source has a `#update_pr_by_line!` method (which none currently do) (I am very happy to change `#update_pr_by_line!`'s name - it's the best I could come up with at the time!)
* The `MessageGroup` type has been added and has a fairly decent sketch of its methods.
* Violation now has a `#type`, which defaults to `:warning`, can be set from calls to `.new`, and all calls to `.new` inside this codebase pass it - so `fail`,`warn`, `message` all create `Violations` with the correct `type#`
* Markdown now also has `#type` - which returns `:markdown`.
* Violation and Markdown now have a thoroughly-tested `#<=>` method each. in preparation for the sorting code being added to `MessageGroup`.
* Violation and Markdown now inherit from a base type to keep the code a bit DRY-er. In retrospect perhaps the base class should've had a list of types and their orderings - but I only made the base after I'd started on the implementation for sorting them both 😅

A changelog to that effect will be along shortly :)

P.s. is it cool if I add [Fuubar](https://github.com/thekompanee/fuubar) and fiddle with the guardfile so it doesn't run rubocop if the rspec tests fail?


danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a CHANGELOG entry. <br />
You can find it at <a href="https://github.com/danger/danger/blob/master/CHANGELOG.md">CHANGELOG.md</a>.</td>
    </tr>
  </tbody>
</table>

</p>

CHANGELOG.md
```diff
@@ -9,17 +9,21 @@
...
+* Fix incorrect method call in Bitbucket Cloud arising from my own changes in 6.2.1 [@telyn][@telyn] [#1184](https://github.com/danger/danger/pull/1184)
 
 ## 6.2.1
 
 * Fix Bitbucket Pipelines documentation
 * Update Ruby-Git to 1.6.0 to fix output encoding
-* Add inline-comments support to Bitbucket Cloud [@telyn](https://github.com/telyn) [#1182](https://github.com/danger/danger/pull/1182)
-* Make emojis display better on Bitbucket Cloud [@telyn](https://github.com/telyn) [#1182](https://github.com/danger/danger/pull/1182)
+* Add inline-comments support to Bitbucket Cloud [@telyn][@telyn] [#1182](https://github.com/danger/danger/pull/1182)
+* Make emojis display better on Bitbucket Cloud [@telyn][@telyn] [#1182](https://github.com/danger/danger/pull/1182)
```
orta: If you leave this blank, then markdown will assume it is the same as the text on the left -`[@telyn][]` === `[@telyn][@telyn]`

.rubocop.yml
```diff
@@ -1,7 +1,7 @@
 # Defaults can be found here: https://github.com/bbatsov/rubocop/blob/master/config/default.yml
 
 AllCops:
-  TargetRubyVersion: 2.0
+  TargetRubyVersion: 2.3
```
orta: Seems legit

CHANGELOG.md
telyn: oh, neat, didn't know that 😁

.rubocop.yml
telyn: I needed it to stop rubocop complaining, and the gemspec says 2.3 is the minimum supported version anyway 😉

orta:
I'm happy for you to add deverloper-y dependencies, I only get cagey about dependencies which danger consumers end up with.

WRT to how you want to get it in, I prefer that master is always deployable - this looks backwards compat, but if it weren't we could just scope it like:

```ruby
if env.request_source.respond_to?(:update_pr_by_line!) && process.env.USE_NEW_THING
``` 

So, feels good to me

telyn:
Cool - I'm gonna deal with your suggestion on the markdown file, dogfood it more at work to ensure that I've not broken anything, then un-WIP this :)

PR #1200, opened 2020-03-08
Update RegEx pattern for Bitrise projects

RonaldPrithiv:
This should fix the URL matching issue for good. 

Instead of hardcoding `.com` which turned out to be a bad choice, I am using the previously used regex pattern `(%r{([\/:])(([^\/]+\/)+[^\/]+?)(\.git$|$)})` which had worked fine except in the case when the URL contains `https://` as `://` leads to inaccurate matching. Now I added a check if the URL contains `https://` if it does, then we just remove it and proceed to matching


lib/danger/ci_source/bitrise.rb
```diff
@@ -49,14 +49,19 @@ def initialize(env)
...
       self.repo_url = env["GIT_REPOSITORY_URL"]
   
-      if repo_url.include? ".com/"
-        repo_matches = self.repo_url.match(%r{\.com/(.*)})[1].split(/\.git$|$/)[0]
-      elsif repo_url.include? ".com:"
-        repo_matches = self.repo_url.match(%r{\.com:(.*)})[1].split(/\.git$|$/)[0]
+      if repo_url.include? "https://"
+        repo_url["https://"] = ''
+        
+        repo_matches = self.repo_url.match(%r{([\/:])(([^\/]+\/)+[^\/]+?)(\.git$|$)})[2]
```
notjosh: Can you add a comment explaining this being here (even just copy/paste the PR description)? This line matches the one in the `else` block, and would be easy to (mistakenly) refactor it out.

Though, it might be a bit simpler to actually refactor it slightly to something a bit like:

```ruby
matcher_url = self.repo_url

# https URLs mess up the regex, so we remove the protocol part
if repo_url.include? "https://"
  matcher_url["https://"] = ''
end

repo_matches = matcher_url.match(%r{([\/:])(([^\/]+\/)+[^\/]+?)(\.git$|$)})[2]

self.repo_slug = repo_matches unless repo_matches.nil?
```

Thoughts?

RonaldPrithiv: hahaha! I don't know why I didn't use a variable. Yes this is perfect. 

Thank you.

Issue #1215, opened 2020-04-15
Git: Info for added and deleted files.


ethan-riback:
# Report

Note: This issue was brought up here: https://github.com/danger/danger-js/issues/368

## What did you do?

```ruby
files = git.added_files + git.modified_files + git.deleted_files

xib_modified_line_count = 0
files.grep(/.*\.xib/).each do |file|
  info = git.info_for_file(file)
  xib_modified_line_count = xib_modified_line_count + info[:insertions] + info[:deletions]
end
message("xib line changes: #{xib_modified_line_count}.")
```

## What did you expect to happen?

Properly receive `info_for_file` for `added_files` and `deleted_files`. Calling `info_for_file` on an added file would cause insertions to be the file's line count and deletions to be 0. Calling `info_for_file` on a deleted file would cause insertions to be 0 and deletions to be the file's line count.

## What happened instead?

Received the following error, since info_for_file is returning `nil` for `added_files` and `deleted_files`.
```
Danger has errored
[!] Invalid Dangerfile file: undefined method [] for nil:NilClass. Updating the Danger gem might fix the issue. Your Danger version: 5.16.1, latest Danger version: 6.3.2

 #  from Dangerfile:14
 #  from Dangerfile:11
 #  -------------------------------------------
 #    info = git.info_for_file(file)
 >    xib_modified_line_count = xib_modified_line_count + info[:insertions] + info[:deletions]
 >    xib_modified_line_count = xib_modified_line_count + info[:insertions]
 #  end
 #  -------------------------------------------
```

## Your Environment

* Which CI are you running on?
BuddyBuild
* Are you running the latest version of Danger?
No, running on 5.16.1.
* What is your Dangerfile?

  ```ruby
  # Declare a PR to be simple to avoid specific Danger rules
  declared_trivial = (github.pr_title + github.pr_body).include?("#trivial")
  # Make it more obvious that a PR is a work in progress and shouldn't be merged yet
  fail("PR is classed as Work in Progress") if github.pr_title.include?("[WIP]") || github.pr_labels.include?("WIP")
  
  # Make sure the PR does not have too many lines modified, not including the project file or xib files
  
  files = git.added_files + git.modified_files + git.deleted_files
  
  xib_modified_line_count = 0
  files.grep(/.*\.xib/).each do |file|
    info = git.info_for_file(file)
    xib_modified_line_count = xib_modified_line_count + info[:insertions] + info[:deletions]
  end
[...]

PR #1216, opened 2020-04-15
Added support to git.info_for_file for added and deleted files.

ethan-riback:
Problem: As a developer I needed a way to check the link length of added and deleted files.

Solution: Update [`info_for_file`](https://github.com/danger/danger/blob/master/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb#L138) to check against `added_files` and `deleted_files`. If the file was added, insertions would be the file size and deletions would be 0. If the file was deleted, insertions would be 0 and deletions would be the file size.

This pull request addresses issue https://github.com/danger/danger/issues/1215. Strictly having been a mobile developer my professional life and having rarely touched Ruby, I had no idea if I could address this on my own. But when my issue was labeled "You Can Do This", I thought "you know what, I can"! So thanks for the motivation! I hope this PR covers what is needed.

Note: I was unsure if we should define a value for `before:` or `after:`. `diff.blob(:src).contents` and `diff.blob(:dst)` both error when using added or deleted files so for now these fields are made `nil` under the new circumstances.

### DIFF
--- a/spec/lib/danger/danger_core/plugins/dangerfile_git_plugin_spec.rb
+++ b/spec/lib/danger/danger_core/plugins/dangerfile_git_plugin_spec.rb
@@ -11,8 +11,10 @@ def run_in_repo_with_diff
       `git tag 1.0.0`
       `git checkout -b new-branch --quiet`
       File.open(dir + "/file2", "w") { |f| f.write "Pants!" }
+      File.open(dir + "/file3", "w") { |f| f.write "Jawns.\nYou know, jawns." }
+      File.delete(dir + "/file1")
       `git add .`
-      `git commit -m "update file2"`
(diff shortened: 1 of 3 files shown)


spec/lib/danger/danger_core/plugins/dangerfile_git_plugin_spec.rb
```diff
@@ -11,8 +11,10 @@ def run_in_repo_with_diff
       `git tag 1.0.0`
       `git checkout -b new-branch --quiet`
       File.open(dir + "/file2", "w") { |f| f.write "Pants!" }
+      File.open(dir + "/file3", "w") { |f| f.write "Jawns.\nYou know, jawns." }
```
orta, line 14: I'm into this

orta:
Yep, thanks!

PR #1221, opened 2020-04-28
Fix import Dangerfile in directory from GitLab

adzeitor:
According to [GitLab Repositories API docs](https://docs.gitlab.com/ee/api/repositories.html) filepath and project path should be URL encoded.

Before this change all slashes in filepath and project path should be replaced with '%2F'.


danger-public:
## Danger has errored
[!] Invalid `Dangerfile` file: undefined method `nodes` for nil:NilClass
```
 #  from Dangerfile:59
 #  -------------------------------------------
 #  unless ENV["RUNNING_IN_ACTIONS"]
 >    junit.parse "junit-results.xml"
 #    junit.headers = %i(file name)
 #  -------------------------------------------
```

</p>

adzeitor:
To make sure that API works this way we can use public API:

This works (encoded):

https://gitlab.com/api/v4/projects/adzeitor%2Fruby/repository/files/2.5.5%2FDockerfile/raw?ref=master

But this don't:

https://gitlab.com/api/v4/projects/adzeitor/ruby/repository/files/2.5.5/Dockerfile/raw?ref=master

orta:
Great! thanks

PR #1226, opened 2020-05-06
[Infra] Add Ruby 2.7 tests

Kaspik:
- add new ruby 2.7.1 tests to find out if danger works correctly on this ruby version
- fix changelog link


Kaspik:
A lot of warnings about deprecated stuff, but tests work which is good sign. Some of the warnings can be fixed by updating dependencies, will try to do that in another PR.

orta:
Cool 👍

PR #1227, opened 2020-05-06
Bump dependencies in gemspec

Kaspik:
- Bump gemspec dependencies to new versions to silent Ruby 2.7 warnings on outdated dependencies

This might break couple of things, so opening draft PR first to see what happens.


orta:
This just changes the gemfile, and shouldn't affect consumers of danger - looks good to me

PR #1269, opened 2020-10-25
Fix GitHub actions issue with pull_request_target

Chesire:
Using GitHub actions if the event name is `pull_request_target` instead of `pull_request`, then Danger will not detect it as a valid PR and the error ```Not a GitHubActions Pull Request - skipping `danger` run.``` will be displayed when trying to run it.
Updated the check in the `github_actions` file to also check for this event name as well.

Based on https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/ this should be almost identical so hopefully the change will work as intended, I could not see anything related on how to test a GitHub action though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 ## master
 <!-- Your comment below here -->
 
-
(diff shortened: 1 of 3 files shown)


orta:
Agree 👍🏻

PR #1270, opened 2020-10-28
Fix Bitbucket Cloud not updating comments

tskulbru:
This PR should fix the issue that we are currently seeing with Danger not being able to update an already existing comment on Bitbucket Cloud, or deleting existing ones for that matter.

The problem stems from Bitbucket changing their API's away from using `username`s but instead using `uuid` and `nickname` Sometime ago they apparently removed the username field in their api responses which caused the existing filters / query params to fail. By replacing username with uuid in everything related to comments on Bitbucket Cloud, things are working as intended again.


orta:
Cool, this makes sense -thanks

PR #1274, opened 2020-11-04
Fix incorrect command in error message

revolter:
Running `danger local --pry` was printing out

```
[!] Pry was not found, and is required for 'danger pr --pry'.
```

instead of

```
[!] Pry was not found, and is required for 'danger local --pry'.
```

which was confusing.


orta:
Thanks, cool  - looks good to me 👍🏻

Issue #1245, opened 2020-07-14
Remove direct Hash method calls for gitlab client


mrosePTC:
# Report
Since NARKOZ is deprecating the direct usage of Hash methods in the `gitlab` gem. The direct calls to hash methods on `ObjectifiedHash` should be converted to invoke `.to_hash` on them before calling the hash methods.

## What did you do?
`danger --fail-on-errors=true` on a GitLab merge request

## What did you expect to happen?
I expected the output to not contain these warnings:
```
 WARNING: Please convert ObjectifiedHash object to hash before calling Hash methods on it.
```

## What happened instead?
I got those warnings

## Your Environment

* Which CI are you running on?
gitlab-ci

* Are you running the latest version of Danger?
yes

* What is your Dangerfile?

```ruby

warn("MR is classed as Work in Progress") if gitlab.mr_title.include?("WIP")
failure("MR doesn't start with story id") unless gitlab.mr_title.match?(/.*PREFIX-[0-9]+/)

def verify_commit(commit)
  bad_commit = !commit.message.match?(/^[A-Za-z]+\(.+\): \w+/)
  failure("Commit message isn't of the correct format 'foo(bar): baz' #{commit.sha}") if bad_commit

  questionable = commit.message.match?(/^(feat|fix)[^\(]/)
  warn("Prefix must match 'fix(' or 'feat(' to be in release notes #{commit.sha}") if questionable

  missing_story_id = !commit.message.match?(/^[A-Za-z]+\(DT-[0-9]+\)/)
  warn("Commit message does not have a story id 'foo(DT-123): baz' #{commit.sha}") if missing_story_id
end

if gitlab.mr_json["squash"]
  # Make sure first commit follows correct formatting
  verify_commit(git.commits.first)
else
  # Make sure all git commits follow the correct formatting
  git.commits.each do |commit|
    verify_commit(commit)
  end
end
```

PR #1278, opened 2020-11-23
Update to silence the objectified hash warnings

dstull:
- since notes were calling merge, they are an ObjectifiedHash
  and need to call to_h first as seen in the following warning:

WARNING: Please convert ObjectifiedHash object to hash before calling Hash methods on it.

which comes from https://github.com/NARKOZ/gitlab/blob/3dd41722670570069693cc9b84c73f475d95e6be/lib/gitlab/objectified_hash.rb#L40

Resolves https://github.com/danger/danger/issues/1245

### DIFF
--- a/spec/lib/danger/request_sources/gitlab_spec.rb
+++ b/spec/lib/danger/request_sources/gitlab_spec.rb
@@ -362,7 +362,7 @@ RSpec.describe Danger::RequestSources::GitLab, host: :gitlab do
               :headers => {'Accept'=>'application/json', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Content-Type'=>'application/x-www-form-urlencoded', 'Private-Token'=>'a86e56d46ac78b', 'User-Agent'=>'Gitlab Ruby Gem 4.16.1'}).
             to_return(:status => 200, :body => "", :headers => {})
 
-          
+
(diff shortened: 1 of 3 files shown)


dstull:
Hi @orta can you give this one a review?

spec/lib/danger/request_sources/gitlab_spec.rb
```diff
@@ -400,7 +400,7 @@
 
           subject.fetch_details
 
-          subject.update_pull_request!(warnings: [v], errors: [], messages: [])
+          expect { subject.update_pull_request!(warnings: [v], errors: [], messages: []) }.not_to output(/Please convert ObjectifiedHash object/).to_stderr
```
dstull, line 403: meh - not ideal to test here, but there weren't specific tests for the methods this one hits, so this will ensure it doesn't happen again...at least in this specific call

orta: heh, seems reasonable enough

orta:
👍🏻

orta:
Thanks!

dstull:
Hi @orta, thanks for merging.  Do you know when a release/new gem version will be made with this change?

orta:
Usually after 2-3 PRs are merged, given that this just fixes a warning, it's not too pressing for me to go through the deploy process

PR #1225, opened 2020-05-06
Fix strike-out comments on Github for Danger 7+, drop Ruby 2.3

Kaspik:
- Fixes: https://github.com/danger/danger/issues/1189
- How: Re apply change from https://github.com/danger/danger/pull/1208 as Danger 7.0.0 doesn't have it anymore somehow
- Drops Ruby 2.3 - Circle CI dropped this version support last week and tests are not working anymore - it's dangerous to guarantee support for version that we don't have tests for, so I believe we should drop it. It's in EOL state for years anyways.

If you check the file history, the PR is not even there and the commit never made it to Danger 7.0 even tho other parts of the PR are correctly on master. This should fix the issue again for Danger 7+.

@orta I have no idea how this happened, Git history doesn't show anything. No idea how Danger 7 was released either... but the only way how to fix this is downgrade to Danger 6.3.2. Could we release 7.0.2 with this? Or do you want 7.1 as this also drops Ruby to keep tests green?

PR #1292, opened 2021-02-19
Allow terminal-table versions through 3.x

benasher44:
As of Jekyll 4.2.0, Jekyll requires terminal-table 2.0 (my motivation here is to be able to upgrade my Jekyll version). Looking at their release notes, the breaking change for terminal-table 2 is requiring ruby 2.4+, and it appears danger dropped 2.3 support in #1225.

Side note: I had unfortunately not pinned danger properly when I upgraded to Jekyll 4.2.0, so bundler rolled my danger all the way back to 0.3.0, which is presumably the most recent version of danger that didn't use terminal-table. Anyway, that's a lesson in pinning versions properly for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 ## master
 <!-- Your comment below here -->
 
+* Allow teriminal-table versions through 3.x - [@benasher44](https://github.com/benasher44)
+
 <!-- Your comment above here -->
 
 ## 8.2.2
--- a/danger.gemspec
+++ b/danger.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "kramdown", "~> 2.3"
   spec.add_runtime_dependency "kramdown-parser-gfm", "~> 1.0"
   spec.add_runtime_dependency "octokit", "~> 4.7"
-  spec.add_runtime_dependency "terminal-table", "~> 1"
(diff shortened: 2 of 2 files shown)


benasher44:
I think this change is okay from danger's perspective, but this may push other consumers to open similar PRs/issues on other projects that use danger but have other dependencies that aren't ready for terminal-table 2. Alternatively, we could go with a slightly more forgiving pin like `">= 1", "< 3"`. Assuming terminal-table is a good citizen of semver, this shouldn't be an issue, as terminal-table didn't use 2.0 as an opportunity to make other breaking API changes.

danger-public:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">External contributor has edited the Gemspec</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">.gemspec modified</td>
    </tr>
  </tbody>
</table>

</p>

orta:
Given v3 is also out, and has no API breaking changes, I think the more forgiving `">= 1", "< 4"` is probably a safe bet 👍🏻

benasher44:
Ah! Missed that version. Done!

orta:
Will ship it when some other PRs land, deploying gems requires me going to another computer 👍🏻

PR #1294, opened 2021-03-09
Namespace the Commits class

rymai:
We have a class name clash in GitLab where we also have a non-namespaced `Commits` class.

Namespacing the `Commits` class here seems like the right change, given that gems should always namespace their classes and that other classes in the same folder are already namespaced.


orta:
Yeah, looks good to me 👍🏻 

The travis fail is acceptable

PR #1295, opened 2021-03-16
Use html comment to identify comments from danger

Ram-Z:
Avoid adding the "Generated by :no_entry_sign: " footer in inline comments even when there is markdown.

This is using a html comment instead of a `data-meta` attribute.

Arguable the `data-meta` from the table above could be remove if this is printed always, or the html comment could be made to only print when there is markdown or no table, suggestions are welcome.


Ram-Z:
@orta Any thoughts on this MR?

orta:
Yeah, this seems good to me 👍🏻

orta:
I don't have time to release this though, so will just merge

Ram-Z:
Thanks.

PR #1297, opened 2021-03-30
update kramdow package because of high severity vulnerability

reshadf:
Fixing below error. 

CVE-2021-28834

high severity
Vulnerable versions: >= 1.16.0, < 2.3.1
Patched version: 2.3.1
Kramdown before 2.3.1 does not restrict Rouge formatters to the Rouge::Formatters namespace, and thus arbitrary classes can be instantiated.


orta:
This PR does not do what you think it does: https://depfu.com/blog/2016/12/14/get-to-know-your-twiddle-wakka

I'd be open to it being: `spec.add_runtime_dependency "kramdown", "~> 2.3", "> 2.3.1" ` though

RomanMinenok:
@orta I noticed the PR was never merged. Why so? Is there some alternative fix? I had to delete Danger from my repo until this is fixed. 

Even though 2.3.1 should technically be fetched instead of 2.3.0 I won't be able to convince my manager in that until huge red box saying that our repo has security vulnerability stays in our Github.

orta:
This doesn't fix the issue, and I gave you a command to fix it in https://github.com/danger/danger/issues/1298#issuecomment-811143873

reshadf:
@orta I changed it according to your suggestion.

orta:
It's been a few years since I've written Ruby, so judging on the CI results that maybe that syntax might need tweaking a bit to get right 👍🏻

reshadf:
I think it only works for < I think ~> would be sufficient or people should `bundle update kramdow` and we close this PR all together.

RomanMinenok:
@orta Thanks, I'll try that out, but the moment I tried to fix this with bundle update Jenkins started to fail miserably. I guess I have to give a try one more time. 

If this doesn't work when the release with this fix is planned then?

orta:
@RomanMinenok learn how Ruby dependency management works and understand the tooling, it's not on me to hold your hand in my spare time. Danger doesn't _need_ a release for the kramdown issue, it'd just be a nice to have. While the CI is red for these changes, we can't make a release of Danger, so when this PR is merged I'll ship a release

orta:
Alright, I'm going to close this up - thanks for giving it a shot @reshadf

PR #1305, opened 2021-05-20
Handle GitLab URL containing /-/ scope on Jenkins

adamprice:
GitLab now uses a `/-/` scope in its URLs as documented [here](https://gitlab.com/gitlab-org/gitlab/-/issues/214217).

The regex in the Jenkins CI class was incorrectly parsing the project URL as `https://<gitlab_host>/<project>/-/` when the URL contained the `/-/` scope, which was resulting in a 404 when trying to access the GitLab project API.

Sorry if I've missed anything, this is my first time contributing 🙂.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## master
 <!-- Your comment below here -->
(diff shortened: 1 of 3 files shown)


orta:
👍🏻 cool, everything looks good to me

PR #1306, opened 2021-05-21
Add pr_draft? method to GitHub DSL

### LINKED ISSUE #1150: Add DSL attribute for draft PR status
Is there any planned support for draft PRs, so we'd be able to show a warning if the github PR is in draft mode, for example?

connorshea:
This checks the mergeable state of the PR JSON to determine whether a PR is a draft. This doesn't add a real proper test for the method because it didn't seem like the other methods (pr_author, pr_labels, etc.) had dedicated tests, so I figured this didn't need one.

I can add a specific test with a fixture if we want that, although I may need a bit of guidance on how to get the fixture to use. :)

Resolves #1150.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## master
 <!-- Your comment below here -->
 * Handle GitLab URL containing /-/ scope on Jenkins - [@adamprice](https://github.com/adamprice)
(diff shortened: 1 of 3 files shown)


orta:
Looks good to me 👍🏻

PR #1312, opened 2021-07-22
Test with Ruby 3.0

mataku:
Since Danger also supports Ruby 3, add Ruby 3 to CI workflow to make it easier to notice problems. 

In the Ruby setup, actions/setup-ruby was deprecated and replaced with ruby/setup-ruby: https://github.com/actions/setup-ruby#setup-ruby.


orta:
Thanks, looks good to me

Issue #1173, opened 2019-11-18
AppCenter setup is broken on version 6.1.0


lucasecf:
# Report
I'm using danger on our AppCenter CI for iOS builds, with the simple setup described in the website (pretty much providing `DANGER_GITHUB_API_TOKEN` and running `bundle exec danger`)

Everything was always working fine, but after updating danger to the latest version (6.1.0), the danger command during build started to give an error. 

Just in case, I reverted back to the previous version published on rubygems (6.0.9) and everything started to work fine again.

## What did you do?
Run danger on AppCenter CI: `bundle exec danger`

## What did you expect to happen?
For build to work properly like it was on version 6.0.9

## What happened instead?
Here's the failing log:
```
bundler: failed to load command: danger (/usr/local/lib/ruby/gems/2.6.0/bin/danger)
RuntimeError: Credentials not available. Provide DANGER_BITBUCKETCLOUD_USERNAME and DANGER_BITBUCKETCLOUD_PASSWORD as environment variables.
  /usr/local/lib/ruby/gems/2.6.0/gems/danger-6.1.0/lib/danger/request_sources/bitbucket_cloud_api.rb:99:in `fetch_json'
  /usr/local/lib/ruby/gems/2.6.0/gems/danger-6.1.0/lib/danger/request_sources/bitbucket_cloud_api.rb:78:in `fetch_pr_from_branch'
  /usr/local/lib/ruby/gems/2.6.0/gems/danger-6.1.0/lib/danger/request_sources/bitbucket_cloud_api.rb:15:in `initialize'
  /usr/local/lib/ruby/gems/2.6.0/gems/danger-6.1.0/lib/danger/request_sources/bitbucket_cloud.rb:27:in `new'
  /usr/local/lib/ruby/gems/2.6.0/gems/danger-6.1.0/lib/danger/request_sources/bitbucket_cloud.rb:27:in `initialize'
  /usr/local/lib/ruby/gems/2.6.0/gems/danger-6.1.0/lib/danger/danger_core/environment_manager.rb:37:in `new'
  /usr/local/lib/ruby/gems/2.6.0/gems/danger-6.1.0/lib/danger/danger_core/environment_manager.rb:37:in `block in initialize'
  /usr/local/Cellar/ruby/2.6.5/lib/ruby/2.6.0/set.rb:338:in `each_key'
  /usr/local/Cellar/ruby/2.6.5/lib/ruby/2.6.0/set.rb:338:in `each'
[...]

PR #1317, opened 2021-08-11
Fix AzurePipelines giving false positive to AppCenter builds

lucasecf:
Fixes issue https://github.com/danger/danger/issues/1173 

As reported in the issue above, running `danger` on AppCenter builds was broken since release 6.1.0. I had some time now to investigate and found out the problem was introduced by these changes: https://github.com/danger/danger/pull/1145

The root cause here is that `AzurePipelines` implementation is checking an env var with key `AGENT_ID` to be detected as a CI source.
Turns out that, AppCenter, a product also from Microsoft, have this same env var defined, so it was returning a false positive here when `danger` is running on AppCenter CI. You can see the full list of AppCenter's env vars in the linked issue. 

I don't have access to `AzurePipelines` to really know a proper unique env var key to replace `AGENT_ID`, so for now my workaround needed to be a bit 'hacky', but it shouldn't break current usages of Azure. The key used to validate AppCenter, `APPCENTER_BUILD_ID`, should not be present on Azure, given it contains the product name.

I left a comment anyways in hopes that someone with access can improve that in the future.

I tested my fork on our AppCenter CI and it fixed the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## master
 <!-- Your comment below here -->
 * Test with Ruby 2.7 and 3.0 on CI. - [@mataku](https://github.com/mataku)
+* Fix AppCenter CI source broken due to conflict with AzurePipelines - [@lucasecf](https://github.com/lucasecf)
 <!-- Your comment above here -->
 
 ## 8.3.1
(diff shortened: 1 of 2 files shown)


orta:
Thanks! Yeah, this makes sense to me, great writeup - sorry for the delay!

orta:
Thanks, this is shipped in 8.4.0!

PR #1319, opened 2021-08-30
Custom success message support

omarzl:
Hello! This pr adds support for a custom success message via the env var `DANGER_SUCCESS_MESSAGE`
I think it is better to handle it via an env var that adding a flag only for this


orta:
Sure, this seems reasonable to me 👍🏻

orta:
Thanks, this is shipped in 8.4.0!

PR #1325, opened 2021-10-01
Add option to bypass SSL verification for Bitbucket Server requests

robnadin:
We have an internal Bitbucket Server that cannot be posted to from our Bamboo instance without disabling SSL verification. I noticed this was possible to toggle for GitHub but not for BBS, so I've added support for it.


orta:
Yeah, this seems reasonable to me

PR #1328, opened 2021-10-07
Add a separation b/w filename-line number and message in gitlab inline comments

shashank68:
### Before

image

### From this PR

image


orta:
👍🏻 strange that this broke no tests, but seems reasonable to me

PR #1334, opened 2021-11-10
Support bearer_token for GitHub integrations

giginet:
## Motivation

I'd like to use danger with the GitHub App.
https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps

However, currently, danger only supports Personal Access Token. 

This PR makes an internal GitHub client to enable using `bearer_token` to authenticate.

## Description

Introduce a new environment variable `DANGER_GITHUB_BEARER_TOKEN`.
when it is set, danger makes GitHub clients use this value.

Of course, you can still use an access token via `DANGER_GITHUB_API_TOKEN`.

Thanks to this patch, we can use danger with GitHub App.

Screen Shot 2021-11-10 at 19 20 17

## Points of reviews

- I concern about the naming of the environment variable.
- how to update documentation?

### DIFF
--- a/lib/danger/ci_source/support/pull_request_finder.rb
+++ b/lib/danger/ci_source/support/pull_request_finder.rb
@@ -155,7 +155,7 @@ module Danger
         bearer_token = ENV["DANGER_GITHUB_BEARER_TOKEN"]
         if bearer_token && !bearer_token.empty?
           Octokit::Client.new(bearer_token: bearer_token, api_endpoint: api_url)
-        elsif access_token && !access_token&.empty?
+        elsif access_token && !access_token.empty?
           Octokit::Client.new(access_token: access_token, api_endpoint: api_url)
         else
           raise "No API token given, please provide one using `DANGER_GITHUB_API_TOKEN` or `DANGER_GITHUB_BEARER_TOKEN`"
--- a/lib/danger/ci_source/support/pull_request_finder.rb
+++ b/lib/danger/ci_source/support/pull_request_finder.rb
@@ -153,9 +153,9 @@ module Danger
         require "octokit"
         access_token = ENV["DANGER_GITHUB_API_TOKEN"]
         bearer_token = ENV["DANGER_GITHUB_BEARER_TOKEN"]
-        if !bearer_token&.empty?
+        if bearer_token && !bearer_token.empty?
           Octokit::Client.new(bearer_token: bearer_token, api_endpoint: api_url)
-        elsif !access_token&.empty?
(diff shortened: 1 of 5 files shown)


spec/lib/danger/ci_sources/support/pull_request_finder_spec.rb
```diff
@@ -128,6 +129,25 @@ def open_pull_requests_info
...
+        allow(Octokit::Client).to receive(:new).with(
+          bearer_token: ENV["DANGER_GITHUB_BEARER_TOKEN"],
+          api_endpoint: be_a(String)
+        ) { spy("Octokit::Client") }
+
+        finder(pull_request_id: "42", remote: true, logs: "not important").call
+      end
+    end
+
+    after do
```
giginet, line 145: these specs were fragile. These specs seem to share `ENV`s between test cases.
So they should be reset.

lib/danger/request_sources/github/github.rb
```diff
@@ -30,7 +30,8 @@ def initialize(ci_source, environment)
         self.support_tokenless_auth = false
         self.dismiss_out_of_range_messages = false
 
-        @token = @environment["DANGER_GITHUB_API_TOKEN"]
+        @access_token = @environment["DANGER_GITHUB_API_TOKEN"]
```
giginet, line 33: I renamed this instance variable. 
  This value is concealed. so it may not affect.

lib/danger/ci_source/support/pull_request_finder.rb
```diff
@@ -151,8 +151,15 @@ def client
...
-        Octokit::Client.new(access_token: ENV["DANGER_GITHUB_API_TOKEN"], api_endpoint: api_url)
-
+        access_token = ENV["DANGER_GITHUB_API_TOKEN"]
+        bearer_token = ENV["DANGER_GITHUB_BEARER_TOKEN"]
+        if bearer_token && !bearer_token.empty?
+          Octokit::Client.new(bearer_token: bearer_token, api_endpoint: api_url)
+        elsif access_token && !access_token.empty?
+          Octokit::Client.new(access_token: access_token, api_endpoint: api_url)
+        else
+          raise "No API token given, please provide one using `DANGER_GITHUB_API_TOKEN` or `DANGER_GITHUB_BEARER_TOKEN`"
```
giginet, line 161: before this patch, this implementation doesn't care existence of `access_token`.
It's better to be more strict for user-friendly.

orta:
Cool, yeah, this makes sense to me thanks

Issue #806, opened 2017-05-04
Pull Request Diff Comment Inline for Bitbucket Server


RishabhTayal:
# Report

This is a tracking issue for adding support for inline comments to Danger's Bitbucket Server integration.

For reference Bitbucket Server provides an [API](https://developer.atlassian.com/static/rest/bitbucket-server/5.0.0/bitbucket-rest.html#idm45659055842560) to implement this feature.

RishabhTayal:
Can someone point me in the right direction on how to implement this the right way?

orta:
I'd look into how it [works for GitHub](https://github.com/danger/danger/blob/master/lib/danger/request_sources/github/github.rb#L127), then see how much of it can be re-used for BitBucket.

vassilevsky:
Hi 👻 I wrote this tool a while back to post inline comments to Stash:

https://github.com/vassilevsky/face_control

You can use my code if you want.

I can also answer questions maybe, though I might not remember something already.

PR #1335, opened 2021-12-09
Improve how violations are handled with inline commenting on Bitbucket Server 

pahnev:
As BitBucket Server inline comments are done with the Code Insights annotations, some of the comments may get left out. As per their [API documentation](https://docs.atlassian.com/bitbucket-server/rest/7.18.2/bitbucket-code-insights-rest.html#idp10) only the annotations that are on lines changed in the unified diff will be displayed.

This PR makes a change in how the "rest of the violations" are gathered after the inline comments have been posted.
If environment variable `DANGER_BITBUCKETSERVER_INCLUDE_OUT_OF_RANGE_MESSAGES` is set to true, instead of rejecting all inline comments, it checks for the PR diff and compares the violations against it to determine if the violation should be included in the final comment message.

And thanks to @vassilevsky for doing the ground work for this https://github.com/danger/danger/issues/806#issuecomment-366610399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## master
 <!-- Your comment below here -->
+* Add option to post file specific comments outside of the PR diff to Bitbucket Server, when using Code Insights API. [@pahnev](https://github.com/pahnev)
 <!-- Your comment above here -->
 
 ## 8.5.0
--- a/lib/danger/request_sources/bitbucket_server.rb
+++ b/lib/danger/request_sources/bitbucket_server.rb
@@ -177,6 +177,7 @@ module Danger
 
       def find_position_in_diff?(file, line)
         return nil if file.nil? || line.nil?
+        return nil if file.empty?
         added_lines(file).include?(line)
       end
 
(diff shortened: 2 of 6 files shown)


orta:
Oof, sorry for missing this earlier - this conceptually looks reasonable to me but I'd like to have some unit tests to prove that we don't break it further down the line

pahnev:
Hey @orta, took a while to look into this again, but I rethought the change and opted out not to not break the existing behavior and instead have this to be activated with an env variable. I've also added some unit tests now.

orta:
Nice, thinking about it, we should probably default `DANGER_BITBUCKETSERVER_INCLUDE_OUT_OF_RANGE_MESSAGES` to be true? It feels like people might be more confused about the missing errors/warnings than be annoyed by the extra messages in the main comment

pahnev:
Honestly either way works for me. I guess if we set it to true, it's more inline how it works for GitHub?
Would changing the name to `DANGER_BITBUCKETSERVER_DISMISS_OUT_OF_RANGE_MESSAGES` make more sense in that case?

pahnev:
Hi @orta, how does it look now?

orta:
Thanks, yeah, lets get this in - it looks good to me!

PR #1346, opened 2022-01-31
Fix wrong arguments in ArraySubclass#respond_to_missing?

manicmaniac:
## Problem

`ArraySubclass#respond_to_missing?` has only one argument but `Object#respond_to_missing?` is defined to receive two arguments, according to the official API reference through 2.6.0 to 3.0.0.

So potentially it causes a bug even though I couldn't find it in a practical use case.

https://github.com/danger/danger/blob/43000055d824ca05a8c7ab63366d10e3d5c21529/lib/danger/helpers/array_subclass.rb#L22

https://docs.ruby-lang.org/en/2.6.0/Object.html#method-i-respond_to_missing-3F

> respond_to_missing?(symbol, include_all) → true or falseclick to toggle source
> respond_to_missing?(string, include_all) → true or false

## Solution

- Added a spec to reproduce a bug
- Fixed `ArraySubclass#respond_to_missing?` to receive two arguments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## master
 <!-- Your comment below here -->
 * Fix issue in GitHub `pr_draft?` method which was returning false for draft Pull Requests - [@rogerluan](https://github.com/rogerluan)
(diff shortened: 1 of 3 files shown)


orta:
Thanks, seems reasonable to me

PR #1348, opened 2022-02-14
Improve GitHub Actions workflows

mishina2228:
- Enclose 3.0 in quotes
  because ruby/setup-ruby uses Ruby 3.1 unless 3.0 is enclosed in quotes.
- Add Ruby 3.1 to the CI matrix
- bundler-cache enabled, so `bundle install` is no longer needed
- Split the steps to make the build status easier to see

I wonder why `bundle exec danger --verbose` is not executed(just printed), but this PR left it as it is.


orta:
Thanks

Issue #1352, opened 2022-02-23
Consider not posting "Invalid `Dangerfile` file" error as comment


rymai:
# Report

## What did you do?

When a `Dangerfile` is invalid, [the error is posted as a comment](https://github.com/danger/danger/blob/626c450619d19377424feb1071b22fda8a8a8141/lib/danger/danger_core/dangerfile.rb#L298). This can create security problems if the error contains tokens that are masked as CI/CD variable, but wouldn't be masked when printed in a comment.

Similarly to https://github.com/danger/danger/issues/1351, here the problem is that the error is actually posted as a comment, circumventing any CI/CD "masking" feature (such as the one GitLab provides) that hides variable values from job logs.

## What did you expect to happen?

Danger should not post a message in the PR/MR when the `Dangerfile` is invalid. The error can already be seen in the job logs, which should be sufficient and less confusing for people looking at the PR/MR comments.

## What happened instead?

Similarly to https://github.com/danger/danger/issues/1351, here the problem is that the error is actually posted as a comment, circumventing any CI/CD "masking" feature (such as the one GitLab provides) that hides variable values from job logs.

## Your Environment

* Which CI are you running on?
  * GitLab
* Are you running the latest version of Danger?
  * `8.4.2`
* What is your Dangerfile?

Doesn't really matter here.

manicmaniac:
Fixed in #1354

PR #1353, opened 2022-02-23
Don't save environment as an instance variable in GitLabCI

rymai:
This addresses https://github.com/danger/danger/issues/1351.

PR #1354, opened 2022-02-23
Don't post Dangefile errors as comment on PR when the `DANGER_DO_NOT_POST_INVALID_DANGERFILE_ERROR` env var is set

rymai:
This addresses https://github.com/danger/danger/issues/1352.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ## master
 <!-- Your comment below here -->
-
+* Don't post Dangefile errors as comment on PR when the `DANGER_DO_NOT_POST_INVALID_DANGERFILE_ERROR` env var is set - [@rymai](https://github.com/rymai) [#1354](https://github.com/danger/danger/pull/1354)
 <!-- Your comment above here -->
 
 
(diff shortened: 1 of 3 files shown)


orta:
I'd be open to us filtering the output here with the env vars, I'm not really sold on removing the feature

rymai:
> I'd be open to us filtering the output here with the env vars

Do you merging https://github.com/danger/danger/pull/1353? I think that would address the main issue for us. 👍 Then I'm ok adding an option to not to print errors as comments in PR/MR, if you'd accept that.

orta:
Yeah, I agree it could leak - I'm totally fine with having an option to skip the inline error report 👍🏻

rymai:
@orta Thanks! I changed the PR to not post failure comment when `DANGER_DO_NOT_POST_INVALID_DANGERFILE_ERROR` is set, and keeps the previous behavior otherwise. WDYT?

orta:
I think this is a good answer 👍🏻

rymai:
@orta Would it be possible to have a new `8.4.4.` version released with this fix? 🙏

orta:
Sure

orta:
Shipped

PR #1355, opened 2022-03-04
Never save environment as instance variables

rymai:
This is similar to https://github.com/danger/danger/pull/1353. I found out that there were other places where environment variables were saved as instance variables, leading to potential leakage in the case of the raising of a `DslError`.

This also redact a few more instance variables through the `#inspect` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## master
 <!-- Your comment below here -->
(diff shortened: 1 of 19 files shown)


rymai:
@orta This is another improvement to avoid any leakage of sensitive variables.

orta (approved):
Yeah, cool, this all looks good to me - thanks, will ship a build

PR #1356, opened 2022-03-15
New feature on VSTS/Azure DevOps: inline comment

damien-danglard:
# New feature on VSTS/Azure DevOps

This PR add a new feature to VSTS/Azure DevOps request source that already exists on other request sources: Inline comment.

This feature is simply an adaptation of the inline comments on the github for Azure DevOps / VSTS.


orta:
Nice, structurally looks good to me - I'd like to see some a tests around it (you can probably C&P the ones from github/gitlab and replace) mainly so that we dont end up breaking the changed in the future 👍🏻

damien-danglard:
> Nice, structurally looks good to me - I'd like to see some a tests around it (you can probably C&P the ones from github/gitlab and replace) mainly so that we dont end up breaking the changed in the future 👍🏻

It's done.

PR #1357, opened 2022-03-30
Fix `diff_for_file` on deleted files

AlexEvanczuk:
Similar to https://github.com/danger/danger/pull/788, we want to be able to get a diff for a file when it is deleted.


orta:
Looks reasonable to me 👍🏻

AlexEvanczuk:
@orta Can you unblock CI for me?

orta:
Thanks!

PR #1369, opened 2022-05-11
Support batch build in AWS codebuild

tomoya-yoshida-dwango:
`CODEBUILD_SOURCE_VERSION` environment variable returns commit hash(example: `6548dbc49fe57e1fe7507a7a4b815639a62e9f90`) in a batch build when triggered by a pull request.

`CODEBUILD_WEBHOOK_TRIGGER` environment variable rerurns `{event_name}/{event_variable}`(example `pr/123`) in both normal build and batch build.

> CODEBUILD_WEBHOOK_TRIGGER
Shows the webhook event that triggered the build. This variable is available only for builds triggered by a webhook. The value is parsed from the payload sent to CodeBuild by GitHub, GitHub Enterprise Server, or Bitbucket. The value's format depends on what type of event triggered the build.
For builds triggered by a pull request, it is pr/pull-request-number.
For builds triggered by creating a new branch or pushing a commit to a branch, it is branch/branch-name.
For builds triggered by a pushing a tag to a repository, it is tag/tag-name.

https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html

Therefore, use `CODEBUILD_WEBHOOK_TRIGGER` environment variable so that danger works only when event_type is pr.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 -->
 
 ## master
+* Support batch builds in AWS CodeBuild [@ihatov08](https://github.com/ihatov08)
 <!-- Your comment below here -->
 <!-- Your comment above here -->
 
--- a/lib/danger/ci_source/code_build.rb
+++ b/lib/danger/ci_source/code_build.rb
@@ -4,7 +4,8 @@ require "danger/request_sources/github/github"
 module Danger
   # ### CI Setup
   #
-  # In CodeBuild, make sure to correctly forward CODEBUILD_BUILD_ID, CODEBUILD_WEBHOOK_TRIGGER, CODEBUILD_SOURCE_REPO_URL and DANGER_GITHUB_API_TOKEN.
+  # In CodeBuild, make sure to correctly forward CODEBUILD_BUILD_ID, CODEBUILD_SOURCE_VERSION, CODEBUILD_SOURCE_REPO_URL and DANGER_GITHUB_API_TOKEN.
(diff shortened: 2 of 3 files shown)


orta:
Ah, I don't know enough about codebuild to know if this is a breaking change - can people self host?

tomoya-yoshida-dwango:
> Ah, I don't know enough about codebuild to know if this is a breaking change - can people self host?

Yes, we can.
We are using codebuild.
It is working.

orta:
If you can self-host then there may be cases where this new code breaks old builds - any chance you can make it backwards compatible?

tomoya-yoshida-dwango:
For safety, if you want to keep the old ones working, I can implement it.

`CODEBUILD_BATCH_BUILD_IDENTIFIER` environment variable seems to be able to determine the mode of the code build, so use this to maintain backward compatibility.

tomoya-yoshida-dwango:
Or is it better to have backward compatibility with a new environment variable that is not a `CODEBUILD_BATCH_BUILD_IDENTIFIER` environment variable?

orta:
Yep, keep `CODEBUILD_WEBHOOK_TRIGGER` working if `CODEBUILD_WEBHOOK_TRIGGER` is not present

tomoya-yoshida-dwango:
@orta 
`CODEBUILD_WEBHOOK_TRIGGER` environment variable always returns a value at coudebuild when created PR.(ex `pr/121`)

`CODEBUILD_BATCH_BUILD_IDENTIFIER` returns a value only with batch build(ex `build1`)

I implement it has backward compatibility depending on the present `CODEBUILD_BATCH_BUILD_IDENTIFIER`.
Please review 🙇

orta:
Looks great, sorry of the delay!

PR #1377, opened 2022-08-22
Update Octokit gem dependency to version 5.x

mathroule:
Danger relies on Octokit gem dependency version `~> 4.7`, which makes it impossible to use Octokit version 5.0 and above when dependencies are resolved with Bundler.

PR #1378, opened 2022-08-22
Drop support for Ruby 2.4, support Ruby 2.7+

mathroule:
Require Ruby version 2.7+. Ruby 2.6 has reached end of life: https://www.ruby-lang.org/en/downloads/branches. This change is a prerequisite to https://github.com/danger/danger/pull/1377.

Changes:
- Use Ruby 2.7
- Use bundler 2.3.20

AppVeyor project configuration needs to be changed to use Visual Studio 2022 (or at least 2019) to support Ruby 2.7 (https://www.appveyor.com/docs/windows-images-software/#ruby).
Screen Shot 2022-08-22 at 19 19 47

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/danger
     docker:
-      - image: ruby:2.4
+      - image: ruby:2.7
         environment:
         - RAILS_ENV=test
         - RACK_ENV=test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - checkout
       - run: gem update --system 2.5.0
-      - run: gem install bundler -v '2.0.1'
+      - run: gem install bundler -v '2.3.20'
       - restore_cache:
           key: gem-cache-{{ .Branch }}-{{ checksum "Gemfile" }}
       - run: bundle install --path vendor/bundle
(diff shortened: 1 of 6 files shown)


Kaspik (approved):
LGTM, soudn good. We should add support (tests) for 3.0+ too.

mathroule:
> LGTM, soudn good. We should add support (tests) for 3.0+ too.

Danger is tested against Ruby 3.+ on GitHub Actions: https://github.com/danger/danger/blob/master/.github/workflows/CI.yml#L10

I updated the CircleCI config file, but Danger appears to be not built anymore on this platform.

mathroule:
@orta could you please help with the AppVeyor project settings?

Kaspik:
I think we can skip AppVeyor completely as we moved to Github Actions?

mathroule:
> I think we can skip AppVeyor completely as we moved to Github Actions?

I'm not against merging the PR as is. But it's just a matter of changing an option in AppVeyor to keep the compatibility.

orta:
Updated to VS 2022 - we have the appveyor for testing windows compat, dropping that is OK if we check on windows using GitHub Actions  👍🏻

orta:
This'll have to be a semver major bump alas

mathroule:
> Updated to VS 2022 - we have the appveyor for testing windows compat, dropping that is OK if we check on windows using GitHub Actions 👍🏻

@orta Windows checks done on GitHub Actions in https://github.com/danger/danger/pull/1381.

orta:
Cool, I've carved out some time to go through all this tomorrow 👍🏻

PR #1380, opened 2022-09-01
Allow import_dangerfile to accept full path

### LINKED ISSUE #1379: Allow `danger.import_dangerfile(path:)` with full file path
# Report
I'm trying to split Dangerfile into multiple smaller files.
For that, I am trying to use `danger.import_dangerfile(path:'scripts/DangerExtensions/child_danger1.rb')` method.
But this gives the error 
`[!] Invalid `Dangerfile` file: Not a directory @ rb_sysopen - scripts/DangerExtensions/child_danger1.rb/Dangerfile. Updating the Danger gem might fix the issue. Your Danger version: 8.5.0, latest Danger version: 8.6.1`

When giving the full path to this method, danger should import the file and run it.

## What happened instead?
When giving the full path to this method, danger blindly appends `Dangerfile` to it and tries to open the file.
https://github.com/danger/danger/blob/314bb927605beb862137498e36a3b4bfff413dc7/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb#L133-L137

With this, if I have multiple files, I have to place each file with the name `Dangerfile` in separate folders, which is a limitation I would say.

## Your Environment

* Which CI are you running on?
   Github actions
* Are you running the latest version of Danger?
   Danger version `8.5.0`
* What is your Dangerfile?

  ```ruby
  danger.import_dangerfile(path: 'scripts/DangerExtensions/child_danger1.rb')
  ```

PS: I am happy to open a PR to fix this if maintainers agree with this 🙂

jkmathew:
Currently `import_dangerfile(path: '...')` assumes the user provides the path to the directory containing a Dangerfile.
This is not the case when we are splitting large Dangerfiles into smaller ones. We have to place each Dangerfile in separate directories. 
In this PR I modified the `import_dangerfile_from_path` method to check if the provided path is a file, if yes, then use that path to import. Otherwise, append `Dangerfile` to the path and import it.

Fixes #1379

### DIFF
--- a/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb
+++ b/lib/danger/danger_core/plugins/dangerfile_danger_plugin.rb
@@ -132,7 +132,7 @@ module Danger
     #
     def import_dangerfile_from_path(path)
       raise "`import_dangerfile_from_path` requires a string" unless path.kind_of?(String)
-      local_path = File.join(path, "Dangerfile")
+      local_path = File.file?(path) ? path : File.join(path, "Dangerfile")
       @dangerfile.parse(Pathname.new(local_path))
     end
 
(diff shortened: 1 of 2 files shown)


orta:
Cool, thanks 👍🏻

PR #1381, opened 2022-09-01
Run GitHub Actions workflow on macOS, Ubuntu and Windows

mathroule:
# Description
Use GitHub Actions matrix to run checks on macOS, Ubuntu and Windows.

# Changes
- Use GitHub Actions matrix with OS
- Cancel concurrent GitHub Actions workflow run
- Check only pull requests to `master` branch
- Reformat code

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     strategy:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,11 +4,11 @@ on: [pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
     strategy:
       matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
         ruby-version: [2.7, '3.0', 3.1]
-
+    runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v1
     - name: Set up Ruby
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -1,30 +1,41 @@
...
-        git config --global user.name "Danger.Systems"
-    - name: Running Tests
-      run: bundle exec rake spec
-    - name: Running the Dangerfile for this repo
-      run: |
-        TOKEN='7469b4e94ce21b43e3ab7a'
-        TOKEN+='79960c12a1e067f2ec'
-        DANGER_GITHUB_API_TOKEN=$TOKEN RUNNING_IN_ACTIONS=true echo 'bundle exec danger --verbose'
+      - name: Running the Dangerfile for this repo
+        if: runner.os != 'Windows'
```
mathroule, line 37: Skip these commands on Windows, since it does not work. This should work on Windows:
```yaml
          set TOKEN=7469b4e94ce21b43e3ab7a
          set TOKEN=%TOKEN%79960c12a1e067f2ec
          set DANGER_GITHUB_API_TOKEN=%TOKEN%
          set RUNNING_IN_ACTIONS=true
          bundle exec danger --verbose
```

By the way Danger is not properly run on CI, only the command is displayed with the `echo`.

orta: ha, oh dear. Well, this shouldn't block your PR.

orta:
OK, I've removed the webhooks for AppVeyor, so this should be some of the last builds over there

.github/workflows/CI.yml
mathroule: Fix the Danger run issue in https://github.com/danger/danger/pull/1382.

PR #1382, opened 2022-09-03
Run Danger on pull requests to `danger/danger` GitHub repo

mathroule:
Run Danger on pull requests to `danger/danger` GitHub repository.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -58,4 +58,5 @@ jobs:
       - name: Running the Dangerfile for this repo
         env:
           DANGER_GITHUB_API_TOKEN: 7469b4e94ce21b43e3ab7a79960c12a1e067f2ec # FIXME Store it in GitHub Actions secrets
+          RUNNING_IN_ACTIONS: true
         run: bundle exec danger --verbose
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  run-tests:
+  build:
     strategy:
       matrix:
         os:
@@ -44,18 +44,8 @@ jobs:
       - name: Running tests
         run: bundle exec rake spec
 
-  run-danger:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v1
-
-      - name: Set up Ruby
-        uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: 3.1
-          bundler-cache: true
-
       - name: Running the Dangerfile for this repo
+        if: runner.os != 'Windows'
         env:
           DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           RUNNING_IN_ACTIONS: true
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -30,12 +30,21 @@ jobs:
...
+          bundler-cache: true
+
       - name: Running the Dangerfile for this repo
-        if: runner.os != 'Windows'
-        run: |
-          TOKEN='7469b4e94ce21b43e3ab7a'
-          TOKEN+='79960c12a1e067f2ec'
-          DANGER_GITHUB_API_TOKEN=$TOKEN RUNNING_IN_ACTIONS=true echo 'bundle exec danger --verbose'
+        env:
+          DANGER_GITHUB_API_TOKEN: 7469b4e94ce21b43e3ab7a79960c12a1e067f2ec # FIXME Store it in GitHub Actions secrets
```
mathroule, line 49: @orta do you think a new GitHub API token can be generated add added as GitHub Actions secret for https://github.com/danger-public?

orta: Afraid not, needs to be in the code like above, secrets wont work here because it only runs on branch PRs and not forks

orta: Ah, you probably broke the old token by putting it in a single string too, will need to find the password for that account

mathroule: The token is not authorized anymore, if my change deauthorized the token sorry about that.
Indeed storing it as GitHub Actions secrets, will make it not available for fork PRs, but I'm not sure splitting the token in 2 parts, is the best workaround to use in terms of security.

mathroule: Actually, GitHub advises solving the problem differently: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows.

I'll test that in this PR.

mathroule: In any case the `DANGER_GITHUB_API_TOKEN` must be added as GitHub Actions secrets.

mathroule: Unfortunately using `pull_request_target` or `workflow_run` triggers will not work out of the box. Indeed, the workflow will be executed in the context of the base branch, not of the head branch / PR context.

mathroule: Using env var this way works, but it will not work for forked pull requests:
```
env:
  DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/CI.yml
```diff
@@ -45,8 +45,7 @@ jobs:
         run: bundle exec rake spec
 
       - name: Running the Dangerfile for this repo
-        if: runner.os != 'Windows'
-        run: |
-          TOKEN='7469b4e94ce21b43e3ab7a'
-          TOKEN+='79960c12a1e067f2ec'
-          DANGER_GITHUB_API_TOKEN=$TOKEN RUNNING_IN_ACTIONS=true echo 'bundle exec danger --verbose'
+        env:
+          DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
orta, line 49: secrets cannot be accessed by forked PRs, so this change isn't doing what you want it to do I think

PR #1387, opened 2022-09-21
Timestamp changelog

sandstrom:
Added timestamps, makes it easier to understand if there are recent changes one should know about.

I only backfilled the past few releases. But would be great with timestamps going forward.

Thanks for an awesome project! 🏅


orta:
Thanks for thinking of a cool idea, but I'm not really into adding more work for myself here - I've maintained this since 2015 by keeping the process light

PR #1400, opened 2022-11-13
Fix example for checking merge commit

### LINKED ISSUE #1350: Unreliable merge commit check in examples
https://github.com/danger/danger/blob/626c450619d19377424feb1071b22fda8a8a8141/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb#L25

Checking for merge commits this way causes a false positive on regular commits with message beginning with `Merge branch 'master'` and false negative on merge commits with commit subject altered so it wouldn't match the regex.

Consider using something like `c.parents.count > 1` instead

manicmaniac:
Close #1350


orta:
This makes sense to me, thanks

PR #1401, opened 2022-11-13
Use `in_reply_to_id` to lookup GitHub replies

kemchenj:
Currently, danger's comment replies lookup logic for GitHub is basically guessing, cause at that time GitHub doesn't have a concept for "reply".

The good news is GitHub API has filled this pit:

```javascript
// https://docs.github.com/en/rest/pulls/reviews#list-reviews-for-a-pull-request
"in_reply_to_id": {
  "description": "The comment ID to reply to.",
  "type": "integer",
  "examples": [
    8
  ]
}
```

So let's move on with this new right way.


manicmaniac:
@kemchenj

Looks really nice :+1:
Could you add some tests for this change?

At a glance, there seems no tests for inline comments with replies but this test case looks similar.

https://github.com/danger/danger/blob/bd48382369778d37ac229304d32f113ffdd758be/spec/lib/danger/request_sources/github_spec.rb#L498

orta:
Does anyone know when this was added? Older GitHub enterprises may not have this in their API, which is what I'd be wary of

xximwon (approved):

PR #1402, opened 2022-11-13
Fix GitHub inline commens cleanup logic

kemchenj:
Currently, danger cleans up all the outdated inline comments after submitting inline comments. 

But there is an early return when there are no inline comments needed to submit before all, and it breaks the cleanup logic. So we might want to remove it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## master
 
+* Fix GitHub inline commens cleanup logic - [@kemchenj](https://kemchen.github.io/)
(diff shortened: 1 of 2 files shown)


orta:
( the CI fails seem to be legit? )

reedrolemodel:
Could an admin re-run CI? I am interested in getting this merged.

manicmaniac:
I'm sorry for the delay but https://github.com/danger/danger/pull/1477 solves the same problem as well :eyes:

PR #1406, opened 2022-11-20
Fix offenses for RuboCop that can be fixed automatically

ydah:
This PR fix offenses for RuboCop that can be fixed automatically.
The automatic modification performed in this PR is performed by `rubocop -a`.

In the process, I also implement the following:

Removed RuboCop version specification so that the latest RuboCop can be used.
This is because using the latest version will prevent false positive offenses and erroneous automatic corrections.

The newly added cops is not currently working.
This is because the purpose of this PR is to address the current violations of RuboCop, not the newly added cops. 

Ref: https://docs.rubocop.org/rubocop/versioning.html#enablingdisabling-pending-cops-in-bulk

Add inherit_mode for `merge: Exclude`.
The following warnings have been addressed:

```
`Lint/UselessAssignment: Exclude` in `.rubocop.yml` overrides a generated `Exclude` in `.rubocop_todo.yml`. Either remove `Lint/UselessAssignment: Exclude` or add `inherit_mode: merge: - Exclude`.
Created .rubocop_todo.yml.
```

For those that cannot be modified automatically, they are once put into todo status in `.rubocop_todo.yml`.
I would like to check these as well if this PR is merged.


lib/danger/danger_core/plugins/dangerfile_git_plugin.rb
```diff
@@ -46,7 +46,7 @@ def self.instance_name
 
     def initialize(dangerfile)
       super(dangerfile)
-      raise unless dangerfile.env.scm.instance_of?(Danger::GitRepo)
+      raise unless dangerfile.env.scm.class == Danger::GitRepo # rubocop:disable Style/ClassEqualityComparison
```
ydah, line 49: Addressing this offense would change the behavior.
It actually causes the test to fail, so it is disabled: https://github.com/danger/danger/actions/runs/3508633313/jobs/5877172874

orta:
Looks good to me, given it a few reads though and everything seems reasonable to me- thanks (sorry for the months of delay!)

PR #1410, opened 2023-01-05
Added dependabot for GitHub Actions #trivial

ydah:
How about using dependabot in this way?
When this works, a PullRequest is created as follows:
- ruby/bigdecimal#242

It was created with reference to the following:
- ruby/csv#265

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: "weekly"


orta:
I'm not too wild on this, it's been a very long time since I kept up with the ruby world - I'm pretty happy with the rate of occasionally handling a pr or two a month but this would just mean I ignore loads of PRs probably!

ydah:
I didn't explain myself well enough. This response will create a Pull Request to be updated when the version of, for example, actions/checkout used within the GitHub Actions workflow is upgraded.

Basically, since a major version is specified in GitHub Actions workflow, a Pull Request is created when a major version is raised. So I am thinking that since it is not updated that frequently, there is less noise and it is not likely to be missed.
https://github.com/danger/danger/blob/7706e9c3872821cc6ce888f91aae8c0c6d627a1c/.github/workflows/CI.yml#L27
https://github.com/danger/danger/blob/7706e9c3872821cc6ce888f91aae8c0c6d627a1c/.github/workflows/CI.yml#L31

orta:
Ah, that seems reasonable then - thanks!

PR #1415, opened 2023-01-16
Remove `< 2.0` Faraday version restriction

vilmantas:
Updating to allow Faraday 2.0.0. 

Locking between versions of ">= 0.9.0", "< 2.0" was causing new projects to have dependency conflicts.

Related: https://github.com/danger/danger/issues/1349


orta:
You've got CI running properly now, looks like its got some failing tests in the gitlab side

vilmantas:
@orta 

not sure if I understand the errors. Are the other gems broken due to this change?

orta:
I dont know much about web mock - but maybe it also uses faraday and by allowing for the pdated faraday it breaks webmock?

orta:
Thanks folks

orta:
Shipped as 9.2.0

PR #1419, opened 2023-01-28
Security update: ruby-git

manicmaniac:
Addresses https://nvd.nist.gov/vuln/detail/CVE-2022-47318

> ruby-git versions prior to v1.13.0 allows a remote authenticated attacker to execute an arbitrary ruby code by having a user to load a repository containing a specially crafted filename to the product. This vulnerability is different from CVE-2022-46648.

I think this issue doesn't affect on users in a regular use case.
But for someone who uses Danger and exposes internet-facing API that uses ruby-git, this PR could be helpful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ## master
 
-* Update ruby-git gem [@manicmaniac](https://github.com/manicmaniac) [#1419](https://github.com/danger/danger/pull/1419)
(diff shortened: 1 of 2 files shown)


orta:
👍🏻

mfechner:
@manicmaniac 
Is there a special reason why it was set to `1.13.0` and not to `1.13`, like it was before with `1.7`?

manicmaniac:
> @manicmaniac 
> Is there a special reason why it was set to `1.13.0` and not to `1.13`, like it was before with `1.7`?

I guess there is no special reason to do that.
Now I think it is better to set to 1.13.

PR #1420, opened 2023-01-28
Explicitly specify initial branch to avoid being affected by global git config

manicmaniac:
## Problem

Some specs always fail if `init.defaultBranch` is not `master` in `~/.gitconfig`.

## Cause

Those spec call `git init` and it determines the initial branch name by global git config since [git 2.28](https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch).

## Solution

I replaced every `git init` with `git init -b master` to set the initial branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ## master
 
-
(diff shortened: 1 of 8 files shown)


orta:
Seems very reasonable to me 👍🏻

Issue #1405, opened 2022-11-18
`danger pr` is too verbose, and can't turn it off


technicalpickles:
# Report

## What did you do?

```shell
$ danger run <pr>
```


## What did you expect to happen?

I expected Danger to run, and for me to be able to scroll back to look at the output.


## What happened instead?

I get a flood verbose output for (seemingly) each plugin , ie:

```
Running your Dangerfile against this PR - <url>
Turning on --verbose

# pages and pages of the following repeated, enough to fill my terminal's buffer

Info:

+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|                                                                                                          Danger v8.4.2                                                                                                           |
|                                                                                                          DSL Attributes                                                                                                          |
+------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|    status_report |                                                                                                                                                                                                               |
| violation_report |                                                                                                                                                                                                               |
|         warnings | []                                                                                                                                                                                                            |
|        markdowns | []
```

I tried `danger --verbose=false pr <url>` and `danger pr --verbose=false pr <url>`, but it doesn't change anything. It does seem to accept the flag though, because if it wasn't a registered one, there would be a 'unknown option' error.


## Your Environment

* Which CI are you running on? n/a
* Are you running the latest version of Danger?
* What is your Dangerfile?

```ruby
danger.import_dangerfile(gem: 'some-internal-gem')
danger.import_dangerfile(gem: 'another-internal-gem')

# this repeated dozens of times
danger.import_plugin('danger/some_local_check.rb')
some_local_check.check
  ```

manicmaniac:
It looks like an intended behavior since `Danger::LocalSetup#setup` forces `--verbose` option here.

https://github.com/danger/danger/blob/v9.2.0/lib/danger/commands/local_helpers/local_setup.rb#L31-L34

Not only `danger local` but also `danger pr` calls `Danger::LocalSetup#setup`, as a result they have the same behavior about forsing `--verbose` flag.

manicmaniac:
@orta
How do you think about removing the behavior that forces `--verbose` flag in `danger local` and `danger pr`?
Since both of them are mainly for debugging purpose, I don't think no one cares about this breaking change. Those who want the current behavior are still able to add `--verbose` flag explicitly.

orta:
Seems reasonable to me

manicmaniac:
[...]

PR #1424, opened 2023-01-30
Do not force verbose in `danger local` and `danger pr`

manicmaniac:
Close https://github.com/danger/danger/issues/1405

- Prevent `danger pr` and `danger local` from enforcing verbose flags
- Explicitly passing `--verbose` like `danger pr --verbose` prints verbose output


xximwon (approved):

davkutalek:
Can we get this into a release? The always verbose behavior is really annoying.

Mondjoe (approved):

Issue #1431, opened 2023-03-27
Support custom CI check names


Vacxe:
# Feature request:

We have a multiple checks on CI with Danger
```
@acoe-mobile
danger/pull-request
danger/swiftformat
danger/swiftlint 
danger/androidcodestyle
....
````

Is it possible to customise prefix name? for example `android/codestyle, android/detekt, global/pull-request`

If this functionality not provided or not documented Ill be happy to implement it

orta:
I think the danger id argument does this, if not it probably should 

On Mon, Mar 27, 2023, at 6:38 AM, Konstantin Aksenov wrote:
> 
> 
> Feature request:
> 
> We have a multiple checks on CI with Danger
> 
> ***@***.***
> danger/pull-request
> danger/swiftformat
> danger/swiftlint 
> danger/androidcodestyle
> ....
`

Vacxe:
@orta appreciate for assistance! Looks right to me.

Vacxe:
I checked with `danger_id`. The behaviour is slightly different. 
For example I setting it as `Foo` then report id will be "danger/Foo". Wanna have the same configuration flexibility for prefix. to have smth like "Bar/Foo" instead of "danger/Foo"

PR #1432, opened 2023-03-29
Remove redundant github `danger/` prefix

Vacxe:
Close: #1431

Technically all CI checks on GH from Danger having prefix `danger/` what actually very inflexible if we can have an option to use `danger_id` CLI argument

---
By default check result is `danger/danger` what actually looks pretty strange. Probably just a `danger`  will be enough as a default configuration.

---
Without prefix CI check will be more flexible for some specific cases:
For example we having a monorepo with PlatformA and PlatformB. Both of them is independent and we having **required** checks:
PlatformA/check1
PlatformA/check2
PlatformA/danger

PlatformB/check1
PlatformB/check2
PlatformB/danger

Then we can create a some short circuiting bases on platform name kind of "If directory with PlarformA have no changes then set all checks PlatformA/* as Skipped"`

---

Or even if it required it can be declared `--danger_id="danger/danger"`

---

However those changes will break backward compatibility if previous danger check are required. And project maintainers should update required check name in this case.

What will be the best option in this case? IMO having 2 variables for declaring prefix and postfix is redundant and having only one looks straight forward for me

### DIFF
--- a/lib/danger/request_sources/github/github.rb
+++ b/lib/danger/request_sources/github/github.rb
@@ -219,7 +219,7 @@ module Danger
         begin
           client.create_status(ci_source.repo_slug, latest_pr_commit_ref, status, {
             description: message,
-            context: "danger/#{danger_id}",
+            context: danger_id,
             target_url: details_url
           })
         rescue StandardError
--- a/spec/lib/danger/request_sources/github_spec.rb
+++ b/spec/lib/danger/request_sources/github_spec.rb
@@ -263,7 +263,7 @@ RSpec.describe Danger::RequestSources::GitHub, host: :github do
       end
 
       it "uses danger_id as context of status" do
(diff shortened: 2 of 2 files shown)


Vacxe:
@orta

orta:
Nah, I don't think this is worth the major semver breaking change. I know I've relied on checking for danger/danger (which represents this repo, danger js has `danger/danger-js`) from other tools for example, which means there's very likely a bunch of people doing that now for `danger/mysubset`. It will not be obvious to them why something like this broke.

I wondered if including a `/` would be enough to allow skipping the prefix, but even that I can see would probably break someone's infra. I think this is going to need to be some other way, or just be OK with the prefix

Vacxe:
No worries. Thanks for response.

PR #1435, opened 2023-04-11
feat: Add new CustomCiWithGithub CI Source Type

jmeridth:
Add CI Source that doesn't match any third party vendors (like an internal CI system) but it uses GitHub for source control.


jmeridth:
This is a good v0.1 approach.  May want to make this its own abstraction for each souce control type (GitHub, Bitbucket, Gitlab, etc). `Danger::CI::CustomCI::GitHub` or something like that

orta:
What is your actual goal with this? We already have some env vars for running a FakeCI in Danger I believe, and this code is effectively duplicating that

jmeridth:
@orta bump

orta:
OK, yeah, I think this is reasonable - if you add the inline docs to the class (as it will show up on the site) I'm fine with merging it

jmeridth:
Ran tests locally

```bash
Finished in 59.1 seconds (files took 1.14 seconds to load)
1240 examples, 0 failures, 2 pending
```

jmeridth:
@orta added the inline docs and fixed merge conflict with `CONTRIBUTING.md`.  Fixed typo in a spec and now they are passing locally.

jmeridth:
@orta bump. Thank you for reviewing.

lib/danger/ci_source/custom_ci_with_github.rb
```diff
@@ -0,0 +1,44 @@
...
+  # And so, you can use GITHUB_TOKEN secret as `DANGER_GITHUB_API_TOKEN` environment variable.
+  #
+  #  ```yml
+  #  ...
+  #    steps:
+  #      - uses: actions/checkout@v3
+  #      - uses: danger/danger@master
+  #        env:
+  #          DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  #  ```
```
orta, line 16: These docs don't describe how to use this setup (via the env var: `CUSTOM_CI_WITH_GITHUB`) - it seems to be a copy of the docs for GitHub Actions support? https://github.com/danger/danger/blob/master/lib/danger/ci_source/github_actions.rb

jmeridth: yes.  usage is identical.  only diff is the environment key.  there is opportunity for refactoring to a base class maybe.  is that preferred?

orta: I don't really care about making it a shared abstraction, they're shallow and have hard boundaries - I worry as someone reading the docs looking at this and thinking "which do I use for github actions?" because this to me looks like the GitHub Actions CI setup but wouldn't actually work

The docs need to explain what an "Custom CI" is, what env var needs to be set, and why you would use it instead of the github actions docs

jmeridth: @orta apologies.  very bad copy paste previously.  inline docs updated.

orta (approved):
Nice, describes it perfectly 👍🏻

orta:
Shipped as 9.3.0

PR #1440, opened 2023-06-02
Suppress new RuboCop offenses #trivial

koic:
This PR suppresses the following new RuboCop offenses:

```console
% bundle exec rubocop
(snip)

Offenses:

lib/danger/ci_source/azure_pipelines.rb:32:1: C: [Correctable] Layout/EmptyLinesAroundClassBody: Extra empty line detected at class body beginning.
lib/danger/ci_source/azure_pipelines.rb:33:1: C: [Correctable] Layout/EmptyLines: Extra blank line detected.
lib/danger/ci_source/custom_ci_with_github.rb:1:1: C: [Correctable] Style/FrozenStringLiteralComment: Missing frozen string literal comment.
require "danger/request_sources/github/github"
^
lib/danger/ci_source/custom_ci_with_github.rb:33:5: W: Lint/MissingSuper: Call super to initialize state of the parent class.
    def initialize(env) ...
    ^^^^^^^^^^^^^^^^^^^
lib/danger/ci_source/support/pull_request_finder.rb:149:1: C: [Correctable] Layout/TrailingWhitespace: Trailing whitespace detected.
lib/danger/ci_source/support/pull_request_finder.rb:184:14: C: [Correctable] Layout/TrailingWhitespace: Trailing whitespace detected.
        :vsts
             ^
lib/danger/request_sources/bitbucket_cloud.rb:74:9: W: [Correctable] Lint/UselessAssignment: Useless assignment to variable - comment.
        comment += "\n\n"
        ^^^^^^^
lib/danger/request_sources/bitbucket_cloud.rb:81:25: C: [Correctable] Style/NumericPredicate: Use warnings.count.positive? instead of warnings.count > 0.
        has_comments = (warnings.count > 0 || errors.count > 0 || messages.count > 0 || markdowns.count > 0)
                        ^^^^^^^^^^^^^^^^^^
lib/danger/request_sources/bitbucket_cloud.rb:81:47: C: [Correctable] Style/NumericPredicate: Use errors.count.positive? instead of errors.count > 0.
        has_comments = (warnings.count > 0 || errors.count > 0 || messages.count > 0 || markdowns.count > 0)
                                              ^^^^^^^^^^^^^^^^
lib/danger/request_sources/bitbucket_cloud.rb:81:67: C: [Correctable] Style/NumericPredicate: Use messages.count.positive? instead of messages.count > 0.
        has_comments = (warnings.count > 0 || errors.count > 0 || messages.count > 0 || markdowns.count > 0)
[...]

PR #1444, opened 2023-06-24
Run RuboCop for linting on CI #trivial

koic:
I noticed in #1440 that CI is not running RuboCop. So this PR makes CI run RuboCop for linting.
RuboCop version is specified in the Gemfile to avoid sudden offenses from new cops.


orta:
Yeah, as long as it's locked, I think that's pretty reasonable - thanks!

Issue #1110, opened 2019-03-20
Provide an official Docker image


pbendersky:
# Report

For projects that are not Ruby based (let's say Java), it would be great to have an official Docker image that tracks the latest version of Danger so we can run it easily from CI.

In my use case, we run maven to compile a Java project, and that leaves some checkstyle output files. Without a Danger image, I need to add Ruby to the Docker image that runs the build. With a danger image, I could simple run my Dangerfile from it.

I can contribute it, but need some pointers on what repo to use, and who can set up the automation in Docker Hub (I noticed there's something set for danger/peril already).

orta:
Cool, I've added the auto generator to this repo: https://cloud.docker.com/repository/registry-1.docker.io/dangersystems/danger-rb

It fails right now, but I'll take PRs adding it 👍

RDIL:
If you invite me to the Docker organization, I can give it a try.  Docker ID is jumbocakeyumyum.

orta:
I'd try building it locally to fix it, it's hooked up to deploy on any merged PRs to master (I don't have access myself either, I have it going through a bot account)

jk:
For what it's worth, we use the docker image, GitLab uses for its GitLab itself and install some additional stuff. Our dockerfile (which is build on our internal gitlab-ci) looks like this:

```dockerfile
FROM registry.gitlab.com/gitlab-org/gitlab-build-images:danger

RUN env | sort && \
    danger --version && \
    gem update danger danger-gitlab && \
    gem install \
    danger-commit_lint \
    danger-the_coding_love \
    danger-mention \
    danger-welcome_message \
    danger-prose \
    danger-lgtm \
    danger-junit \
    danger-linear_history \
    --no-document
```

The base image itself is defined here: https://gitlab.com/gitlab-org/gitlab-build-images/blob/master/Dockerfile.danger

okainov:
Is there any update on this one?

orta:
Nope, you're welcome to take a look

fkorotkov:
BTW it seems the GitLab image mentioned above is published as `registry.gitlab.com/gitlab-org/gitlab-build-images:danger`

PR #1445, opened 2023-06-30
Release to GitHub Container Registry

unlobito:
[[dangergem.slack.com context](https://dangergem.slack.com/archives/C0BHS23PG/p1688132250783649)]

Fixes https://github.com/danger/danger/issues/1110

This PR adds a workflow similar to [danger/swift](https://github.com/danger/swift/blob/master/.github/workflows/publish_package.yml), [danger/kotlin](https://github.com/danger/kotlin/blob/master/.github/workflows/publish_package.yml), and [danger/danger-js](https://github.com/danger/danger-js/blob/main/.github/workflows/publish_package.yml), but has been updated to make use of new `github` context variables for easier setup (see https://github.com/docker/build-push-action#usage + https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages). If this PR is accepted, I'd be happy to port these changes to the aforementioned repos.

It will publish an image on `ghcr.io/danger/danger` on each new tag (e.g., `ghcr.io/danger/danger:v9.3.1` for latest) and on new pushes to the default branch (at `ghcr.io/danger/danger:master`).

### DIFF
--- a/.github/workflows/publish_package.yml
+++ b/.github/workflows/publish_package.yml
@@ -1,0 +1,46 @@
+name: Release Danger package
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - '*'
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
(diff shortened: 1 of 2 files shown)


orta:
Seems reasonable to me 👍🏻

unlobito:
It seems we're succeeding in publishing the container image (https://github.com/danger/danger/pkgs/container/danger/106254064), but struggle to tag it correctly (https://github.com/danger/danger/actions/runs/5436611587/jobs/9886592800#step:5:686).

Investigating & will follow up with another PR. Apologies for the disruption here.

PR #1448, opened 2023-07-10
Drop no_proxy_fix gem

stanhu:
This gem was only needed to fix [a bug in Ruby 2.4](https://github.com/ruby/ruby/pull/1513), but since the required Ruby version of this gem is 2.7.0 we do not need this patch.


orta:
Thanks, yeah, this looks reasonable to me

manicmaniac:
@stanhu I'm sorry for the delay. The change looks good. Could you rebase it on master?

manicmaniac:
This file still requires `no_proxy_fix` :eyes:

https://github.com/danger/danger/blob/96d86e362a8e9218105a82054010ed9d6a5c50ef/lib/danger/commands/pr.rb#L7

manicmaniac:
@stanhu Thank you for your contribution 👍

PR #1449, opened 2023-08-01
No need to install gitlab for less than Ruby 2.5.0 #trivial

mataku:
The current required_ruby_version is specified as ">= 2.7.0", so there is no need to consider Ruby version "< 2.7.0".

It seems that the minimum Ruby version was changed to 2.5 in gitlab gem 4.14.1, so special handling was required.


orta:
Agree, thanks!

PR #1460, opened 2023-09-30
Support octokit 7.x

ngan:
Update gemspec to support octokit 7.x.


ngan:
@orta I ran the test suite locally and everything passed. I think this is a simple bump to [octokit 7](https://github.com/octokit/octokit.rb/releases/tag/v7.0.0). I think the only [breaking change](https://github.com/octokit/octokit.rb/pull/1588) is `delete_reaction`, which danger doesn't seem to use.

orta:
👍🏻

PR #1464, opened 2023-10-24
♻️ Prettify Github JSON response

danielvdao:
### Context

I was trying to look through the tests for the github response and inspecting the response was kind of difficult. Figured I could just prettify it 🙂 

#trivial


danielvdao:
👋🏽 @orta curious if you can take a look? Small PR so would just like to have it closed out if possible!

orta:
We have tonnes of fixtured json in here, not really worth one being pretty formatted when they are probably autogenerated by some script somewhere

PR #1469, opened 2023-11-30
Allow octokit 8.x

ngan:
Update gemspec to support octokit 8.x.

https://github.com/octokit/octokit.rb/releases/tag/v8.0.0


ngan:
@orta -- I'm not seeing any breaking changes for octokit 8.x...I'm starting to wonder if we should not have an upper limit for octokit.

PR #1470, opened 2023-11-30
Don't pin octokit dependency

ngan:
We've [agreed](https://github.com/danger/danger/pull/1469#issuecomment-1834074314) that we don't need to pin the upper version limit for octokit.


ngan:
@orta this is ready 😄

PR #1471, opened 2023-12-13
Lower the minimum Octokit version required

mattr-:
I'm working with Danger on an older project that can't upgrade to Octokit v6 or greater yet but is on v4. Since none of Danger's functionality seems to need things from v6 or greater, we can lower the requirements a bit and hopefully soon I can stop vendoring a local version. :smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 ## master
 
 <!-- Your comment below here -->
-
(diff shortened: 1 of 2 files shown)


orta:
I'd say this is probably fine, yeah, and as it doesn't do a top constraint we can be pretty flexible

PR #1477, opened 2024-01-15
Delete stale violations when there are no new ones

iangmaia:
I've stumbled upon this bug that specially showed when using https://github.com/ashfurrow/danger-rubocop.

Very easy to reproduce:
1. Create a PR with RuboCop violations and `danger-rubocop` configured to report violations inline.
2. **Danger runs** -> Violations are added as PR comments
3. Fix some violations in the code
4. **Danger runs** -> Inline PR comments of the removed violations are correctly gone
5. Fix all violations in the code
6. **Danger runs** -> Inline PR comments of the remaining violations aren't deleted 😞 

I went ahead and tried to see if there was something in the code that could point to a misuse on my side, but it didn't look like it.
After some investigation, I saw that the method `submit_inline_comments` was removing the old violations only when there were new violations to be added, exactly like I was experiencing, so I slightly changed the method to do it when there were no violations as well.
It fixed my bug, but I'm not experienced enough in the codebase to foresee eventual side effects or if there was a better solution for this bug -- please let me know of possible alternatives that being the case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 <!-- Your comment below here -->
 * Add keyword arguments to MessageAggregator.aggregate for Ruby 3.X compatibility - [@dirtyhabits97](https://github.com/dirtyhabits97) [#1466](https://github.com/danger/danger/pull/1466)
+* Fix: remove stale violations when there are no new violations to report - [@iangmaia](https://github.com/iangmaia) [#1477](https://github.com/danger/danger/pull/1477)
 <!-- Your comment above here -->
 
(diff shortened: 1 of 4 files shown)


orta:
Yeah, I can see why it was built this way, and agree on it being a bug - thanks, this looks good to me!

manicmaniac (approved):

manicmaniac (commented):
Looks good :+1: Could you rebase on master?

iangmaia:
@manicmaniac:
> Looks good 👍 Could you rebase on master?

Done! 👍

Issue #968, opened 2018-03-14
Incorrect Gitlab repo slug on Bitrise


authiatr:
# Report

## What did you do?

I followed the Danger [getting started](http://danger.systems/guides/getting_started.html) tutorial and  I ran `danger-gitlab` on Bitrise CI for a **Gitlab** repository. 
I set the `DANGER_GITLAB_API_TOKEN` env var in the secrets variables.
Our GitLab **isn't** self hosted.

## What did you expect to happen?

I expected Danger to run correctly by checking my PR.

## What happened instead?

Danger wasn't able to find the project and got an 404 HTTP error:
```
Gitlab::Error::NotFound: Server responded with code 404, message: 404 Project Not Found. Request URI:
```
The gitlab repository url looks like this format: `https://gitlab.com/my-team/path/to/project_name`
When Danger try to reach the Gitlab API it request the next URI:
`https://gitlab.com/api/v4/projects/to%2Fproject_name/merge_requests/123`

It looks like the Gitlab slug is invalid it should be `path/to/project_name` URL encoded so `patho%2Ftoo%2Fproject_name`. When I manually reach the URL as I think it should be (`https://gitlab.com/api/v4/projects/patho%2Ftoo%2Fproject_name/merge_requests/123`) I have the valid Gitlab JSON.

## Your Environment

* Which CI are you running on? **Bitrise.io**
* Are you running the latest version of Danger? **Yes**
* What is your Dangerfile? (Useless in my case because Danger isn't reaching this step)

  ```ruby
  warn("PR is classed as Work in Progress") if github.pr_title.include? "[WIP]"

  ```

I'll really appreciate your help, I have been stuck with this problem since few days trying to figure it out...

orta:
This could be that you have a different version of GitLab to what's expected, or that the API has changed?

authiatr:
I also tried to import the `gitlab` gem in my Gemfile but it didn't change the result.

authiatr:
And our gitlab version is: GitLab Enterprise Edition 10.6.0-rc4-ee

orta:
I assume you'll might need to set these then:

- `DANGER_GITLAB_HOST` to the host that GitLab is running on.
- `DANGER_GITLAB_API_BASE_URL` to the host that the GitLab API is reachable on.

And maybe the defaults in Danger are wrong now?

authiatr:
[...]

PR #1478, opened 2024-01-31
Use PROJECT_ID instead of PROJECT_PATH in Gitlab urls

simonfelding:
This PR fixes a bug where danger would not work with (some?) self-hosted instances.

Before this patch, I got a 404 from our self-hosted Gitlab CE API. Now it works as expected.

It also fixes issues like https://github.com/danger/danger/issues/968 because the problem is deriving a valid URL for the project. Since PROJECT_ID is unique on a gitlab instance, you can always just use that instead of the PROJECT_PATH.

I tested this and it works. I'm not sure how to run tests in ruby locally, so CI will probably fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## master
 
 <!-- Your comment below here -->
+* Use PROJECT_ID instead of PROJECT_PATH in Gitlab urls. This fixes a bug where Danger would not work with Gitlab self-hosted instances. - [@simonfelding](https://github.com/simonfelding) [#x](https://github.com/danger/danger/pull/x)
 <!-- Your comment above here -->
 
 ## 9.4.1
--- a/lib/danger/ci_source/gitlab_ci.rb
+++ b/lib/danger/ci_source/gitlab_ci.rb
@@ -31,7 +31,7 @@ module Danger
 
     def self.validates_as_pr?(env)
       exists = [
-        "GITLAB_CI", "CI_PROJECT_PATH"
+        "GITLAB_CI", "CI_PROJECT_ID"
       ].all? { |x| env[x] }
 
       exists && determine_pull_or_merge_request_id(env).to_i > 0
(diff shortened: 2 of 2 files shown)


simonfelding:
I have the same patch ready for danger-js, if this one is acceptable.

CHANGELOG.md
```diff
@@ -11,7 +11,7 @@
 ## master
 
 <!-- Your comment below here -->
-* Add keyword arguments to MessageAggregator.aggregate for Ruby 3.X compatibility - [@dirtyhabits97](https://github.com/dirtyhabits97) [#1466](https://github.com/danger/danger/pull/1466)
+* Use PROJECT_ID instead of PROJECT_PATH in Gitlab urls. This fixes a bug where Danger would not work with Gitlab self-hosted instances. - [@simonfelding](https://github.com/simonfelding) [#1478](https://github.com/danger/danger/pull/1478)
```
manicmaniac, line 14: Please do not delete the old one, just append a new entry.

lib/danger/ci_source/gitlab_ci.rb
```diff
@@ -42,7 +42,7 @@ def self.determine_pull_or_merge_request_id(env)
       return env["CI_EXTERNAL_PULL_REQUEST_IID"] if env["CI_EXTERNAL_PULL_REQUEST_IID"]
       return 0 unless env["CI_COMMIT_SHA"]
 
-      project_path = env["CI_MERGE_REQUEST_PROJECT_PATH"] || env["CI_PROJECT_PATH"]
+      project_path = env["CI_MERGE_REQUEST_PROJECT_ID"] || env["CI_PROJECT_ID"]
```
manicmaniac, line 45: I think the variable `project_path` should be renamed as well.

manicmaniac (approved):
Please fix a test cases in `spec/lib/danger/ci_sources/gitlab_ci_spec.rb`.
It seems to use `CI_MERGE_REQUEST_PROJECT_PATH`.

You can run specs locally with the following command.

```
bundle install
bundle exec rake spec
```

manicmaniac (changes requested):
Let me change review status to request changes until the issues are resolved.

lexispike:
Any updates on this change? I'm experiencing this same issue right now while trying to setup DangerJS on a self-hosted Gitlab Instance!

xximwon (approved):

PR #1492, opened 2024-06-21
Update SCM module to handle invalid UTF-8 characters from git output

fuzzyrichie:
In some cases, it seems that Danger will fail if the output contains a single character out of place, failing with "invalid byte sequence in UTF-8 (Encoding::CompatibilityError)" when running rstrip.

### DIFF
--- a/lib/danger/scm_source/git_repo.rb
+++ b/lib/danger/scm_source/git_repo.rb
@@ -53,7 +53,7 @@ module Danger
       Dir.chdir(self.folder || ".") do
         git_command = string.split(" ").dup.unshift("git")
         Open3.popen2(default_env, *git_command) do |_stdin, stdout, _wait_thr|
-          stdout.read.rstrip
+          stdout.read.scrub.rstrip
         end
       end
     end


lib/danger/scm_source/git_repo.rb
```diff
@@ -53,7 +53,7 @@ def exec(string)
       Dir.chdir(self.folder || ".") do
         git_command = string.split(" ").dup.unshift("git")
         Open3.popen2(default_env, *git_command) do |_stdin, stdout, _wait_thr|
-          stdout.read.rstrip
+          stdout.read.scrub.rstrip
```
manicmaniac, line 56: @fuzzyrichie Sorry I don't understand what *invalid UTF-8 character* comes here.

Could you describe minimal steps to reproduce it, and add some test cases?

fuzzyrichie: I'll look to get a sample of the output that caused the error, give me a day or two to set that up!

xximwon (approved):

xximwon (approved):

xximwon (approved):

PR #1500, opened 2024-09-29
Run Danger on this repo

manicmaniac:
https://github.com/danger/danger/blob/cb381b813d0199d415a365fcb36ef7a4b66ed25c/.github/workflows/CI.yml enabled running Danger. But the next commit https://github.com/danger/danger/blob/7da261774d02359b723aec83abc3ba72fe7e8149/.github/workflows/CI.yml disabled it by adding `echo` before the danger command.

I don't know the reason but I think it is useful to run Danger on this repo.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -44,10 +44,27 @@ jobs:
 
       - name: Running Tests
         run: bundle exec rake spec
-
+  danger:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      issues: write
+      pull-requests: write
+      statuses: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -44,10 +44,27 @@ jobs:
 
       - name: Running Tests
         run: bundle exec rake spec
-
+  danger:
+    runs-on: ubuntu-latest
+    permissions:
```
manicmaniac, line 49: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#changing-the-permissions-in-a-forked-repository

> You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access.

PR #1505, opened 2025-02-06
Use #to_h instead of Hash[Array#collect]

ojab:
See commit messages, should be pretty clear. 

Was prompted to do it due to https://github.com/ruby/json/issues/697, found some other issues after enforcing ASCII locale in specs and fixed https://github.com/danger/danger/pull/1492 as well along the way.

### DIFF
--- a/lib/danger/ci_source/circle_api.rb
+++ b/lib/danger/ci_source/circle_api.rb
@@ -45,7 +45,7 @@ module Danger
       url = "project/#{repo_slug}/#{build_number}"
       params = { "circle-token" => token }
       response = client.get url, params, accept: "application/json"
-      JSON.parse(response.body, symbolize_names: true)
+      JSON.parse(response.body, symbolize_names: true, encoding: Encoding::UTF_8)
     end
   end
 end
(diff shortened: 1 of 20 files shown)


ojab:
`1 workflow awaiting approval`, sending to review, `bundle exec rspec/rubocop` passes locally.

lib/danger/ci_source/circle_api.rb
```diff
@@ -45,7 +45,7 @@ def fetch_build(repo_slug, build_number, token)
       url = "project/#{repo_slug}/#{build_number}"
       params = { "circle-token" => token }
       response = client.get url, params, accept: "application/json"
-      JSON.parse(response.body, symbolize_names: true)
+      JSON.parse(response.body, symbolize_names: true, encoding: Encoding::UTF_8)
```
manicmaniac, line 48: I'm afraid that `JSON.parse()` does not have option `encoding`.
https://docs.ruby-lang.org/en/master/JSON.html#module-JSON-label-Parsing+Options

The workaround described in ruby/json is adding `encoding` option to `File.read()`.
https://github.com/ruby/json/issues/697#issuecomment-2462442823

PR #1526, opened 2025-08-26
Remove upper bound constraint on the git gem

andrewmarkle:
This removes the upper bound constraint for using the git gem. Right now git is on 4.0.5 https://github.com/ruby-git/ruby-git/releases which makes danger 2 major versions behind. It leaves us unable to update git-ruby.  

Most of the breaking changes in the git gem are around which version of ruby the gem supports.


manicmaniac (approved):

manicmaniac:
Thank you 👍

PR #1530, opened 2025-09-28
Apply frozen string literal magic comment

ydah:
Starting with Ruby 4.x, there is a plan to gradually freeze strings: https://bugs.ruby-lang.org/issues/20205#note-35

If we use Ruby 3.5 dev with code before this PR, you will get the following warning:

```
$ RUBYOPT=-W:deprecated bundle exec rake
(snip)

/Users/ydah/danger/lib/danger/danger_core/dangerfile.rb:188: warning: literal string will be frozen in the future (run with --debug-frozen-string-literal for more information)
```

This warning suggests that applying the frozen string magic comment will lead to `FrozenError`. So, this indicates that the future behavior of (default frozen) string will produce `FrozenError`.
Therefore, I resolved `Style/FrozenStringLiteralComment` and fixed FrozenError.


ydah:
The impact is bigger than I thought, so I will close this PR currently.

eiskrenkov:
Hey @ydah! What is the biggest issue with incorporating frozen string literals into danger's codebase?

Issue #1544, opened 2026-04-14
TeamCity + Bitbucket Cloud: support BITBUCKET_PULL_REQUEST_ID env var to avoid API lookup


vakhidbetrakhmadov:
**Problem:**

When running Danger on TeamCity with Bitbucket Cloud, [bitbucket_pr_from_env](https://github.com/danger/danger/blob/master/lib/danger/ci_source/teamcity.rb#L155) always resolves the pull request ID by querying the Bitbucket API via [fetch_pr_from_branch](https://github.com/danger/danger/blob/20f1243775da2ce53f55ad5ba65112de5da6e118/lib/danger/request_sources/bitbucket_cloud_api.rb#L17). This API call is not always necessary and is susceptible to rate limiting (HTTP 429), which crashes Danger with a `RuntimeError` since [fetch_json](https://github.com/danger/danger/blob/20f1243775da2ce53f55ad5ba65112de5da6e118/lib/danger/request_sources/bitbucket_cloud_api.rb#L132) has no retry logic.

This is especially problematic in large monorepos where many PR builds run concurrently under a shared service account, quickly exhausting the Bitbucket API rate limit.

**Context:**

The other three CI source integrations in TeamCity all read the pull request ID directly from environment variables - `GITHUB_PULL_REQUEST_ID`, `GITLAB_PULL_REQUEST_ID`, and `BITBUCKETSERVER_PULL_REQUEST_ID`. Bitbucket Cloud is the only path that requires an API call to resolve it.
   
TeamCity's [Pull Requests](https://www.jetbrains.com/help/teamcity/pull-requests.html) build feature exposes the PR number via [predefined build parameters](https://www.jetbrains.com/help/teamcity/pull-requests.html#Predefined+build+parameters+for+pull+requests), making it straightforward to set as an env var.

**Proposed change:**

Update `bitbucket_pr_from_env` to check for a `BITBUCKET_PULL_REQUEST_ID` env var first, falling back to the existing API lookup if not set:
```
def bitbucket_pr_from_env(env)
  return env["BITBUCKET_PULL_REQUEST_ID"].to_i if env["BITBUCKET_PULL_REQUEST_ID"] # added

  branch_name = env["BITBUCKET_BRANCH_NAME"]
  repo_slug   = env["BITBUCKET_REPO_SLUG"]
  begin
    Danger::RequestSources::BitbucketCloudAPI.new(repo_slug, nil, branch_name, env).pull_request_id
  rescue StandardError
[...]

PR #1545, opened 2026-04-14
Support BITBUCKET_PULL_REQUEST_ID env var for TeamCity + Bitbucket Cloud (#1544)

vakhidbetrakhmadov:
When running on TeamCity with Bitbucket Cloud, Danger always resolves the pull request ID by querying the Bitbucket API. This is prone to HTTP 429 (rate limiting) in setups with many concurrent builds sharing a single service account.

Check for BITBUCKET_PULL_REQUEST_ID in the environment and use it directly when available, falling back to the existing API lookup. This is consistent with how GitHub, GitLab, and Bitbucket Server integrations already accept the PR ID from the environment.


acecilia (approved):

acecilia (approved):

acecilia:
Thanks for the PR! Looks simple and safe enough to merge

Issue #1546, opened 2026-06-09
Danger Docker Image Large


beninabox:
The size of the docker image produced here is significant because it contains all of the `ruby:3.2` image as a base. We can significantly reduce the size of it by leveraging multi stage builds. This will also reduce the attack surface significantly.

PR #1547, opened 2026-06-09
#1546 Reduce Docker Image Size

beninabox:
This change improves the size of the Docker image by utilising multi stage builds and the alpine image so that we can reduce the attack surface area and the overall size of the image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.git
+spec
+docs
+Guardfile
+appveyor.yml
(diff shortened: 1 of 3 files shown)


manicmaniac (approved):
Thank you 🎉

beninabox:
Thanks for merging @manicmaniac, would it be possible to make a patch release containing this? As my org needs to pull a specific version of the docker image rather than master. Cheers

